PR #12, opened 2021-09-23
Configure package according to proposed directory structure

asogaard:
I have moved the files around a little bit to match the proposed directory structure (see wiki), added setup and init files to expose a `gnn_reco` package, and have move the SQLiteDataConverter MWE script to a new `examples` directory. It should be possible to run this script by activating the icetray shell, installing the `gnn_reco` package, and running `python examples/test_sqlite.py`.


examples/test_sqlite.py
```diff
@@ -0,0 +1,20 @@
...
+    pulsemap = 'SRTInIcePulses'
+    gcd_rescue = '/groups/icecube/stuttard/data/oscNext/pass2/gcd/GeoCalibDetectorStatus_AVG_55697-57531_PASS2_SPE_withScaledNoise.i3.gz'
+    outdir = '/groups/icecube/asogaard/temp/sqlite_test'
+    db_name = 'gnn-reco_test'
+    workers = 2
+
+    SQLiteDataConverter(paths, mode, pulsemap, gcd_rescue, outdir, db_name, workers)
+
+if __name__ == '__main__':
+    main()
```
RasmusOrsoe, line 20: I would personally prefer to move `if __name__` checks into `SQLiteDataConverter` such that the user doesnt have to worry about this.

src/gnn_reco/data/sqlite_dataconverter.py
```diff
@@ -233,15 +232,14 @@ def _processfiles(self):
...
-                p.join()
-                print('pool')
-                self._merge_databases()
+            
+            print('starting pool!')
+            p = Pool(processes = workers)
+            p.map_async(parallel_extraction, settings)
+            p.close()
+            p.join()
+            print('pool')
```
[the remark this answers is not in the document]
asogaard: As far as I can tell, `_process_files` is not called in any child process; therefore I don't know when this would be an issue. I have also run the script several times with no problems, so it seems like this should work. Otherwise I figure we would need to define a minimal working example (MWE) for recreating such a potential child-processes-spawning-child-processes problem?

examples/test_sqlite.py
asogaard: It is good practice that all scripts that are meant to be used as executables include such a `__name__` guard to avoid executing the script on import (see https://google.github.io/styleguide/pyguide.html#317-main). However, I have also tried to run the script without this guard in place with the same results, meaning that this is strictly not something users _have_ to worry about, although it would be good form if they did. :)  Also, this check serves a different purpose that the `if __name__ == ...` check in `SQLiteDataConverter` and so it won't be a choice of one or the other.

asogaard:
I have addressed the review comments, @RasmusOrsoe. Please let me know whether you are happy with the proposed resolution. :)

asogaard:
Have merged the changes from #13 into this PR, which should therefore be ready to merge.

RasmusOrsoe (approved):
Glad to hear you tested it! I dont get why it works but I'll read up on that later.

Rasmus

PR #13, opened 2021-09-24
Added Google-style Docstrings

RasmusOrsoe:
Removed old docstrings and added Google-style docstrings.
Removed a few debugging print messages, but not all!


asogaard (approved):
Looks good! Approving right away.

Issue #11, opened 2021-09-23
Move parallelised functions to private methods in `SQLiteDataConverter`


asogaard:
It has not been possible to put a private method (e.g. a function defined within the class) as an argument for map_async, and therefore the functionally currently resides in a regular function.

asogaard:
Will work on this now

PR #18, opened 2021-09-24
Refactor SQLiteDataConverter and DataConverter

asogaard:
Working on #11 as well as a general refactor and simplification of these two classes. Creating a PR draft, as the work is still underway, in order to be able to start having a discussion.

### DIFF
--- a/examples/test_sqlite.py
+++ b/examples/test_sqlite.py
@@ -14,7 +14,8 @@ def main():
     db_name = 'gnn-reco_test'
     workers = 2
 
-    SQLiteDataConverter(paths, mode, pulsemap, gcd_rescue, outdir, db_name, workers)
+    converter = SQLiteDataConverter(outdir, mode, pulsemap, gcd_rescue, db_name=db_name, workers=workers)
+    converter(paths)
 
 if __name__ == '__main__':
     main()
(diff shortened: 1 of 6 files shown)


asogaard:
Hi @RasmusOrsoe,
The PR now goes some way towards what it says on the tin (refactoring the two classes), and perhaps most importantly it addresses issue #11. To avoid ballooning this PR any further I propose not making any more changes (except for requests during review) before we have had a chance to discuss and merge this first.

RasmusOrsoe (commented):
Hi! Looks nice. 

I can see that the i3/gcd file search has changed quite  a bit and I understand very little of the new code. Are we 100% sure its identical?

If so, lets merge.

Rasmus

asogaard:
Thanks for reviewing, @RasmusOrsoe! I have checked, and the two should agree, but I will prepare a double-check just to be sure.

asogaard:
For reference, the new function(s) make two changes:
* They explicitly ignore files that don't resemble I3 or GCD files.
* Return all I3 files in a certain directory, regardless of depth.

RasmusOrsoe (approved):
i3/gcd finder tested and works.

PR #32, opened 2021-10-08
Add torch Dataset-inheriting class for SQLite "backend"

asogaard:
Create a torch Dataset class for reading from intermediate SQLite databases. This requires installing pytorch and pytorch-geometric as part of the package setup, which is also implemented. (NB: This unfortunately requires an annoying hack to install pytorch strictly before pytorch-geometric, which is a known problem.*) Finally, the setup script itself is refactored to utilise "extras tags" to allow for configuring setup specifically for e.g. build tests and linting.
_____________________
* See e.g. https://github.com/pyg-team/pytorch_geometric/issues/861#issuecomment-566424944

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -22,6 +22,7 @@ INSTALL_REQUIRES = [
     'sqlalchemy',
     'pandas',
     'numpy',
+    'timer',
     'tqdm',
     'torch-scatter',
     'torch-sparse',


RasmusOrsoe (approved):
```
graph = Data(
            x=x,
            y=y.unsqueeze(0),
            **labels_dict,
        )

```
Cool trick with the labels!!

PR #33, opened 2021-10-14
Major changes

RasmusOrsoe:
This contains main functionality additions.

- Trainer class (v1.0)
- Predictor class (v0.1)
- loss functions 
- dynedge model 
- custom learning rate schedule
- adhoc utility 
- minor changes to sqlite_dataset.py

All of which are tested (and it works!) in `examples/`

We now have train/test functionality :-)

### DIFF
--- a/src/gnn_reco/components/loss_functions.py
+++ b/src/gnn_reco/components/loss_functions.py
@@ -1,0 +1,33 @@
+import torch
+import numpy as np
+from torch.nn import CrossEntropyLoss
+
+
+def LogCosh(prediction, graph, target):
+    return torch.sum(torch.log(torch.cosh(((prediction[:,0]-graph[target])))))
+
+def CustomCrossEntropyLoss(prediction, graph, target):
+    f = CrossEntropyLoss()
+    return f(prediction,graph[target].long())
+
+
+def VonMisesSineCosineLoss(prediction, graph, target):
+    k            = torch.abs(prediction[:,2])
+    angle  = graph[target]
+    
+    u_1 = (1/torch.sqrt(torch.tensor(2,dtype = torch.float)))*torch.sin(angle)
+    u_2 = (1/torch.sqrt(torch.tensor(2,dtype = torch.float)))*torch.cos(angle)
+    u_3 = (1/torch.sqrt(torch.tensor(2,dtype = torch.float)))*(1)
+    
+    norm_x  = torch.sqrt(1 + prediction[:,0]**2 + prediction[:,1]**2)
+    
+    x_1 = (1/norm_x)*prediction[:,0]
+    x_2 = (1/norm_x)*prediction[:,1]
+    x_3 = (1/norm_x)*(1)
+    
+    dotprod = u_1*x_1 + u_2*x_2 + u_3*x_3
+    logc_3 = - torch.log(k) + k + torch.log(1 - torch.exp(-2*k))    
+    loss = torch.mean(-k*dotprod + logc_3)
+
+    
+    return loss
--- a/src/gnn_reco/components/utils.py
+++ b/src/gnn_reco/components/utils.py
@@ -1,0 +1,179 @@
+
+import torch
+from tqdm import tqdm
+import pandas as pd
+import numpy as np
+
+class EarlyStopping(object):
+    def __init__(self, mode='min', min_delta=0, patience=10, percentage=False):
+        self.mode = mode
+        self.min_delta = min_delta
+        self.patience = patience
+        self.best = None
+        self.num_bad_epochs = 0
+        self.is_better = None
+        self._init_is_better(mode, min_delta, percentage)
+
+        if patience == 0:
+            self.is_better = lambda a, b: True
+            self.step = lambda a: False
+
+    def step(self, metrics,model):
+        if self.best is None:
+            self.best = metrics
+            return False
+
+        if torch.isnan(metrics):
+            return True
+
+        if self.is_better(metrics, self.best):
+            self.num_bad_epochs = 0
+            self.best = metrics
+            self.best_params = model.state_dict()
+        else:
+            self.num_bad_epochs += 1
+
+        if self.num_bad_epochs >= self.patience:
+            return True
+
+        return False
+
+    def _init_is_better(self, mode, min_delta, percentage):
+        if mode not in {'min', 'max'}:
+            raise ValueError('mode ' + mode + ' is unknown!')
+        if not percentage:
+            if mode == 'min':
+                self.is_better = lambda a, best: a < best - min_delta
+            if mode == 'max':
+                self.is_better = lambda a, best: a > best + min_delta
+        else:
+            if mode == 'min':
+                self.is_better = lambda a, best: a < best - (
+                            best * min_delta / 100)
+            if mode == 'max':
+                self.is_better = lambda a, best: a > best + (
+                            best * min_delta / 100)
+    def GetBestParams(self):
+        return self.best_params
+
+class PiecewiseLinearScheduler(object):
+    def __init__(self, training_dataset, start_lr, max_lr, end_lr, max_epochs):
+        self._start_lr = start_lr
+        self._max_lr   = max_lr
+        self._end_lr   = end_lr
+        self._steps_up = int(len(training_dataset)/2)
+        self._steps_down = len(training_dataset)*max_epochs - self._steps_up
+        self._current_step = 0
+        self._lr_list = self._calculate_lr_list()
+
+    def _calculate_lr_list(self):
+        res = list()
+        for step in range(0,self._steps_up+self._steps_down):
+            slope_up = (self._max_lr - self._start_lr)/self._steps_up
+            slope_down = (self._end_lr - self._max_lr)/self._steps_down
+            if step <= self._steps_up:
+                res.append(step*slope_up + self._start_lr)
+            if step > self._steps_up:
+                res.append(step*slope_down + self._max_lr -((self._end_lr - self._max_lr)/self._steps_down)*self._steps_up)
+        return torch.tensor(res)
+
+    def get_next_lr(self):
+        lr = self._lr_list[self._current_step]
+        self._current_step = self._current_step + 1
+        return lr
+    
+class Trainer(object):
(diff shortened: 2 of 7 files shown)


src/gnn_reco/components/loss_functions.py
```diff
@@ -0,0 +1,30 @@
+import torch
+import numpy as np
+from torch.nn import CrossEntropyLoss
+
+
+def LogCosh(prediction, graph, target):
```
asogaard, line 6: As this is used as a function (compared to e.g. `CrossEntropyLoss` from `torch.nn`, which is a class) I would propose using "snake_case" here, as well as in L9 and L14 of this file. Alternatively, these three could be changed to classes to conform with torch standards. I would probably vote for the latter, but have no strong opinions as long as we're just following standard naming conventions.

src/gnn_reco/components/loss_functions.py
```diff
@@ -0,0 +1,30 @@
...
+    return torch.sum(torch.log(torch.cosh(((prediction[:,0]-graph[target])))))
+
+def CustomCrossEntropyLoss(prediction, graph, target):
```
asogaard, line 9: This function, and the one above, are currently not being used — but I think that's fine, if we foresee using them. This function, however, seems to be a minimal wrapper around `torch.nn.CrossEntropyLoss`, where the `graph[target].long()` bit easily could (should?) be moved to the relevant part of the model code.

src/gnn_reco/components/loss_functions.py
```diff
@@ -0,0 +1,30 @@
...
+    f = CrossEntropyLoss()
+    return f(prediction,graph[target].long())
+
+
+def VonMisesSineCosineLoss(prediction, graph, target):
```
asogaard, line 14: This function seems to make non-trivial assumptions about the structure of `prediction`. It might be beneficial to have a docstring that detail these assumptions?

src/gnn_reco/components/utils.py
```diff
@@ -0,0 +1,266 @@
...
+from torch_geometric.data.batch import Batch
+import os
+from sklearn.preprocessing import RobustScaler
+import sqlite3
+import pandas as pd
+import pickle
+import numpy as np
+import os
+
+class EarlyStopping(object):
```
asogaard, line 19: From here and to L155, most of this code could probably be imported from `pytorch-ligtning` (see e.g. https://pytorch-lightning.readthedocs.io/en/1.4.0/api/pytorch_lightning.callbacks.early_stopping.html; without me knowing the details of that package) which could reduce maintenance overhead. I think we should leave the code in this PR as-is, but am mainly commenting this to hear your thoughts on this, @RasmusOrsoe and @mhaminh.

src/gnn_reco/components/utils.py
```diff
@@ -0,0 +1,266 @@
...
+                batch_of_graphs.to(self.device)
+                target.extend(batch_of_graphs[self.target].detach().cpu().numpy())
+                predictions.extend(self.model(batch_of_graphs).detach().cpu().numpy())
+                event_nos.extend(batch_of_graphs['event_no'].detach().cpu().numpy())
+        out = pd.DataFrame(data = predictions, columns = self.output_column_names)
+        out['event_no'] = event_nos
+        out[self.target] = target
+        return out
+
+def vonmises_sine_cosine_post_processing(predictions,target):
```
asogaard, line 189: A similar calculation is done three places in this PR. I think we should leave the code as-is for now, but this might be one area where DRY (don't repeat yourself) could be applied later.

src/gnn_reco/components/utils.py
```diff
@@ -0,0 +1,266 @@
...
+    predictions[target + '_pred'] = np.arctan2(predictions['sine'], predictions['cosine'])
+    predictions['k'] = abs(predictions['k'])
+    predictions['sigma'] = 1/predictions['k']
+    return predictions        
+
+
+def make_train_validation_dataloader(db, selection, pulsemap, batch_size, FEATURES, TRUTH, num_workers):
+    training_selection, validation_selection = train_test_split(selection, test_size=0.33, random_state=42)
```
asogaard, line 197: Not a change request, just a note: We should decide where in the pipeline, from I3 files to model inference, we want to define train-and-test-splits such that these become unambiguous and transferrable between models/runs/projects. I have always used modulo of event numbers as a reproducible way of making these splits, but am interested to hear your thoughts, @RasmusOrsoe, @mhaminh, and @philippeller.

src/gnn_reco/components/utils.py
```diff
@@ -0,0 +1,266 @@
...
+    #         scaler = RobustScaler()
+    #         if truth == 'energy':
+    #             truth_scalers[truth] = scaler.fit(np.array(np.log10(truth_data[truth])).reshape(-1,1))
+    #         else:
+    #             truth_scalers[truth] = scaler.fit(np.array(truth_data[truth]).reshape(-1,1))
+
+    #     comb_scalers = {'truth': truth_scalers, 'input': feature_scaler}
+    #     os.makedirs(outdir + '/meta', exist_ok= True)
+    #     with open(outdir + '/meta/transformersv2.pkl','wb') as handle:
+    #         pickle.dump(comb_scalers,handle,protocol = pickle.HIGHEST_PROTOCOL)
```
asogaard, line 264: Should this code be commented out?

src/gnn_reco/data/sqlite_dataset.py
```diff
@@ -116,20 +119,20 @@ def _create_graph(self, features, truth):
...
+            edge_index= None
         )
-        
+        graph.n_pulses = n_pulses
+
+        for label in labels_dict.keys():
+            graph[label] = labels_dict[label]    
+
+        for truth_key in ['energy', 'position_x', 'position_y', 'position_z', 'azimuth', 'zenith']:
+            graph[truth_key] = torch.tensor(truth_dict[truth_key], dtype = torch.float64)
```
asogaard, line 134: Am happy with all of these changes, but out of curiosity: If these are the reconstruction targets, what is the argument for treating these as custom labels rather than a standard target (`y=...`)? Is it to force named access to the targets, as opposed to integer indexing a generic target tensor?

src/gnn_reco/models/dynedge.py
```diff
@@ -0,0 +1,181 @@
...
+from torch_scatter import scatter_min
+from torch_scatter import scatter_max
+from torch_cluster import knn_graph
+from torch_geometric.nn import EdgeConv
+from torch_geometric.nn.norm import LayerNorm
+from gnn_reco.models.utils import calculate_xyzt_homophily
+
+
+
+class dynedge_energy_xfeats(torch.nn.Module):
```
asogaard, line 15: I would suggest using "PascalCase" for class names.

src/gnn_reco/models/dynedge.py
```diff
@@ -0,0 +1,181 @@
...
+        x2 = self.relu(x2)
+        x2 = self.nn3(x2)
+        
+        x2 = self.relu(x2)
+        x2 = self.nn4(x2)
+
+        return x2
+
+
+class dynedge_angle_xfeats(torch.nn.Module):
```
asogaard, line 92: I would suggest using "PascalCase" for class names.

src/gnn_reco/models/convnet.py
```diff
@@ -0,0 +1,125 @@
+import torch
```
asogaard, line 1: As mentioned in the general comments: As the focus is on integrating model training code in the repo quickly, I would review the model-specific files as they already work.

src/gnn_reco/models/dynedge.py
```diff
@@ -0,0 +1,181 @@
+from math import atan2
```
asogaard, line 1: As mentioned in the general comments: As the focus is on integrating model training code in the repo quickly, I would review the model-specific files as they already work.

src/gnn_reco/models/utils.py
```diff
@@ -0,0 +1,7 @@
+from torch_geometric.utils.homophily import homophily
+def calculate_xyzt_homophily(x,edge_index, batch):
```
asogaard, line 2: To an "outsider," at least, this is not a standard approach, and so a docstring detailing the intention behind this function would be useful.

asogaard (changes requested):
Hi @RasmusOrsoe,
Please see my review comments. Once you have had a change to go over them, I will approve the merge. 👍

src/gnn_reco/models/utils.py
RasmusOrsoe: Thanks. Added docstring.

src/gnn_reco/models/dynedge.py
RasmusOrsoe: OK.

src/gnn_reco/models/convnet.py
RasmusOrsoe: OK.

src/gnn_reco/models/dynedge.py
RasmusOrsoe: New commit - now just a single class

src/gnn_reco/components/utils.py
RasmusOrsoe: I agree. I removed this as its unused.

src/gnn_reco/components/loss_functions.py
RasmusOrsoe: doc string added

src/gnn_reco/components/loss_functions.py
RasmusOrsoe: I changed the names

src/gnn_reco/data/i3extractor.py
```diff
@@ -74,6 +75,13 @@ def __call__(self, frame) -> dict:
             'width' : [],
             'pmt_area': [],
             'rde': [],
+            'string': [],
+            'pmt_number': [],
+            'dom_number': [],
+            'pmt_dir_x': [],
+            'pmt_dir_y': [],
+            'pmt_dir_z': [],
+            'dom_type': [],
```
asogaard, line 84: @RasmusOrsoe here, and below (L97-103) an Upgrade-specific change seems to have snuck in. Is this intentional? Because I suspect this might give errors when running the existing example code/using the default GCD files. Would it be okay to remove these changes for this PR, and include them as part of a separate, Upgrade-specific PR?

PR #66, opened 2021-11-11
Update test_model_training_sqlite.py

RasmusOrsoe:
quickfix


asogaard (approved):
Looks good! We will still need to resolve the duplication in `gnn_reco.components.utils` wrt. `gnn_reco.model_training.utils`, as noted in the former file, but let's leave it at that for now.

PR #68, opened 2021-11-11
Reproducibility

asogaard:
Hi @RasmusOrsoe,

Ahead of the training runs in #55 I am submitting a PR to make the preparation of training and validations splits reproducible between runs, such that it is not necessary to save them to file.


RasmusOrsoe (commented):
Looks good.

RasmusOrsoe (approved):

PR #79, opened 2021-11-18
Minor tweaks

asogaard:
Fix some broken imports, avoid modifying the frame inplace in the I3Extractor (this is bad for inference in an I3Module), and fix the pytorch-geometric version to 2.0.1 for consistency.

@RasmusOrsoe Consider pulling these changes before a clean clone and retraining.


RasmusOrsoe (approved):
Cool beans. Approved

PR #81, opened 2021-11-21
Fixed error related to new model saving/loading method and early stopping method

RasmusOrsoe:


src/gnn_reco/components/loss_functions.py
```diff
@@ -208,7 +208,7 @@ def _forward(self, prediction: Tensor, target: Tensor) -> Tensor:
 
         # Formatting prediction
         angle_pred = prediction[:,0]
-        kappa = prediction[:,1]
+        kappa = abs(prediction[:,1])
```
asogaard, line 211: I don't think this `abs` statement is necessary, as it would imply that the model (in this case, `Task` components) might return a negative value for kappa, which would not be meaningful. If that is the case, it should be fixed in the corresponding model/task. Therefore, I suggest we leave it to the model to impose the correct restrictions on the output, and in the loss function(s) assume that inputs are as expected.

asogaard (changes requested):

src/gnn_reco/components/loss_functions.py
RasmusOrsoe: I have removed the abs() from the PR as its besides the point for what this PR does. I do think theres merit to adding the abs(), but lets talk about that some other time

asogaard (approved):

PR #88, opened 2021-11-24
Tweaks to energy regression task and a few utils

RasmusOrsoe:
Tweaks to energy regression task (required for paper) and a few utils. Tried to merge locally with main branch but I get a conflict but I'm not told where the conflict is....... @asogaard can you help me figure out what needs to be changed in order for this to merge?


asogaard:
Have resolved the merge conflict as per @RasmusOrsoe's requests, but there are a few unit tests that fail. These will need to be resolved before this PR can be merged. Will try to provide a few pointers in the comments. It also seems like this PR does a few different things, which may make it a little harder to merge than a single-issue PR.

src/gnn_reco/components/loss_functions.py
```diff
@@ -73,12 +73,12 @@ def _log_cosh(cls, x: Tensor) -> Tensor:  # pylint: disable=invalid-name
         Used to avoid `inf` for even moderately large differences.
         See [https://github.com/keras-team/keras/blob/v2.6.0/keras/losses.py#L1580-L1617]
         """
-        return x + torch.nn.functional.softplus(-2. * x) - np.log(2.0)
+        return torch.log(torch.cosh(x))#x + torch.nn.functional.softplus(-2. * x) - np.log(2.0)
```
asogaard, line 76: I don't think we should replace the original: The proposed formula produce invalid result for even moderately large `x`; and for the values where the proposed formula does produce valid results, the two expressions are numerically identical, as ensured by the (now failing) unit test here: https://github.com/icecube/gnn-reco/blob/main/tests/test_loss_functions.py#L38

src/gnn_reco/components/loss_functions.py
```diff
@@ -73,12 +73,12 @@ def _log_cosh(cls, x: Tensor) -> Tensor:  # pylint: disable=invalid-name
...
 
     def _forward(self, prediction: Tensor, target: Tensor) -> Tensor:
         """Implementation of loss calculation."""
         assert prediction.dim() == target.dim() + 1
-        diff = prediction[:,0] - target
+        diff = prediction[:,0] - torch.log10(target)
```
asogaard, line 81: I really don't think we should hard-code a log10 transform into a generic LogCosh loss. This is specific to energy regression, and so I would propose this transform be imposed only there, e.g. as:
```
task = EnergyReconstruction(
    loss_function=LogCoshLoss(
        transform_output=lambda x: torch.log10(x),
    ),
    ...
)
```

src/gnn_reco/models/task/reconstruction.py
```diff
@@ -62,7 +62,7 @@ class EnergyReconstruction(Task):
 
     def _forward(self, x):
         # Transform energy
-        return torch.pow(10, x[:,0] + 1.).unsqueeze(1)
+        return x[:,0].unsqueeze(1) #torch.pow(10, x[:,0] + 1.).unsqueeze(1)
```
asogaard, line 65: I think we should aim to have each task return the natural, physical quantity — here: energy — rather than any transformed quantity that the user then needs to (manually) inverse-transform. I think that any scaling of physical quantities for the purposes of training alone should be a matter for the loss function. See comment above to `LogCoshLoss` along with the proposed resolution.

src/gnn_reco/components/loss_functions.py
asogaard: This is also one of the changes that leads to a failing unit test.

asogaard (changes requested):

RasmusOrsoe:
Hi Andreas! Thanks for comments. There are some technical reasons behind some of choices that you comment on and I think we should discuss this on zoom when you have a chance.

asogaard:
Hi @RasmusOrsoe,
Just wanted to follow up on this PR. The energy regression should have been addressed by #94 with examples provided for training that runs. The addition of the interaction time variable is not yet implemented elsewhere, but I figure this is something @kaareendrup is also looking into and that could be added in a separate, dedicated (atomic) PR. Is the rest of this PR something you still want to see merged, or has it been superseded?

RasmusOrsoe:
Hi Andreas!

Yeah I think we can close this PR. On interaction time: I've added this variable to the database production in the current version I'm running such that I could produce a database for @kaareendrup's tests. But merging that will be like merging this PR....

Rasmus

PR #89, opened 2021-11-26
Add tasks for Vertex (xyzt) and Position (xyz) reconstruction

kaareendrup:
Added task classes for position and vertex reconstruction. They are meant to go together with updates to data/constants, training/trainers, task/task and components/loss_functions, that have not been pushed yet. At the moment they do not transform the output from the model in any way.


src/gnn_reco/models/task/reconstruction.py
```diff
@@ -75,3 +75,19 @@ def _forward(self, x):
...
         pred = torch.stack((energy, log_var), dim=1)
         return pred
+
+class VertexReconstruction(Task):
+    # Requires four features, x, y, z and t
+    nb_inputs = 4
+
+    def _forward(self, x):
+        # Initally do nothing to the output
+        return x.unsqueeze(1)
```
asogaard, line 85: The input `x` to `_forward` should be of shape [N, 4], I believe. By unsqueezing, I suspect the output ends up having shape [N, 1, 4]. Is that the case? And is that what you want to happen. I would expect that just returning `x` would suffice, but I might be wrong.

asogaard: Perhaps training would be simpler if the components in `x` were scaled up according to the rough physical dimensions of the IceCube detector; e.g. `x[:,:3] = 100. * x[:,:3]` as initially the elements in `x` will of order 1. Have you had a chance to train with this task? How have you handled this scale?

src/gnn_reco/models/task/reconstruction.py
```diff
@@ -75,3 +75,19 @@ def _forward(self, x):
...
+        # Initally do nothing to the output
+        return x.unsqueeze(1) 
+
+class PositionReconstruction(Task):
+    # Requires three features, x, y, z
+    nb_inputs = 3
+
+    def _forward(self, x):
+        # Initally do nothing to the output
+        return x.unsqueeze(1)
```
asogaard, line 93: Same as first comment above.

asogaard: At some point (not for this PR, necessarily) it might be beneficial to use this task (position) as part of the vertex reconstruction task above, since the former is essential a sub-task of the latter.

src/gnn_reco/models/task/reconstruction.py
kaareendrup: The unsqueeze is just copied over from the other simple tasks like `AngularReconstruction` and `EnergyReconstruction`. It works, but only because I in the loss function do: `diff = prediction[:,0] - target`  which is done in the current `LogCoshLoss `function. Ive tested removing the unsqueezing and the `[:,0] `, and everything still works as before, so I will implement that

kaareendrup: Right now, I only scale the target in the loss function (and only for time). The model is able to output values in the correct xyz range by it's own, but it will probably make it easier to give it a bit of help by scaling by rough orders of magnitude in the task, so I will do that as well!

PR #92, opened 2021-12-01
Adding BinaryClassification Task, BCELoss and GetEventNumbers in utils

BozianuLeon:
BinaryClassification for any 0,1 target column. Fixed booleans to 0 or 1 in sqlite_dataset. Added basic binary cross entropy loss with element wise calculation. Get event numbers from my old work, found easier to get desired number of events out.


src/gnn_reco/components/loss_functions.py
```diff
@@ -82,6 +82,16 @@ def _forward(self, prediction: Tensor, target: Tensor) -> Tensor:
         elements = self._log_cosh(diff)
         return elements
 
+class BinaryLoss(LossFunction):
```
asogaard, line 85:
```suggestion
class BinaryCrossEntropyLoss(LossFunction):
```

src/gnn_reco/components/loss_functions.py
```diff
@@ -82,6 +82,16 @@ def _forward(self, prediction: Tensor, target: Tensor) -> Tensor:
...
+    """ Computes binary cross entropy for a vector of predictions (between 0 and 1), 
+    targets should be 0 and 1 for muon and neutrino respectively
+    where prediction is prob. the PID is neutrino (12,14,16)
+    loss should be reported elementwise, so set reduction to None
+    """
+    
+    def _forward(self, prediction: Tensor, target: Tensor) -> Tensor: 
+        return torch.nn.functional.binary_cross_entropy(prediction,target,reduction='none')
```
asogaard, line 93:
```suggestion
        return torch.nn.functional.binary_cross_entropy(prediction, target, reduction='none')
```

src/gnn_reco/data/sqlite_dataset.py
```diff
@@ -153,7 +153,7 @@ def _create_graph(self, features, truth):
         for write_dict in [labels_dict, truth_dict]:
             for key, value in write_dict.items():
                 try:
-                    graph[key] = torch.tensor(value)
+                    graph[key] = torch.tensor(value,dtype=torch.float)
```
asogaard, line 156:
```suggestion
                    graph[key] = torch.tensor(value, dtype=torch.float)
```

asogaard: This will convert e.g. integers and booleans to floats; for the GNN training itself I reckon this is fine, but perhaps there are other cases that may be affected, e.g. if a boolean attribute is checked? What is the benefit for casting all attributes in this way?

src/gnn_reco/models/task/reconstruction.py
```diff
@@ -75,3 +75,15 @@ def _forward(self, x):
         log_var = x[:,1]
         pred = torch.stack((energy, log_var), dim=1)
         return pred
+
+
+
+
+
+class BinaryClassificationToTarget(Task):
```
asogaard, line 83:
```suggestion
class BinaryClassificationTask(Task):
```

src/gnn_reco/data/utils.py
```diff
@@ -9,6 +9,46 @@
 from typing import List, Tuple
 import sqlite3
 
+
+def GetDesiredEventNumbers(db_path, desired_size, fraction_noise = 0, fraction_nu_e = 0, fraction_muon = 0, fraction_nu_mu = 0, fraction_nu_tau = 0):
```
asogaard, line 13:
```suggestion
def get_desired_event_numbers(db_path, desired_size, fraction_noise=0, fraction_nu_e=0, fraction_muon=0, fraction_nu_mu=0, fraction_nu_tau=0):
```

asogaard: I would probably reorder the arguments so as to keep the three neutrino flavours next to each other rather than having the muon in the middle, e.g.:`noise`,  `muon`, `nu_e`, `nu_mu`, `nu_tau`.

asogaard: The name of this function could probably also be a bit more descriptive, as "desired" is quite subjective/contextual. I think it is perfectly fine for now, and the intention is very clear from the arguments, just something to think about going forward.

src/gnn_reco/data/utils.py
```diff
@@ -9,6 +9,46 @@
...
+            desired_size = len(tot_event_nos)
+            numbers_desired = [x * desired_size for x in fracs]
+            print('Only {} events in database, using this number instead.'.format(len(tot_event_nos)))
+
+        list_of_dataframes = []
+        restart_trigger = True
+        while restart_trigger:
+            restart_trigger = False
+            for number,particle_type in zip(numbers_desired,pids):
+                query_is = 'SELECT event_no FROM truth WHERE abs(pid) == {} LIMIT {}'.format(particle_type,number)
```
asogaard, line 33: By using `LIMIT` here, we are selecting a non-random subset of events. It would probably be better practices (although slower to run) to load in all event numbers for a specific PDG ID and then take a random subset of these.

src/gnn_reco/data/utils.py
```diff
@@ -9,6 +9,46 @@
...
+                        print('There are no particles of type {} in this database please make new request.'.format(particle_type))
+                        return None
+                    numbers_desired = [int(new_x * (len(dataframe)/number)) for new_x in numbers_desired]
+                    restart_trigger = True
+                    list_of_dataframes = []
+                    break
+                list_of_dataframes.append(dataframe)
+        
+        retrieved_event_nos_pd = pd.concat(list_of_dataframes)
+        event_no_list = retrieved_event_nos_pd.sample(frac=1, replace=False, random_state=np.random.RandomState(seed=0)).values.ravel().tolist()
```
asogaard, line 47: Here, and in conjunction with the comment regarding `LIMIT` above: We typically want these get-event-numbers-for-training functions to be random but reproducible, i.e. you should get the same (random) list of event numbers if you run this function multiple times _provided you use the same seed._ Therefore, I would propose to add an optional `seed` argument to this function, just as for the `get_equal_propoirtion_neutrino_indices` function, and use this argument in a similar manner as in that function to make the output reproducible.

asogaard (changes requested):
Hi @BozianuLeon,
Thanks so much for this contribution! 🥳   I think the binary classification parts are good to go; and I have added a few comments for the event-number-getter-function. I have provided a few concrete suggestions for naming and such that you should be able to accept or discard directly, hopefully making it easier to get the PR.

src/gnn_reco/data/sqlite_dataset.py
BozianuLeon: This change should only affect labels_dict where values are booleans, and in binary classification cases the target you aim for must be 0 or 1. However if the booleans are needed then bool(0) = False or bool(1)=True could be used to retrieve this.

BozianuLeon (commented):
Including random seed to make extraction random and reproducible with slightly more control over how many events you get out and with what PID contents.

asogaard (approved):
Thanks for the updates to this PR, @BozianuLeon! I think it looks good and am happy to approve. :)

PR #95, opened 2021-12-09
Vertex reconstruction, add tasks and constant

kaareendrup:
Added tasks for xyzt, xyz and time-only for vertex reconstruction. Also added interaction_time to list of constants.
Includes a few do-nothing commits that were just learning examples for yours truly.


asogaard (approved):
Looks great, @kaareendrup!

Issue #47, opened 2021-11-09
Train GNN on Upgrade MC


asogaard:
Is blocked by #46

asogaard:
As agreed at today's meeting I'll just tag @mhaminh. We agreed that Martin will focus on just getting the GNN predictions (using code outside of this repo) into PISA for the sensitivity studies. Once that it is in a good place, we Martin and I will discuss how to best updating the repo code accordingly to support these studies.

PR #96, opened 2021-12-09
Training on upgrade MC

asogaard:
Have added a script to train on mDOM data from Upgrade MC. The two main additions are:

1. Adding a `Detector` class for IceCube-Upgrade that preprocesses the new features.
2. Filtering events with < 2 DOM hits when batching. This is not an uncommon occurrence in the available Upgrade MC (probably all truth events are kept if they are reconstructed with any of the multiple available pulse maps).

Addresses #47 but doesn't close it quite yet.

Tagging @mhaminh for you information


RasmusOrsoe (commented):
Hi Andreas! Nice!

One question; why are we throwing away 2-pulse events? I think these would be extremely rare if we included all DOM types as input and not just the MDoms from your training example.

Rasmus

asogaard:
We're keeping 2-pulse events but throwing away 0-pulse events (ill-defined graph) and 1-pulse events (homophily or whatever it's called doesn't seem to work for just one node). And I agree, in practice this should not occur, so it's mainly just as a safeguard to make to it possible to run over the current Upgrade MC files.

RasmusOrsoe (approved):
Yes ofc! I was a bit tired when I went over it. Approved!

PR #99, opened 2021-12-14
Adding a few common GCD files

asogaard:
Is it fine to include files like these in Git, @philippeller? 

cc @BozianuLeon


philippeller:
Hmm, good question. I think this might in fact be proprietary IceCube information

asogaard:
Thanks, @philippeller! I had a suspicion something like that might be the case. I was also going to ask if we could package a small I3 file with the repo as test data, but I figure that would be even less kosher. :)  Closing the PR, then.

philippeller:
we cold maybe create an artificial test data file that resembles the IceCube data...?

PR #117, opened 2021-12-15
changed i3-file check to be compatible with lvl2

RasmusOrsoe:
changed i3-file check to be compatible with lvl2 filenaming. Without this fix, valid i3-files are overlooked in the is_i3 check.

fixed lvl2 noise, neutrino and muon extraction


asogaard (approved):
Looks good!

PR #118, opened 2021-12-15
Double bang classification

RasmusOrsoe:
Added and tested functionality : extraction of level2 double bang events.


asogaard (approved):
Looks good!

PR #121, opened 2021-12-21
upgrade hotfix

RasmusOrsoe:
Fixed small error that made processes fail at EOF and therefore leading to missing events in sql.


asogaard (approved):
Looks good!

PR #123, opened 2022-01-06
Bug fix(es)

asogaard:
This PR does two things:

1. In `src/graphnet/data/sqlite_dataset.py`, removes an unnecessary cast to float for all attributes on the graph (`Data`) object. This has been found by @kaareendrup to lead to discrepancies in the `event_no` attribute due to conversions of large integers to float32s and back, resulting in a potential misalignment between events during training and inference. This should be a **high priority** to get fixed.
2. In `src/graphnet/components/loss_functions.py`, switches to only evaluating the exact expression for `log_cmk` in the Von Mises-Fisher loss for valid values of `kappa`, instead of relying on `torch.swith(...)`. This removes a runtime warning that may interfere with the training.


asogaard:
With this PR in place (i.e. with the cast to float removed) all attributes on the `Data` object should have their original type, i.e. the one used in the SQLite database or one with which they are defined earlier in the `_create_graph` method. Specifically, `neutrino` and `track` are defined in the same method to both have type `int` (see https://github.com/icecube/graphnet/blob/main/src/graphnet/data/sqlite_dataset.py#L123-L135). I think it makes sense to cast the individual attributes to their appropriate types before creating the `Data` object, just as we were doing before. :)

RasmusOrsoe (approved):
Thanks.

Let's keep an eye on the labels after this merge. I've found in the past that torch.tensor(integer) sometimes unexpectedly casts integers to booleans when dtype is not explicitly set.

PR #125, opened 2022-01-10
Implementation of Inverse Transform at Inference Step

kaareendrup:
Implements Inverse Transformation by passing the transformation functions (e.g. `transform_prediction_and_target`) to the task instead of the loss function, and passing the not-instanced loss function class to the task function as shown in the training example (line 43). 

The task then instances the loss function with the proper transformation functions, but remembers the inference transformation for later use (task, line 47). The forward method of the task includes a transformation (line 58), which is initially set to a unit transformation (line 49).

In the `get_predictions.py` the model is set to inference mode, which simple calls a similar method for the task (model, line 136) which subsitutes the forward transformation for the inference function (task, line 71).

This way, it is only nessecary to specify the transformation functions in one place (when creating the task), and the framework will make sure to accept only `transform_prediction_and_target` or both `transform_inference` and `transform_target` or none of them.

### DIFF
--- a/examples/train_model.py
+++ b/examples/train_model.py
@@ -70,9 +70,8 @@ def main():
     task = EnergyReconstruction(
         hidden_size=gnn.nb_outputs,
         target_label=target,
-        loss_function=LogCoshLoss(
-            transform_prediction_and_target=torch.log10,
-        ),
+        loss_function=LogCoshLoss,
+        transform_prediction_and_target=torch.log10,
     )
     model = Model(
         detector=detector,
--- a/src/graphnet/models/model.py
+++ b/src/graphnet/models/model.py
@@ -132,3 +132,8 @@ class Model(LightningModule):
 
     def _get_batch_size(self, data: Data) -> int:
         return torch.numel(torch.unique(data.batch))
+
+    def inference(self):
+        """Sets model to inference mode."""
+        for task in self._tasks:
+            task.inference()
--- a/src/graphnet/models/task/task.py
+++ b/src/graphnet/models/task/task.py
@@ -1,5 +1,6 @@
 from abc import abstractmethod
 from typing import Union
+from typing import Callable, Optional
 
 try:
     from typing import final
@@ -22,14 +23,30 @@ class Task(LightningModule):
     def nb_inputs(self) -> int:
         """Number of inputs assumed by task."""
 
-    def __init__(self, hidden_size: int, target_label: str, loss_function: LossFunction):
+    def __init__(
+        self, 
+        hidden_size: int, 
+        target_label: str, 
+        loss_function: LossFunction,
+        transform_prediction_and_target: Optional[Callable] = None,
+        transform_target: Optional[Callable] = None,
+        transform_inference: Optional[Callable] = None,
+    ):
         # Base class constructor
         super().__init__()
+
+        # Check(s)
+        assert not((transform_prediction_and_target is not None) and (transform_target is not None)), \
+            "Please specify at most one of `transform_prediction_and_target` and `transform_target`"
+        assert (transform_target is not None) == (transform_inference is not None), \
+            "Please specify both `transform_inference` and `transform_target`"
 
         # Member variables
         self._regularisation_loss = None
         self._target_label = target_label
-        self._loss_function = loss_function
+        self._loss_function = loss_function(transform_prediction_and_target=transform_prediction_and_target, transform_target=transform_target) # Added
+        self._transform_inference = transform_inference
+        self._transform_forward = lambda x: x
 
         # Mapping from last hidden layer to required size of input
         self._affine = Linear(hidden_size, self.nb_inputs)
@@ -38,6 +55,7 @@ class Task(LightningModule):
     def forward(self, x: Union[Tensor, Data]) -> Union[Tensor, Data]:
         self._regularisation_loss = 0  # Reset
         x = self._affine(x)
+        x = self._transform_forward(x)
         return self._forward(x)
 
     @abstractmethod
@@ -49,3 +67,8 @@ class Task(LightningModule):
         target = data[self._target_label]
         loss = self._loss_function(pred, target) + self._regularisation_loss
         return loss
+
+    @final
+    def inference(self):
+        '''Set task to inference mode by substituting unitary forward transform with inference transform'''
+        self._transform_forward = self._transform_inference
--- a/misc/badges/pylint.svg
+++ b/misc/badges/pylint.svg
@@ -17,7 +17,7 @@
         <text x="22.0" y="14">pylint</text>
     </g>
     <g fill="#fff" text-anchor="middle" font-family="DejaVu Sans,Verdana,Geneva,sans-serif" font-size="11">
-        <text x="63.0" y="15" fill="#010101" fill-opacity=".3">6.51</text>
-        <text x="62.0" y="14">6.51</text>
+        <text x="63.0" y="15" fill="#010101" fill-opacity=".3">7.04</text>
+        <text x="62.0" y="14">7.04</text>
     </g>
 </svg>
--- a/src/graphnet/models/training/utils.py
+++ b/src/graphnet/models/training/utils.py
@@ -110,6 +110,9 @@ def get_predictions(trainer, model, dataloader, prediction_columns, additional_a
         additional_attributes = []
     assert isinstance(additional_attributes, list)
 
+    # Set model to inference mode
+    model.inference()
+
     # Get predictions
     predictions_torch = trainer.predict(model, dataloader)
     predictions = [p[0].detach().cpu().numpy() for p in predictions_torch]  # Assuming single task


examples/train_model.py
```diff
@@ -70,9 +70,8 @@ def main():
     task = EnergyReconstruction(
         hidden_size=gnn.nb_outputs,
         target_label=target,
-        loss_function=LogCoshLoss(
```
asogaard, line 73: My immediate thought is that it might be a bit confusing for both the `LossFunction` class and the `Task` class to support `transform_*` arguments, and the latter passing them on to the former. I like the idea, as proposed in this PR, to let the `Task` class handle all of the transforms. Then the `LossFunction` class is simpler, and just handles the actual computation of the loss. I am going to propose edits in this vein below, and then we can discuss if that is the path we'd like to take. If so, this PR should also update the `LossFunction` class accordingly.

examples/train_model.py
```diff
@@ -70,9 +70,8 @@ def main():
...
-            transform_prediction_and_target=torch.log10,
-        ),
+        loss_function=LogCoshLoss,
+        transform_prediction_and_target=torch.log10,
```
asogaard, line 74:
```suggestion
        loss_function=LogCoshLoss(),
```

src/graphnet/models/model.py
```diff
@@ -132,3 +132,8 @@ def load_state_dict(self, path: str) -> 'Model':  # pylint: disable=arguments-di
 
     def _get_batch_size(self, data: Data) -> int:
         return torch.numel(torch.unique(data.batch))
+
+    def inference(self):
```
asogaard, line 136: I figure we should also have functionality to switch back from inference mode to train/eval mode. This could be done by overriding the pytorch built-in `.train()` and `.eval()` methods (see https://pytorch.org/docs/stable/generated/torch.nn.Module.html#torch.nn.Module.train and https://pytorch.org/docs/stable/generated/torch.nn.Module.html#torch.nn.Module.eval) to switch the internal inference "flag" to "off", and then calling the corresponding base class method (i.e. `super().train()` and `super().eval()`).

src/graphnet/models/task/task.py
```diff
@@ -22,14 +23,30 @@ class Task(LightningModule):
...
+        assert not((transform_prediction_and_target is not None) and (transform_target is not None)), \
+            "Please specify at most one of `transform_prediction_and_target` and `transform_target`"
+        assert (transform_target is not None) == (transform_inference is not None), \
+            "Please specify both `transform_inference` and `transform_target`"
+
         # Member variables
         self._regularisation_loss = None
         self._target_label = target_label
-        self._loss_function = loss_function
+        self._loss_function = loss_function(transform_prediction_and_target=transform_prediction_and_target, transform_target=transform_target) # Added
```
asogaard, line 47:
```suggestion
        self._loss_function = loss_function
```

src/graphnet/models/task/task.py
```diff
@@ -22,14 +23,30 @@ class Task(LightningModule):
...
+        self._transform_inference = transform_inference
+        self._transform_forward = lambda x: x
```
asogaard, line 50:
```suggestion
        self._inference = False
```

src/graphnet/models/task/task.py
```diff
@@ -22,14 +23,30 @@ class Task(LightningModule):
...
```
asogaard, line 49:
```suggestion
        if transform_prediction_and_target is not None:
            self._transform_prediction_training = transform_prediction_and_target
            self._transform_prediction_inference = lambda x: x
            self._transform_target = transform_prediction_and_target
        elif transform_target is not None:
            self._transform_prediction_training = lambda x: x
            self._transform_prediction_inference = transform_inference
            self._transform_target = transform_target
        else:
            self._transform_prediction_training = lambda x: x
            self._transform_prediction_inference = lambda x: x
            self._transform_target = lambda x: x
```

src/graphnet/models/task/task.py
```diff
@@ -38,6 +55,7 @@ def __init__(self, hidden_size: int, target_label: str, loss_function: LossFunct
     def forward(self, x: Union[Tensor, Data]) -> Union[Tensor, Data]:
         self._regularisation_loss = 0  # Reset
         x = self._affine(x)
+        x = self._transform_forward(x)
```
asogaard, line 58:
```suggestion
        x = self._transform_prediction(x)
```

src/graphnet/models/task/task.py
```diff
@@ -49,3 +67,8 @@ def compute_loss(self, pred: Union[Tensor, Data], data: Data) -> Tensor:
         target = data[self._target_label]
         loss = self._loss_function(pred, target) + self._regularisation_loss
         return loss
+
+    @final
+    def inference(self):
+        '''Set task to inference mode by substituting unitary forward transform with inference transform'''
+        self._transform_forward = self._transform_inference
```
asogaard, line 74:
```suggestion
        self._inference = True
```

src/graphnet/models/task/task.py
```diff
@@ -49,3 +67,8 @@ def compute_loss(self, pred: Union[Tensor, Data], data: Data) -> Tensor:
         target = data[self._target_label]
```
asogaard, line 67:
```suggestion
        target = data[self._target_label]
        target = self._transform_target(target)
```

src/graphnet/models/task/task.py
```diff
@@ -38,6 +55,7 @@ def __init__(self, hidden_size: int, target_label: str, loss_function: LossFunct
     def forward(self, x: Union[Tensor, Data]) -> Union[Tensor, Data]:
         self._regularisation_loss = 0  # Reset
         x = self._affine(x)
+        x = self._transform_forward(x)
         return self._forward(x)
```
asogaard, line 60:
```suggestion
    @final
    def _transform_prediction(self, prediction: Union[Tensor, Data]) -> Union[Tensor, Data]:
        if self._inference:
            return self._transform_prediction_inference(prediction)
        else:
            return self._transform_prediction_training(prediction)
```

src/graphnet/models/task/task.py
```diff
@@ -22,14 +23,30 @@ class Task(LightningModule):
...
+    ):
         # Base class constructor
         super().__init__()
 
+        # Check(s)
+        assert not((transform_prediction_and_target is not None) and (transform_target is not None)), \
+            "Please specify at most one of `transform_prediction_and_target` and `transform_target`"
+        assert (transform_target is not None) == (transform_inference is not None), \
+            "Please specify both `transform_inference` and `transform_target`"
+
```
asogaard, line 43: Maybe here it would be good to add a sanity check that `transform_target` and `transform_inference` are indeed inverse? Could potentially help catch a few bugs.
```suggestion
        if transform_target is not None:
            x_test = np.logspace(-6, 6, 12 + 1)
            x_test = torch.from_numpy(np.concatenate([-x_test[::-1], [0], x_test]))
            t_test = transform_target(x_test)
            valid = torch.isfinite(t_test)
            assert torch.allclose(transform_inference(t_test)[valid], x_test[valid]), \
                "The provided transforms for targets during training and predictions during inference at not inverse."
            del x_test, t_test, valid
```

asogaard (commented):
Hi @kaareendrup!
Thanks for this very nice PR! 🥳   As you asked, I have tried to be pretty thorough and think about was to restructure the code slightly to hopefully make it a bit more transparent. I have tried to include specific suggestions (NB: No guarantees wrt. syntax and spelling) that you are free to include if you agree. 

The main suggestion I have is to keep all transforms within the `Task` class, remove them from the `LossFunction` class, and make the `inference` functionality a bit more complete.

Once we have converged on a structure that we're happy with, ideally this PR should also update the example scripts that use the `Task` and `LossFunction` classes (such that examples in the `main` branch can always be run out of the box), and include unit test(s) of the new functionality.

src/graphnet/models/task/task.py
asogaard: Oops, the `self._transform_forward` should be renamed to `self._transform_prediction`, or the other way around, if the suggested change is accepted. Otherwise the names would disagree.

src/graphnet/models/task/task.py
kaareendrup: This is clever, I like it

src/graphnet/models/model.py
kaareendrup: Unfortunately, I was unable to implement this completely, since `.eval()` is called by pytorch when predicting from somewhere deep inside the framework that we can't easily override. (I have a dump of the exact location and the function calls that lead to it.)

But I think the idea is good enough to still be valid only for `.train()`, so I have implemented it there.

asogaard: I would be interested to better understand the differences between `.train()` and `.eval()` but for the purposes of this PR I think the proposed change is perfectly fine.

src/graphnet/models/task/task.py
```diff
@@ -22,14 +25,60 @@ class Task(LightningModule):
...
+    def __init__(
+        self, 
+        hidden_size: int, 
+        target_label: str, 
+        loss_function: LossFunction,
+        transform_prediction_and_target: Optional[Callable] = None,
+        transform_target: Optional[Callable] = None,
+        transform_inference: Optional[Callable] = None,
+        transform_support: tuple = None,
+    ):
```
asogaard, line 37: Considering how complex the `Task` constructor has become, I think it requires a docstring to explain especially the `transform_*` arguments.

```suggestion
    ):
    """Base class for all reconstruction and classification tasks.

    Args:
        hidden_size: The number of nodes in the layer feeding into this tasks,
          used to construct the affine transformation to the predicted quantity.
        target_label: Name of the quantity being predicted, used to extract the
          target tensor from the `Data` object in `.compute_loss(...)`.
        loss_function: Loss function appropriate to the task.
        transform_prediction_and_target: Optional function to transform both the
          predicted and target tensor before passing them to the loss function.
          Useful e.g. for having the model predict quantities on a physical
          scale, but transforming this scale to O(1) for a numerically stable
          loss computation.
        transform_target: Optional function to transform only the target tensor
          before passing it, and the predicted tensor, to the loss function.
          Useful e.g. for having the model predict a transformed version of the
          target quantity, e.g. the log10-scaled energy, rather than the
          physical quantity itself. Used in conjunction with `transform_inference`
          to perform the inverse transform on the predicted quantity to recover
          the physical scale.
        transform_inference: Optional function to inverse-transform the model
          prediction to recover a physical scale. Used in conjunction with
          `transform_target`.
        transform_support: Optional tuple to specify minimum and maximum of the
          range of validity for the inverse transforms `transform_target` and
          `transform_inference` in case this is restricted. By default the
          invertibility of `transform_target` is tested on the range [-1e6, 1e6].
    """
```

src/graphnet/models/task/task.py
```diff
@@ -22,14 +25,60 @@ class Task(LightningModule):
...
-    def __init__(self, hidden_size: int, target_label: str, loss_function: LossFunction):
+    def __init__(
+        self, 
+        hidden_size: int, 
+        target_label: str, 
+        loss_function: LossFunction,
+        transform_prediction_and_target: Optional[Callable] = None,
+        transform_target: Optional[Callable] = None,
+        transform_inference: Optional[Callable] = None,
+        transform_support: tuple = None,
```
asogaard, line 36:
```suggestion
        transform_support: Optional[tuple] = None,
```

src/graphnet/models/task/task.py
```diff
@@ -22,14 +25,60 @@ class Task(LightningModule):
...
+    ):
         # Base class constructor
         super().__init__()
 
+        # Check(s)
```
asogaard, line 41: Given how complex checking the `transform_*` argument have become, I would perhaps suggest writing just
```python
self._transform_prediction_training = None
self._transform_prediction_inference = None
self._transform_target = None
self._validate_and_set_transforms(
    transform_prediction_and_target,
    transform_target,
    transform_inference,
    transform_support,
)
```
in the constructor and then moving the current functionality to a dedicated utility method.

tests/test_task.py
```diff
@@ -0,0 +1,44 @@
...
+import torch
+
+from graphnet.models.detector.icecube import IceCube86
+from graphnet.models.graph_builders import KNNGraphBuilder
+from graphnet.models.gnn import DynEdge
+from graphnet.models.task.reconstruction import EnergyReconstruction
+from graphnet.components.loss_functions import LogCoshLoss
+
+
+def test_transform_prediction_and_target():
```
asogaard, line 11: These seem alright, so if you have checked them with `pytest` I'm all for it! :)

asogaard (changes requested):
Hi @kaareendrup,

Thanks for a very nice update on this PR! 🥳   I think it is in very good shape. I have made just a few concrete suggestions to improve documentation, and then made one request for moving some quite involved checks into a separate utility method.

Let me know what you think about the above; at any rate, I think this is ready to merge in no time.

examples/train_model.py
```diff
@@ -78,10 +78,9 @@ def main():
     )
     task = EnergyReconstruction(
         hidden_size=gnn.nb_outputs,
-        target_label=config["target"],
-        loss_function=LogCoshLoss(
-            transform_prediction_and_target=torch.log10,
-        ),
+        target_label=target,
```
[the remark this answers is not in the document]
kaareendrup: It does! I think when writing this, the wandb stuff wasn't implementen, maybe my novice github skills are at fault here

tests/test_task.py
kaareendrup: I sure have!

src/graphnet/models/task/task.py
kaareendrup: Very good, this has been done and added to the PR!

asogaard (approved):
Woohoo! 🥳

PR #141, opened 2022-01-14
Train on Upgrade MC

### LINKED ISSUE #47: Train GNN on Upgrade MC

### LINKED ISSUE #112: How should we represent/record standard dataset?
In the interest of reproducibility, benchmarking, and such it could be useful to have simple "cards" that unambiguously define various benchmark datasets. This could be in the form of a simple structure like:
```
datasets/
|-- upgrade-sensitivity-mc.yml
|-- deepcore-doublebang-mc.yml
|-- deepcore-oscillations-2021-data.yml
`-- (...)
```
with each YAML file containing little more than path(s) to I3 files on Cobalt.

What do y'all think?

asogaard:
It's been a long time coming, but this PR adds functionality to process data from IceCube Upgrade MC I3-files, and train GNN models on this data. Biggest changes:

* Adds first examples of a dataset card in `datasets/datasets/upgrade_neutrino_reconstruction.yml` as a way to make it transparent which input files are intended/used for each analysis.
* Adds example script `examples/plot_feature_distributions.py` to make it easy to plot, well, feature distributions from SQLite databases. Useful for feature engineering and normalisation.
* Updates the `SQLiteDataset` class to accept one _or more_ pulsemaps when extracting graphs. This is useful for collating pulses from PDOMs, mDOMs, and DEggs in cases where these are store separately. A single pulsemap can of course still be specified.
* Adds a `DynEdge_V2` class in`src/graphnet/models/gnn/dynedge.py`. This is very similar to `DynEdge` except that it adds global features (i.e. mean-pooled node feature values, homophily values, and the number of pulses in the event) as feature on each node at the start of the forward pass, and not at the last layer after global pooling. The intuition is that "informing" the network about the number of pulses, average DOM charge, etc. early should make it better across very different event sizes. This was prompted by the training on Upgrade MC which has 50% of events with <= 10 pulses, and events with > 1000 pulses. This class also drops the sum-aggregation during global pooling, as this is expected to yield values at very different numerical scales than the other (min, max, mean) pooling methods. This all seems to be working slightly better than the nominal architecture on this specific task, see the W&B project below. 
* Adds scripts to `studies/upgrade_neutrino_reconstruction/` for training the model and plotting the resulting resolutions, initially for zenith.

Training runs can be found in W&B here: https://wandb.ai/graphnet-team/upgrade-zenith

Closes #47, closes #112 

Tagging @BozianuLeon and @kaareendrup for your information


RasmusOrsoe (approved):
Nice PR!

Issue #143, opened 2022-01-20
Force target to have feature dimension


kaareendrup:
This issue is a suggestion to add a feature dimension to target tensors and update all loss/task functions etc. to adopt this functionality. 

Right now, the model output has a feature dimension, while the target does not, and the forward call of a typical loss function (here `LogCoshLoss`) includes this syntax:
`assert prediction.dim() == target.dim() + 1`
 `diff = prediction[:,0] - target`

The background for adding the feature dimension to the target, is that this standard syntax in sklearn, which causes a problem when working with more complex transformation functions. When applying a function like `sklearn.preprocessing.QuantileTransformer` to the target, it is necessary to add a dimension before transformation and remove it again afterwards. This is not a major problem (though it does add room for mistakes), but it causes further challenges when doing things like the sanity check in #125 which asserts that the target and inference transformations are inverse.

These points cause me to believe that the developers of sklearn have a good reason for choosing this syntax as standard, and after discussion with @asogaard, I believe that our code will be simpler and more robust if we adopt it.

PR #144, opened 2022-01-20
Update classification to work with the new predictor class

kaareendrup:
The current binary classification task and loss function do not work with the new predictor class, because the feature dimension is removed in the task instead of the in the loss function. This PR changes this and adds a unit test to the BCE loss class. 

It also changes the truth values in the dataset class from dtype=int to dtype=float, since right now when doing classification, they are converted to dtype=long instead. The line of thinking is that they will need to be floats at some point anyway to be of use in a model, so converting them at this step will do no harm.


asogaard:
Thanks for the PR, @kaareendrup! I think this gets at a fundamental question that I think we should agree on before proceeding:

* Should we require that all targets and all predictions have exactly two dimensions (nb_examples x nb_features/nb_target, resp.), similar to scikit-learn? Because right now, as @kaareendrup correctly points out, e.g. the `BinaryClassificationTask` removes the second dimension.

This is also something @RasmusOrsoe has recently had to contend with in #145, where he went with reshaping to 2D. I would probably be in favour doing just that throughout, and be consistent in our tensor shapes (i.e. not having some tasks returning 1D tensors, and others 2D) as well as being consistent with other, common ML libraries that users/developers are likely familiar with (scikit-learn) — but I also want to hear what @RasmusOrsoe et al. think.

To you second point, @kaareendrup, I don't think we should cast the flag-type attributes (e.g., `muon`, `noise`) to floats right away. They represent booleans, so that type or some non-floating-point type like int/long is a natural representation, and if we want to evaluate exact expressions (e.g., `mask = (data.muon == 1)`) needing to convert back to int/float seems like an unnecessary risk, cf. #123. I completely see your point that, when doing classification and calculating the BCE, these quantities need to be cast to float, but then I think it would be more intuitive to do that when we need it, rather than at some point upstream, i.e.:
```
torch.nn.functional.binary_cross_entropy(prediction, target.float(), reduction='none')
```

asogaard:
See also #143 !

RasmusOrsoe:
Hi! 

I do not fully understand what the float() conversion is intended to fix. I have no issues doing binary classification with the current version of the repo if the index error of the assert statement is fixed as proposed in #145. 

I agree that float conversion should be done in specific tasks when required and not in the dataset.


Rasmus

asogaard:
Have approved @RasmusOrsoe's PR #145 since it works, and any change to how we shape tensors can be made after it is merged.

But I think the fact we're having to catch assert exceptions because the shapes of tensors is not predictable should probably be considered a temporary fix to an underlying design flaw. :)  What if we flip it: Would anyone **object** to the convention that our models should always return 2D tensor and that our targets should always be 2D as well? If so, I can make a PR to implement this change (with no effect on what the models actually do), to address #143 .

RasmusOrsoe:
@asogaard I think that sounds like a fine idea. Lets just make sure that the implementation of this convention doesn't break the currently implemented tasks!

Rasmus

kaareendrup:
Thanks guys!

As you can probably tell from #143, I also think having 2D targets and predictions at all times is a good idea, and I'm glad that you agree. As it seems to be a rather fundamental change, I will definitely feel more comfortable if @asogaard wants to implement it (rather than myself). 

And if #145 fixes whatever problems this may cause until then, I will be happy to close this PR, as I agree with your points about the float cast

PR #145, opened 2022-01-25
mixed beans

RasmusOrsoe:
Added fix to dimension bug in `get_predictions`. The old assert would fail on squeezed dataframes e.g. df.shape = (n,)

Added dbang label maker that accepts decay length

Added true dbang label in dataset class.


asogaard:
I suggest we close #144 first, as that might affect one of the changes in this PR.

asogaard (approved):
👍

PR #148, opened 2022-01-27
Stopmuon func in utils, add labels to i3extractor

BozianuLeon:
Finally adding functionality to calculate finishing point of muons, using starting position, track length, azimuth and zenith. 
Would like to remove hard-coded border-coords using GCD file already opened(?). Will require recreating databases.


src/graphnet/data/utils.py
```diff
@@ -249,3 +250,31 @@ def frame_has_key(frame, key: str):
...
+    border_z = [-512.82,524.56]
+    border = mpath.Path(border_coords)
```
RasmusOrsoe, line 264: Could we refactor this and perhaps change the variable naming? 

It looks like  ` border_coords` is indeed xy-coordinates, and if this is true, then lets please rename to `border_xy`.

It would be nice if we could add a single input parameter `borders` to the truth extractor that defaults to this choice.

We could save the border input variable as `self._borders` and then 

```
if self._borders == None:
  border_xy = np.array([(-256.1400146484375, -521.0800170898438), (-132.8000030517578, -501.45001220703125), (-9.13000011444092, -481.739990234375), (114.38999938964844, -461.989990234375), (237.77999877929688, -442.4200134277344), (361.0, -422.8299865722656), (405.8299865722656, -306.3800048828125), (443.6000061035156, -194.16000366210938), (500.42999267578125, -58.45000076293945), (544.0700073242188, 55.88999938964844), (576.3699951171875, 170.9199981689453), (505.2699890136719, 257.8800048828125), (429.760009765625, 351.0199890136719), (338.44000244140625, 463.7200012207031), (224.5800018310547, 432.3500061035156), (101.04000091552734, 412.7900085449219), (22.11000061035156, 509.5), (-101.05999755859375, 490.2200012207031), (-224.08999633789062, 470.8599853515625), (-347.8800048828125, 451.5199890136719), (-392.3800048828125, 334.239990234375), (-437.0400085449219, 217.8000030517578), (-481.6000061035156, 101.38999938964844), (-526.6300048828125, -15.60000038146973), (-570.9000244140625, -125.13999938964844), (-492.42999267578125, -230.16000366210938), (-413.4599914550781, -327.2699890136719), (-334.79998779296875, -424.5)])
  border_z = [-512.82,524.56]
  border = mpath.Path(border_xy)
else:
     border_xy = borders[0]
     border_z   = borders[1]
```

RasmusOrsoe (changes requested):
Hi Leon! Nice Job!

Have added comments.

src/graphnet/data/utils.py
BozianuLeon: Completely agree, although I left the path.Path to the utils function so no need to import matplotlib in i3extractor.

RasmusOrsoe (approved):
Looks good! Thanks @BozianuLeon

asogaard:
Hi @RasmusOrsoe and @BozianuLeon,

I think this PR is a very nice contribution, and I am happy to see it merged! :)

However, I would point out that several automated checks failed (see the "Checks" tab above), which means that following this PR, in principle, you cannot install and run this package out of the box on a clean system; something I figure we should otherwise strive for. (It also means that all subsequent PRs will report failing automated checks, potentially obscuring other errors, until the issue is resolved.)

The problem is trivial — `matplotlib` is imported in the code, but not added to `setup.py` as a dependency — so I think we should just try to get a slim PR in to fix this issue as soon as possible. This is mainly just to point out the purpose of these pesky li'l checks. :)

troelspetersen:
Hej Andreas

Jeg er helt enig (og lidt vild med) disse checks, idet vi virkelig gerne vil have, at man kan køre det “out of the box”.

Af samme grund, så synes jeg, at vi skal se på Mortens “case”, idet han fik en masse fejl. Det er nok ikke pga. pakken, men nok snarere fra hans/systemets side… men det er stadig god læring, for han er forhåbentlig ikke den sidste, som checker det ud, men desværre nok heller ikke den sidste, som løber ind i en eller anden form for problemer (han havde gcc probs).


	BH. Troels

BozianuLeon:
Yes, of course. Thanks for the heads up Andreas. If no one else has done this I will go ahead and add a matplotlib dependency in setup.py. (also good to know for future more complicated check failures)

PR #151, opened 2022-01-29
Radialgraphbuild

BozianuLeon:
Simple alternative to knn imagining a sphere around each triggered DOM, taken identically from torch_geometric.nn. (Hopefully this time without additional commits)


RasmusOrsoe (approved):
Well done!

PR #153, opened 2022-02-01
Upgrade noise pulsemap

RasmusOrsoe:
Contains  new extractor class for pulsemap noise truth

### DIFF
--- a/src/graphnet/data/i3extractor.py
+++ b/src/graphnet/data/i3extractor.py
@@ -195,6 +195,50 @@ class I3FeatureExtractorIceCubeUpgrade(I3FeatureExtractorIceCube86):
         output.update(output_icecube86)
         return output
 
+class I3PulseNoiseTruthFlagIceCubeUpgrade(I3FeatureExtractorIceCube86):
+
+    def __call__(self, frame) -> dict:
+        """Extract features to be used as inputs to GNN models."""
+
+        output = {
+            'string': [],
+            'pmt_number': [],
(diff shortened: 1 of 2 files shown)


src/graphnet/data/i3extractor.py
```diff
@@ -195,6 +195,50 @@ def __call__(self, frame) -> dict:
...
+            pmt_dir_y = self._gcd_dict[om_key].orientation.y
+            pmt_dir_z = self._gcd_dict[om_key].orientation.z
+            string = om_key[0]
+            dom_number = om_key[1]
+            pmt_number = om_key[2]
+            dom_type = self._gcd_dict[om_key].omtype
+
+            # Loop over pulses for each OM
+            pulses = data[om_key]
+            for truth_flag in pulses:
```
asogaard, line 231: I am not sure I understand how we can extract charge, etc. in the `I3FeatureExtractorIceCube86` base class using a pulsemap like `'SplitInIcePulses_GraphSage_Pulses'` which here seems to just return either 0 or 1 — but, hey, if it works it works. :)

asogaard (approved):

src/graphnet/data/i3extractor.py
RasmusOrsoe: Hi. You are right. This is just for the truth flags.

SplitInIcePulses_GraphSage_Pulses is extracted using the normal upgrade feature extractor as its a regular puslemap

PR #154, opened 2022-02-01
functionality from paper branch to make upgrade sensitivity studies

RasmusOrsoe:
functionality from paper branch to make upgrade sensitivity studies

Sorry for main branch commit. Was in a hurry


asogaard (approved):
As you mentioned to me, @RasmusOrsoe, there is a lot of overlap here with the work in 
#125, but in the intereste of fast turn-around I am happy to merge it now and then resolve these redundancies later.

PR #156, opened 2022-02-01
added data-object and target_label as input to loss functions

RasmusOrsoe:
Hi!

This PR lets loss functions recieve the full `torch_geometric.data.Data-object` and the `target_label`.  Before, loss functions only had the prediction and the truth available.

The reason behind this proposed change is that having the full Data object available allows us to use truth variables to calculate weights inside the loss function we make.

I have updated all current loss functions such that they work with this change.

Rasmus


RasmusOrsoe:
Deleting this because its now redundant. Tagging @asogaard  fyi

PR #157, opened 2022-02-02
Tidyness

RasmusOrsoe:
removed redundant code introduced in earlier PR.

Added general 'PassOutput' tasks for those jobs where we simply want to do nothing.

Rasmus


asogaard:
Thanks for this comprehensive PR! I think a lot of it looks brilliant (e.g. the generic pass-through tasks; the Euclidean graph builder (cc @BozianuLeon); and the general tidying up. :) 

One thing I'd like to discuss though, is passing two additional arguments — the `Data` object and the `target_label` string — to tasks and loss functions. First, I don't see how the `target_label` is being used in the PR. Second, hard-coding the name of the weight attribute in a specialised loss function like `DBangBCELossWeighted`, I think, is a slippery slope towards a lot off single-purpose classes.

I completely agree that it is crucial that we can specify per-event weights, it's just a matter of how we best implement it. I think the simplest would be to have the pytorch `Dataset`-like class (e.g. `SQLiteDataset`) optionally setting the `sample_weight` attribute on the `Data` object (borrowing Keras' naming convention [here](https://www.tensorflow.org/api_docs/python/tf/keras/Model#fit); and then within `Task.compute_loss` checking if this attribute is set and, if it is, pass it on to the `LossFunction`. Then we need only add a single, optional, `Tensor`-type argument (`sample_weight`) to the loss functions, and we can keep using general loss functions. Let me know what you think. :) 

(NB: This looks like it supersedes #156 so I will just comment here — let me know if that is a misunderstanding.)

RasmusOrsoe:
Hi!

I agree that this could lead to many dead loss functions. I'm fine with having this changed to a different method - perhaps the one you mentioned. Maybe we could accept this now as-is and then open an issue to restructure it later? Then we'll have the functionality immediately

Rasmus

RasmusOrsoe:
@asogaard made the changes we discussed

asogaard (approved):
🙌

PR #158, opened 2022-02-02
Allow for multiple task labels, e.g. for vertex reconstruction

asogaard:
Make sure that vertex reconstruction is possible out of the box by allowing each task to accept multiple task labels, that are concatenated. Have also added the script to run all four types of trainings for the upgrade sensitivity studies, but this ought to be rewritten to use the transform-functionality added through #125 

cc @kaareendrup


RasmusOrsoe (approved):
Nice! Good with the str check!

PR #161, opened 2022-02-07
added noise upgrade study

RasmusOrsoe:
Added upgrade noise study code for @asogaard @BozianuLeon


asogaard:
Great to have these studies documented! It seems like there is a fair bit of overlap with the one existing study; at least https://github.com/icecube/graphnet/blob/main/studies/upgrade_neutrino_reconstruction/modelling/train_models.py  Should we remove this file and just use what's in this PR?

RasmusOrsoe:
@asogaard file removed!

asogaard (approved):
👍

PR #167, opened 2022-02-10
Update BC task to fit new 2D regime

kaareendrup:
Removes the cast of the prediction to 1 dimension in the binary classification task


asogaard (approved):
🙌

PR #168, opened 2022-02-11
two changes to i3 extractor

RasmusOrsoe:
Hi! This contain two changes

1.  Update of `frame_is_noise`
2.  Removal of `del frame["I3Calibration"]  # Avoid modifying the frame in-place`

About 1:
The old version relied on certain i3 keys to determine pure noise events which weren't robust. In fact, a bug related to these fields for a special piece of MC triggered this change.

About 2:
When  removing` frame["I3Calibration"]` we automatically set `rde = -1` for all pulses in that event. (That is never good).  

```
def _get_om_keys_and_pulseseries(self, frame):
        """Gets the indicies for the gcd_dict and the pulse series

        Args:
            frame (i3 physics frame): i3 physics frame

        Returns:
            om_keys (index): the indicies for the gcd_dict
            data    (??)   : the pulse series
        """
        data = frame[self._pulsemap]
        try:
            om_keys = data.keys()
        except:
            try:
                if "I3Calibration" in frame.keys():
                    data = frame[self._pulsemap].apply(frame)
                    om_keys = data.keys()
                else:
                    frame["I3Calibration"] = self._calibration
                    data = frame[self._pulsemap].apply(frame)
                    om_keys = data.keys()
                    #del frame["I3Calibration"]  # Avoid modifying the frame in-place
            except:
                data = dataclasses.I3RecoPulseSeriesMap.from_frame(frame, self._pulsemap)
                om_keys = data.keys()
        return om_keys, data
```
This doesn't seem to be an issue for level7 and level2 and level4 because the logic never makes it to the `else` statement that contains this delete statement. At least that is my theory - I've checked level7 and level2 events and found no `rde= -1` cases. We need to do try-catch of different errors because the way in which the pulses are stored in the i3 frames is not consistent between MC in IceCube. Therefore some sets trigger different logic in the code above. So the `rde = -1` bug only trigger for those cases that ends up in the else-statement. 

I don't think there is any reason why `del frame["I3Calibration"] ` is required, so an easy fix is to remove it. I believe @asogaard  added that delete statement a while back, perhaps you can comment on this?

Rasmus

### DIFF
--- a/src/graphnet/data/i3extractor.py
+++ b/src/graphnet/data/i3extractor.py
@@ -93,7 +93,7 @@ class I3FeatureExtractor(I3Extractor):
                     frame["I3Calibration"] = self._calibration
                     data = frame[self._pulsemap].apply(frame)
                     om_keys = data.keys()
-                    #del frame["I3Calibration"]  # Avoid modifying the frame in-place
+                    del frame["I3Calibration"]  # Avoid adding unneccesary data to frame
             except:
                 data = dataclasses.I3RecoPulseSeriesMap.from_frame(frame, self._pulsemap)
                 om_keys = data.keys()
@@ -127,10 +127,7 @@ class I3FeatureExtractorIceCube86(I3FeatureExtractor):
             y = self._gcd_dict[om_key].position.y
             z = self._gcd_dict[om_key].position.z
             area = self._gcd_dict[om_key].area
-            if "I3Calibration" in frame:  # Not available for e.g. mDOMs in IceCube Upgrade
-                rde = frame["I3Calibration"].dom_cal[om_key].relative_dom_eff
-            else:
-                rde = -1.
+            rde = self._get_relative_dom_efficiency(frame, om_key)
 
             # Loop over pulses for each OM
             pulses = data[om_key]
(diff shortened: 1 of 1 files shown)


asogaard:
Hi @RasmusOrsoe,

Thanks for this PR! To your two items:

1. 👍 
2. I see your point about why this will lead to `rde=-1` in cases that reach the `else`-statement, and that should of course be fixed! For context, I introduced the `del ...` statement when developing the `I3Module` for applying the GNNs in reconstruction chains. The `I3Calibration` object can be large, and if we add it to the frame, that data gets passed on, down the reconstruction chain, and may result in unnecessarily large output I3-files. I think the intention of not adding any additional unnecessary data to the frame object should be kept, but you're completely right that thet `rde=-1` problem itself needs to be fixed. Right now, the `_get_om_keys_and_pulseseries` does what it says on the tin, is self-contained, and doesn't add additional data to the frame object, so I don't think we should remove the `del ...` statement. Rather, I think we could extend the check here https://github.com/icecube/graphnet/blob/main/src/graphnet/data/i3extractor.py#L130  to fall back to using `self._calibration` if the `"I3Calibration"` key is not set on the frame. Same procedure as is used in `_get_om_keys_and_pulseseries` but applied where it is needed. This could be wrapped in a `_get_relative_dom_efficiency(self, frame, om_key)` method to make it a little neater. Let me know what you think. :)

RasmusOrsoe:
I've added a `_get_relative_dom_efficiency` method and tested that it works. 

@asogaard

asogaard (approved):
**Chef's kiss**

PR #170, opened 2022-02-15
Deploying upgrade models

### LINKED ISSUE #48: Write GNN predictions to Upgrade I3 files
Is blocked by #46 and #47

asogaard:
This PR:
* Renames and updates `graphnet_module.py` to provide I3Module-inheriting classes for all IceCube detector configurations for deploying GNN models.
* Provides example script (`process_i3_files.py`) for using these modules to process I3-files and save predictions.
* Provides a utility script for loading models from state-dict and then saving the entire, clean model to file, thereby allowing it to be loaded separately from model building. (Currently, saving a pytorch lightning module after training results in references to the used dataset/dataloader being saved with the model, which renders loading the entire model in a separate setting impossible. This ought to be investigated separately.)

Finally, this PR also adds a first (and non-functioning) attempt at a script for converting models to ONNX, which would allow for radically simpler deployment/portability. 

Closes #48


RasmusOrsoe (approved):
Nice!

Issue #91, opened 2021-11-30
Implement a Euclidean graph builder


asogaard:
See https://github.com/icecube/gnn-reco/blob/main/src/gnn_reco/models/graph_builders.py

BozianuLeon:
Hi Andreas I know I have spoken about looking into KNN alternatives. I'd like to tackle this issue and maybe look at other graph builders. Do you have a link to the Euclidean graph builder paper you mentioned previously. Thanks

asogaard:
Hi @BozianuLeon!
That sounds brilliant. 👏   You could look to https://arxiv.org/abs/1809.06166 for inspiration. Sec. IV, item (b) describes how they construct graphs, where `d_{ij}` is the Gaussian kernel over the Euclidean distance between DOMs `i` and `j`, and `a_{ij}` is the normalised node adjacency based on this measure. Let me know if we should have a chat about implementation details. :)

PR #172, opened 2022-02-18
Euclidean Graph Builder w/ weights

BozianuLeon:
Long time coming, first graph builder that also outputs tensor of edge_weights, following procedure from https://arxiv.org/pdf/1809.06166.pdf, and closes issue #91. Potential follow up includes allowing model to infer learnable scalar parameter sigma and possible improvements to batch monitoring.


asogaard:
Hi @BozianuLeon,
Thanks for this highly anticipated PR! 😁   It actually does two things, right: Adding both the radius-based and the Euclidean graph builder. I'll review it right away and try to add as many pedagogical comments/suggestions as I can — hopefully they can be useful!

asogaard:
(Just resolved a merge conflict with the above commit.)

src/graphnet/models/utils.py
```diff
@@ -15,3 +16,18 @@ def calculate_xyzt_homophily(x, edge_index, batch):
     hz = homophily(edge_index, x[:,2], batch).reshape(-1,1)
     ht = homophily(edge_index, x[:,3], batch).reshape(-1,1)
     return hx, hy, hz, ht
+
+
+def calculate_distance_matrix(xyz_coords):
```
asogaard, line 21:
```suggestion
def calculate_distance_matrix(xyz_coords: Tensor) -> Tensor:
```

src/graphnet/models/utils.py
```diff
@@ -1,3 +1,4 @@
+import torch
```
asogaard, line 1:
```suggestion
import torch
from torch import Tensor
```

src/graphnet/models/utils.py
```diff
@@ -15,3 +16,18 @@ def calculate_xyzt_homophily(x, edge_index, batch):
...
+
+def calculate_distance_matrix(xyz_coords):
+    '''
+    Input:
+        x.y,z co-ordinates of active DOMs
+
+    Calculates the matrix of distances squared from DOM i to DOM j in x,y,z co-ordinates
+
+    Returns:
+        matrix: tensor of size [N_doms,N_doms]
```
asogaard, line 29:
```suggestion
    Calculate the matrix of pairwise distances between pulses in (x,y,z)-coordinates.
    
    Args:
        xyz_coords: (x,y,z)-coordinates of pulses, of shape [nb_doms, 3].

    Returns:
        Matrix of pairwise distances, of shape [nb_doms, nb_doms]
```

asogaard: To conform with standard (Google-style) docstring structure, see https://www.python.org/dev/peps/pep-0257/.

src/graphnet/models/utils.py
asogaard: I think we should keep the name, but do what the name suggests: Return the matrix of pairwise, Euclidean distances. Currently, the matrix of squared distances is returned.

src/graphnet/models/utils.py
```diff
@@ -15,3 +16,18 @@ def calculate_xyzt_homophily(x, edge_index, batch):
...
+    '''
+    diff_sqrd = (xyz_coords.unsqueeze(dim=2) - xyz_coords.T.unsqueeze(dim=0))**2    
+    return torch.sum(diff_sqrd, dim=1)
```
asogaard, line 32:
```suggestion
    diff = (xyz_coords.unsqueeze(dim=2) - xyz_coords.T.unsqueeze(dim=0))
    return torch.sqrt(torch.sum(diff**2, dim=1))
```

src/graphnet/models/graph_builders.py
```diff
@@ -75,8 +78,47 @@ def __call__ (self, data: Data) -> Data:
...
+        # Constructs the adjacency matrix from the raw, DOM-level data and returns this matrix
+        if data.edge_index is not None:
+            print("WARNING: GraphBuilder received graph with pre-existing structure. ",
+                  "Will overwrite.")
+
+        xyz_coords = data.x[:,self._columns]
+        batch_mask = (data.batch.unsqueeze(dim=0) == data.batch.unsqueeze(dim=1))
+
+        distance_matrix = calculate_distance_matrix(xyz_coords)
+        affinity_matrix = torch.exp(-0.5*distance_matrix/self._sigma**2)
```
asogaard, line 109:
```suggestion
        affinity_matrix = torch.exp(-0.5 * distance_matrix**2 / self._sigma**2)
```

src/graphnet/models/graph_builders.py
```diff
@@ -75,8 +78,47 @@ def __call__ (self, data: Data) -> Data:
...
+
+        exp_row_sums = torch.exp(affinity_matrix).sum(axis=1)
```
asogaard, line 111:
```suggestion
        # Use softmax to normalise all adjacencies to one for each node
        exp_row_sums = torch.exp(affinity_matrix).sum(axis=1)
```

src/graphnet/models/graph_builders.py
```diff
@@ -75,8 +78,47 @@ def __call__ (self, data: Data) -> Data:
...
+        weighted_adj_matrix = torch.exp(affinity_matrix)/exp_row_sums.unsqueeze(dim=1)
+        
+        #only include edges with weights that exceed the chosen threshold (and are part of the same pulsemap)
+        sources,targets = torch.where((weighted_adj_matrix>self._threshold) & (batch_mask))
```
asogaard, line 115:
```suggestion
        # Only include edges with weights that exceed the chosen threshold (and are part of the same event)
        sources, targets = torch.where((weighted_adj_matrix > self._threshold) & (batch_mask))
```

src/graphnet/models/graph_builders.py
```diff
@@ -75,8 +78,47 @@ def __call__ (self, data: Data) -> Data:
...
+        
+
+    def __call__ (self, data: Data) -> Data:
+        # Constructs the adjacency matrix from the raw, DOM-level data and returns this matrix
+        if data.edge_index is not None:
+            print("WARNING: GraphBuilder received graph with pre-existing structure. ",
+                  "Will overwrite.")
+
+        xyz_coords = data.x[:,self._columns]
+        batch_mask = (data.batch.unsqueeze(dim=0) == data.batch.unsqueeze(dim=1))
```
asogaard, line 106:
```suggestion

        # Construct block-diagonal matrix indicating whether pulses belong to the same event in the batch
        batch_mask = (data.batch.unsqueeze(dim=0) == data.batch.unsqueeze(dim=1))
```

asogaard (changes requested):
Hi @BozianuLeon,

I have added a few mostly cosmetic suggestions, and one functional: the method `calculate_distance_matrix` currently returns the _squared_ distance matrix, rather than the distance matrix itself (as "promised" by the function name). I have proposed to keep the name and intention, and instead change what gets returned. Let me know what you think, and please make sure that you understand and agree with the suggestions that you might accept — otherwise just ask! :)

asogaard (approved):
Hi @BozianuLeon,
Thanks for the updates! 🥳   If you have tested the implementation and are happy with it, feel free to go ahead and merge!

PR #174, opened 2022-02-22
Fix for neutrinos/noise in database creation

BozianuLeon:
Important fix, database creation will throw an error for muons final position and stopped boolean if database has neutrinos/noise in. Requires the truth extractor to have padded values for final_pos and stopped.
Introduced by me in PR #148, now works with neutrino/noise.
@RasmusOrsoe could we make final_position_xyz for neutrinos the interaction vertex?


RasmusOrsoe:
Hey Leon - let's discuss this in person tomorrow.

Rasmus

asogaard:
Just to follow up on this: @RasmusOrsoe, it seems like you have opinions on this, so I'll let the two of you discuss. Let me know if you need anything from me. :)

BozianuLeon:
Have updated the stopped muon method, as well as changed the naming convention as per conversation with @RasmusOrsoe. I also think the stopped muon method belongs in the i3 extractor with other similar methods rather than utils which primarily deals with file handling. I hope that it is clear what we are naming what, if this needs to be said somewhere else. Also returning a dictionary makes the output far more understandable.

**Important** Going forward position_x, position_y, position_z for muons will be their final position i.e where they stop. This is the same name as the interaction vertex for neutrinos.

RasmusOrsoe (approved):

RasmusOrsoe (changes requested):
Hi - nice to see the naming change. Some of the checks are not successful though!! @BozianuLeon let me or @asogaard  know if you need help

RasmusOrsoe (approved):

RasmusOrsoe:
@BozianuLeon  approved - you may click merge!

PR #176, opened 2022-02-25
Add more variables from i3 mc weight dict

RasmusOrsoe:
Hi

This PR does two things:

1) Moves content in I3MCWeightDict from retro table to truth table
2) Adds three additional entries in I3MCWeightDict to the truth table - these quantities are required for PISA analysis

Tested on level7 events and works.

Rasmus

### DIFF
--- a/src/graphnet/data/i3extractor.py
+++ b/src/graphnet/data/i3extractor.py
@@ -387,10 +387,16 @@ class I3RetroExtractor(I3Extractor):
             if frame_contains_retro(frame):
                 output.update({
                     'osc_weight': frame["I3MCWeightDict"]["weight"],
+                    'OneWeight': frame["I3MCWeightDict"]["OneWeight"],
+                    'gen_ratio': frame["I3MCWeightDict"]["gen_ratio"],
+                    'osc_weight': frame["I3MCWeightDict"]["NEvents"],
                 })
             else:
                 output.update({
-                    'osc_weight': -1.,
+                    'osc_weight': -1,
+                    'OneWeight': -1,
+                    'gen_ratio': -1,
+                    'osc_weight': -1,
                 })
 
         return output
--- a/src/graphnet/data/i3extractor.py
+++ b/src/graphnet/data/i3extractor.py
@@ -309,23 +309,23 @@ class I3TruthExtractor(I3Extractor):
                 if abs(MCInIcePrimary.pdg_encoding) != 13: # if not muon
                     output.update({
                         'osc_weight'         : frame["I3MCWeightDict"]["weight"],
-                        'OneWeight'          : frame["I3MCWeightDict"]["OneWeight"],
+                        'one_weight'          : frame["I3MCWeightDict"]["OneWeight"],
                         'gen_ratio'          : frame["I3MCWeightDict"]["gen_ratio"],
-                        'NEvents'            : frame["I3MCWeightDict"]["NEvents"],
+                        'n_events'            : frame["I3MCWeightDict"]["NEvents"],
                     })
                 else:
                     output.update({
                         'osc_weight'        : frame["I3MCWeightDict"]["weight"],
-                        'OneWeight'         : padding_value,
+                        'one_weight'         : padding_value,
                         'gen_ratio'         : padding_value,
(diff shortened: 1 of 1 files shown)


src/graphnet/data/i3extractor.py
```diff
@@ -304,7 +304,29 @@ def __call__(self, frame, padding_value=-1) -> dict:
...
                     'stopped_muon': stopped,
-                })                
+                })  
+            if frame_contains_retro(frame):
+                if abs(MCInIcePrimary.pdg_encoding) != 13: # if not muon
+                    output.update({
+                        'osc_weight'         : frame["I3MCWeightDict"]["weight"],
+                        'OneWeight'          : frame["I3MCWeightDict"]["OneWeight"],
+                        'gen_ratio'          : frame["I3MCWeightDict"]["gen_ratio"],
+                        'NEvents'            : frame["I3MCWeightDict"]["NEvents"],
```
asogaard, line 314: @RasmusOrsoe, are these weight attributes specific to RetroReco, or OscNext?

src/graphnet/data/i3extractor.py
```diff
@@ -304,7 +304,29 @@ def __call__(self, frame, padding_value=-1) -> dict:
                     'final_position_y': final_position[1],
                     'final_position_z': final_position[2],
                     'stopped_muon': stopped,
-                })                
+                })  
+            if frame_contains_retro(frame):
+                if abs(MCInIcePrimary.pdg_encoding) != 13: # if not muon
+                    output.update({
+                        'osc_weight'         : frame["I3MCWeightDict"]["weight"],
+                        'OneWeight'          : frame["I3MCWeightDict"]["OneWeight"],
```
asogaard, line 312: I would suggest that we stick to snake-case naming convention for attributes (i.e. `OneWeight` -> `one_weight` and `NEvents` -> `n_events`) for consistency.

asogaard (changes requested):

src/graphnet/data/i3extractor.py
RasmusOrsoe: These quantities have nothing to do with Retro and are variables that comes from the simulation engine.

src/graphnet/data/i3extractor.py
RasmusOrsoe: Done @asogaard

src/graphnet/data/i3extractor.py
asogaard: In that case, is there a reason we enclose them in the `if frame_contains_retro(frame):` block? If these are from Genie (or whatever), should we check for this instead?

RasmusOrsoe: They are currently enclosed in this check because I know from experience `'osc_weight'` is not always available, but is always available when Retro is applied. I assume its the same for these keys.

I can remove the frame check but I would then instead have to wrap every single quantity into a

```
try:
    output['osc_weight'] = frame["I3MCWeightDict"]["weight"]
except:
    output['osc_weight']  = padding_value
```

asogaard: We could add a utility method like
```python
from typing import Any, Union

def try_get_key(frame: Union[I3Frame, dict], key: str, default_value: Any) -> Any:
    """Return `key` in `frame` if it exists; otherwise return `default_value."""
    try:
        return frame[key]
    except KeyError:
        return default_value
```
or, even better, if we know that `frame` and all of it's "child" objects are python-`dict`-like, we can just use the built-in method `.get(key, default_value)` (see https://docs.python.org/3/library/stdtypes.html#dict.get), i.e.
```python
output['osc_weight'] = try_get_key(frame["I3MCWeightDict"], "weight", padding_value)

# or, even better, if frame["I3MCWeightDict"] is dict-like
output['osc_weight'] = frame["I3MCWeightDict"].get("weight", padding_value)
```

asogaard:
Hi @RasmusOrsoe,
Is this PR still "live"? And if so, do you need anything from me to get it through? :)

RasmusOrsoe:
Hi! Still live and still needed. I have just not had the time to implement your suggestion

RasmusOrsoe:
closed because of #339

PR #177, opened 2022-03-01
CLI implementation

MortenHolmRep:
Added CLI capability to 'examples/train_model.py' using argparse.
The wandbLogger class call was moved into main to use its save location as part of the CLI implementation.


MortenHolmRep:
adding @asogaard @RasmusOrsoe fyi. For some reason I cannot add a reviewer in the github webpage interface.

MortenHolmRep:
Temporally closed, pending further implementation

Issue #173, opened 2022-02-22
Level 2 i3-files P Frame Problems


BozianuLeon:
The simulated data for level 2 (unlike higher levels) has several features that may/can cause the sqlitedataconverter to run into trouble. This issue is for documenting the issues, which should be fixed in an upcoming PR.

**MuonGun Simulation**
Contains 1 NullSplit P frame for every 1 InIceSplit P frame, we are only interested in InIceSplit frames. 
In correct P frames, the only RecoPulseSeriesMapMask is SplitInIcePulses, which is what we should target for all other level2 files (not SRTInIcePulses).

**Genie Simulation**
Also suffers from 50% NullSplit P frames.
The file names for these i3 files do not contain the i3 tag (e.g "2020 NuE_129999_000644_level2.zst") this means that the current i3 extractor will overlook such files. This is true for all flavours of neutrinos 12,14,16.

**Noise simulation**
At level2 there are a LOT of these files, though they contain very few InIceSplit P frames each. There are again NullSplit P Frames (many more, in fact) that will need to be ignored.
Noise i3 files also have SLOPSplit P frames used by others as a trigger searching for some specific non-relativistic monopoles, contain nothing of interest for our reconstruction purposes.

This is the natural extension of issue #131 and I am presently testing whether database builds work for all 3 kinds of simulation, as well as for those currently-working level3 and 7 files.

Issue #131, opened 2022-01-13
Create SQLite database from MC files


asogaard:
Depends on #129. This is only necessary if we will be training new GNN models on this Level 2 data.

BozianuLeon:
I am working on this issue, specifically covering the creation of level2 MC databases from relevant i3 files. Will happily assign myself and liaise with others.

asogaard:
Hi @BozianuLeon, I believe this has been done. Could you confirm, and if so, document the outcome and close the issue? Or is it waiting on #178?

Issue #129, opened 2022-01-13
Identify target MC files


asogaard:
We would need both muon and neutrino MC at Level 2, optionally also noise, to allow for a comparison with data. Identify and document these datasets.

Question:
* Should these MC datasets ideally also contain OscNext scores?

BozianuLeon:
Hey Andreas. As far as I know, level 2 i3 files do not contain OscNext scores (those calculated via RetroReco), as this analysis takes place further along the data pipeline.

asogaard:
That sounds right, @BozianuLeon!

asogaard:
Hi @BozianuLeon, I believe this has been done. Could you confirm, and if so, document the outcome and close the issue?

BozianuLeon:
It has, we have data at level2 in all simulation types and tested database creation as well as running inference using trained models.

Issue #128, opened 2022-01-13
Identify target data files


asogaard:
Tom Stuttard (NBI) has a 24hr burn sample available at Level 2 with OscNext scores for comparison. Identify, and document this sample

BozianuLeon:
Yes, these i3 files have been investigated and turned into a SQLite database. And the PR #178 means that these and other real data i3 files can be processed in the future.

PR #178, opened 2022-03-13
New i3 converter/sqlite_dataconverter method

BozianuLeon:
This PR addresses the following issues: #173,#131, and also allows #128 #129 to move forward. The good news is that this method works for level2 simulation, level2 real data, and reproduces level7 simulation databases. I have tested on all types of simulation at level2 and level7 and muongun at level3. But welcome any comments :)

The changes are as follows.
1. Tidy up i3extractor so that it can be run over real data. Ensuring that any is_mc check happens before any functions access the "TruthExtractor" which, clearly, is not available in real data. 
2. Ensure naming convention is consistent for muons and neutrinos (position_x,position_y,position_z).
3. Add _pulsemap attribute to DataConverter to identify target key in i3 P frames. e.g. SplitInIcePulses. This is achieved via a small function implemented by Rasmus.
4. Important check in _parallel_extraction to check that the P frame contains useful, non-empty information (i.e InIceSplits skipping NullSplits or SLOPSplits). This checks three things: if the pulsemap has a  non-empty dom_x column, if the truth table is non-empty and if the current key (e.g RetroReco) is non-empty. This way only those P frames that have actual information are put into the temporary databases.
5. Add '.gz' extension to the has_extension and find_i3_files utils functions in order to correctly identify the level 2 noise i3 files which end in 'i3.gz'. 

There are some questions I have for anyone reviewing, the first is about the event_counter which currently increases by 1 even for data frames that don't make it into the temporary database. This has no effect other than in the final database event numbers are not increasing in a regular way. If there are many NullSplit P frames then the event_nos could be quite far apart. If this is not desirable then the fix is fairly straightforward. 
The second is the potential for erroneous non-i3 file identification. GCD files also end with a '.gz' extension. This could only become a problem if the GCD file does not contain the strings 'gcd' or 'geo'. 

(There are also some completely non-essential print statements that can definitely be changed).


asogaard:
Hi @BozianuLeon,
Thanks for this nice PR and for the detailed description! I am happy to review right away. To you closing questions: 

- The output event numbers not incrementing by 1 all of the time is fine, I think, if a bit silly. If it is a straightforward fix, I think that would be nice to include; but if it is non-trivial, I don't think it should hold up the rest of the PR. 
- This issue about I3 file suffixes and patterns to match GCD files is a bit more worrying I think. If you have encountered GCD files that do not match the current patterns, I think we should update the latter to also catch these files. Feel free to give examples of such files here in the comments.

src/graphnet/data/i3extractor.py
```diff
@@ -296,6 +296,7 @@ def __call__(self, frame, padding_value=-1) -> dict:
                 'pid': MCInIcePrimary.pdg_encoding,
                 'interaction_type': interaction_type,
                 'elasticity': elasticity,
+                'dbang_decay_length': self.__extract_dbang_decay_length__(frame, padding_value),
```
asogaard, line 299: I think the change in this file makes a lot of sense. However, the naming of this method (which, I realise, was not implemented in this PR) is non-standard. Method names that start and end in double underscores are typically reserved for special, built-in methods (e.g., `__init__`, `__call__`, `__repr__`), and I would suggest we adhere to that. If we want to signify that this is meant to be a private/protected method, then it could be called `_extract_dbang_decay_length`, i.e., with a single leading underscore, which is what has been done elsewhere in the repo.

src/graphnet/data/sqlite_dataconverter.py
```diff
@@ -130,14 +132,12 @@ def _parallel_extraction(self, settings):
         dataframes_big = OrderedDict([
             (key, pd.DataFrame()) for key in self._table_names
         ])
-
```
asogaard, line 133: I would suggest to avoid pure whitespace/newline changes in cases when these are (seemingly?) not important to improve legibility, or similar. See also changes at (old) L64, L140, L164, L310.

src/graphnet/data/sqlite_dataconverter.py
```diff
@@ -334,3 +331,10 @@ def is_empty(features):
         return True
     else:
         return False
+
+
+def is_pulsemap_check(table_name):
```
asogaard, line 336:
```suggestion
def is_pulsemap_check(table_name: str) -> bool:
    """Check whether `table_name` corresponds to a pulsemap, and not a truth or RETRO table."""
```

src/graphnet/data/sqlite_dataconverter.py
```diff
@@ -334,3 +331,10 @@ def is_empty(features):
...
+    if 'retro' in table_name.lower() or 'truth' in table_name.lower():
+        return False
+    else: #could have to include the lower case word 'pulse'?
```
asogaard, line 339: This could be a good think to check for in a unit test.

src/graphnet/data/sqlite_dataconverter.py
```diff
@@ -62,6 +62,8 @@ def __init__(
              "I3TruthExtractor to allow for attaching unique indices.")
 
         self._table_names = [extractor.name for extractor in self._extractors]
+        self._pulsemap = [x for x in self._table_names if is_pulsemap_check(x)][0] #the first non-truth non-retro table name (pulsemap)
```
asogaard, line 65: 1. This may introduce a level of ambiguity in cases where there are multiple non-truth, non-retro pulsemaps, e.g., as may be the case in Upgrade data.
2. If we want to get a list of the (one or more) non-truth, non-retro pulsemap(s), I also think that it might be better to do `self._pulsemaps = [extractor.name for extractor in self._extractors if isinstance(extractor, I3FeatureExtractor)]` rather than doing string-based matching, as all feature extraction inherits from this class, whereas `I3RetroExtractor` and `I3TruthExtractor` don't.

src/graphnet/data/sqlite_dataconverter.py
```diff
@@ -151,17 +151,17 @@ def _parallel_extraction(self, settings):
                 # Concatenate data
                 for key, data in data_dict.items():
                     df = apply_event_no(data, event_no_list, event_count)
-                    if len(df):
+                    if bool( data_dict[self._pulsemap]['dom_x'] and data_dict[self._table_names[0]] and len(df) ): #only include if the dom_x is non-empty and the truth is non-empty
```
asogaard, line 154: 1. I think a somewhat detailed comment could be useful here, to explain the logic behind this statement.
2. This seems to assume that the truth extractor is at index 0, but that may not always be the case.
3. Also: Do we _want_ to require the existence of truth data for writing pulsemaps in data (where truth doesn't exist, unless I'm mistaken)?
4. I don't think the outer `bool(...)` statement is needed, as there are three implicit conversions to bools happening. If anything, I think it could be more useful to make the three boolean conversion explicit instead (e.g., `len(df) > 0`).
5. See also comment above: What happens in the case where we have multiple pulsemaps in the same event (e.g., Upgrade) where you may have hits in one but not the other, which, if you're unlucky, could be the first one.

asogaard (changes requested):
Hi @BozianuLeon,
I have left a few comments — let me know if there are any that we should discuss! :)

src/graphnet/data/sqlite_dataconverter.py
BozianuLeon: Agree almost totally with this comment. I would just say:
1.Definitely, I'll start writing such a description in a subsequent commit.
2.The position of the truth extractor is actually asserted in L60 but I am not opposed to this changing, if understandability improves.
3. There is a check in the I3TruthExtractor is_mc, that guards against trying to extract truth from real data and leaves a table in databases filled with padding values (but the all important event_no's). As things currently stand we don't allow databases to be made without a truth extractor, this could be changed but serious work would be required to ensure attached indices were unique (particularly if data is split across multiple databases)
4. Can, and will, change. I thought it was worthwhile making it explicit that this is a empty check. But I think `len(df) > 0` is even more clear.
5. A very good point, one that I didn't anticipate.  This method still supports multiple pulse maps being extracted but only if the first non-truth non-retro key is non-empty which I guess could be called the "master" pulsemap. If we take out the pulsemap check (i.e just check the truth and current key) then this leads to undesirable outcomes such as events with just truth and retro (for example) tables, which happens a lot in level 2 data since many P frames are "NullSplit" and contain truth, retro (etc.) and no pulse maps.

asogaard:
Hi @BozianuLeon,
The import error in the automatic build jobs have now been resolved. There is still an error, that should be resolved with one of my suggestions above. Let me know if you need any help getting this through. :)

asogaard:
Merging `icecube:main` into the feature branch should resolve the problem regarding the automatic build jobs.

BozianuLeon:
Have now merged, and more build jobs are now passing. Though apparently I am unable to import OrderedDict, not sure if this is something on my end. Let's talk tomorrow @asogaard :)

asogaard (approved):
Brilliant, @BozianuLeon! If you have check that this all works, I am happy to get this merged in. 🥳

PR #179, opened 2022-03-25
Oscillation Study (1/3): InSQLitePipeline Class

RasmusOrsoe:
This pull request adds InSQLitePipeline. 

InSqlitePipeline takes trained GNNs, runs inference on a specified database, and creates a sqlite database with truth, reconstructions from the GNN modules and, if available, retro reconstructions. This database is then the data foundations for analysis in SQLite.

This pipeline is the backbone of a future PR that allows for GNN predictions -> oscillation contours (Coming soon!)

edit: added an example script 

Rasmus


src/graphnet/data/pipeline.py
```diff
@@ -0,0 +1,173 @@
...
+        """Loads each module in _module_dict
+
+        Args:
+            module_dict (dict): dictionary of GNN ModuleList
+
+        Returns:
+            dict: module_dict with "loaded_module" field added.
+        """
+        for target in module_dict.keys():
+            model = torch.load(module_dict[target]['path'], map_location = 'cpu', pickle_module = dill)
```
BozianuLeon, line 82: Hey Rasmus, I was recently doing some model loading and pytorch often suggests loading from the state_dict as sometimes loading the full model can result in funky errors (see https://pytorch.org/tutorials/recipes/recipes/saving_and_loading_models_for_inference.html ) wondered if you had considered loading in from state_dict as we also save that afaik

asogaard:
Merging `icecube:main` into the feature branch should resolve the problem regarding the automatic build jobs.

RasmusOrsoe:
@asogaard  done

src/graphnet/data/pipeline.py
```diff
@@ -0,0 +1,173 @@
...
+            device = int(self._device[-1])
+        if os.path.isdir(outdir) == False:
+            dataframes = []
+            for target in self._module_dict.keys():
+                dataloader = self._make_dataloader(database, pulsemap)
+                trainer = Trainer(gpus = [device])
+                model = torch.load(self._module_dict[target]['path'], map_location = 'cpu', pickle_module = dill)
+                model._gnn.to(self._device)
+                model._device = self._device
+                model.to(self._device)
```
asogaard, line 61: If we're using a pytorch-lightning trainer to perform the inference, I don't think these lines are needed, as the trainer _should_ handle device settings.

src/graphnet/data/pipeline.py
```diff
@@ -0,0 +1,173 @@
...
+                model.inference()
+                results = get_predictions(trainer,model, dataloader, self._module_dict[target]['output_column_names'], ['event_no'])
+                dataframes.append(results.sort_values('event_no').reset_index(drop = True))
+            df = self._combine_outputs(dataframes)
+            self._make_pipeline_database(outdir,df, database)
+        else:
+            print(outdir)
+            print('WARNING - Pipeline named %s already exists! \n Please rename pipeline!'%self._pipeline_name)
+        return
+    def _load_modules(self, module_dict):
```
asogaard, line 72: This method seems to not be used?

src/graphnet/data/pipeline.py
```diff
@@ -0,0 +1,173 @@
...
+            database (path): path to database with pulsemap and truth
+            pulsemap (str): name of pulsemaps
+        """
+        outdir = self._get_outdir(database)
+        if isinstance(self._device, str): # Because pytorch lightning insists on breaking pytorch cuda device naming scheme
+            device = int(self._device[-1])
+        if os.path.isdir(outdir) == False:
+            dataframes = []
+            for target in self._module_dict.keys():
+                dataloader = self._make_dataloader(database, pulsemap)
```
asogaard, line 56: Currently, the dataloader is being constructed once for every model? Is this necessary?

src/graphnet/data/pipeline.py
```diff
@@ -0,0 +1,173 @@
...
+            model._gnn.to(self._device)
+            model._device = self._device
+            model.to(self._device)
+            model.eval()
+            module_dict[target]['loaded_module'] = model
+        return module_dict
+
+    def _make_dataloader(self, database, pulsemap):
+        dataloader = make_dataloader(db = database, pulsemaps = pulsemap, features = self._features, truth = self._truth, batch_size = self._batch_size, shuffle = False, selection = None, num_workers = self.n_workers, persistent_workers= False)
+        return dataloader
```
asogaard, line 92: A lot of the arguments to the `InSQLPipeline` class are only used as arguments to `make_dataloader` here. So I'm wondering if it would be simpler to drop this method and then, in the class constructor, just create an `self._dataloader = make_dataloader(...)` member variable rather than first setting `self._features`, `self._truth`, etc. as members and then only referencing them here?

src/graphnet/data/pipeline.py
asogaard: We could set `persistent_workers=True` if that's the issue?

src/graphnet/data/pipeline.py
```diff
@@ -0,0 +1,173 @@
...
+            query_columns.append(f"{column} {type_}")
+        query_columns = ', '.join(query_columns)
+
+        code = (
+            "PRAGMA foreign_keys=off;\n"
+            f"CREATE TABLE {table_name} ({query_columns});\n"
+            "PRAGMA foreign_keys=on;"
+        )
+        self._run_sql_code(pipeline_database, code)
+        return
```
asogaard, line 173: This part of the code has a lot in common with similarly named methods in `SQLiteDataConverter` that I'm thinking we should try to resolve. (One of the most important software engineering maxims is "DRY — Don't repeat yourself"). I'm noticing that none of these methods really make use of the calling instance (`self`) e.g. through member variables. This means that they could be converted to common utility methods, e.g. in an `src/graphnet/data/sql.py` module, that could contain all general functions that are relevant to SQL(ite)?

src/graphnet/data/pipeline.py
```diff
@@ -0,0 +1,173 @@
...
+        
+
+    def _save_to_sql(self, df, table_name, pipeline_database):
+        print('SUBMITTING %s'%table_name)
+        self._create_table(pipeline_database,  table_name, df)
+        engine = sqlalchemy.create_engine('sqlite:///' + pipeline_database)
+        df.to_sql(table_name, con=engine, index=False, if_exists='append')
+        engine.dispose()
+        print('%s SUBMITTED'%table_name)
+        return
```
asogaard, line 142: This could probably also be generalised into a common utility method (see comment below).

asogaard (changes requested):
Hi @RasmusOrsoe,
Thanks for the first instalment in the PISA trilogy! 😀   I have provided some comments to the PR to reduce code duplication and hopefully add a bit of structure. But I think this is a really nice contribution, and I applaud your object oriented programming! 👏

kaareendrup:
Hi @RasmusOrsoe it seems you have plenty of comments to look at already, but this is just a reminder of what we talked about yesterday about the predictions not being saved continuously (batchwise) :)

examples/make_pipeline_database.py
```diff
@@ -0,0 +1,50 @@
+from graphnet.data.pipeline import InSQLitePipeline
+from graphnet.data.constants import TRUTH, FEATURES
+import torch
+torch.multiprocessing.set_sharing_strategy('file_system')
+
+def get_output_column_names(target):
```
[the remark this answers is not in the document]
RasmusOrsoe: I agree and I started an issue on this a few weeks back.

examples/make_pipeline_database.py
```diff
@@ -0,0 +1,50 @@
...
+        output_column_names=['position_x_pred','position_y_pred','position_z_pred']
+    return output_column_names
+
+def build_module_dictionary(targets):
+    module_dict = {}
+    for target in targets:
+        module_dict[target] = {}
+        module_dict[target]['path'] = '/home/iwsatlas1/oersoe/phd/oscillations/models/final/dynedge_oscillation_final_%s.pth'%(target) #dev_lvl7_robustness_muon_neutrino_0000/dynedge_oscillation_%s/dynedge_oscillation_%s.pth'%(target,target)
+        module_dict[target]['output_column_names'] = get_output_column_names(target)
+    return module_dict
```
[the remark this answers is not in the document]
RasmusOrsoe: I personally do not feel that the code becomes more readable by wrapping this in yet another class. I think the idea with CombinedModel is really good - perhaps it could be a special case of a more general 'EnsembleModel' class? It would be quite handy!

src/graphnet/data/pipeline.py
RasmusOrsoe: removed.

src/graphnet/data/pipeline.py
RasmusOrsoe: removed.

src/graphnet/data/pipeline.py
RasmusOrsoe: its an iterable so by creating an identical one for every model makes me sleep well at night.

src/graphnet/data/pipeline.py
RasmusOrsoe: Not entirely sure what you mean. I did something - perhaps this is what you meant?

src/graphnet/data/pipeline.py
RasmusOrsoe: I have moved most of the sqlite stuff to `data/utils.py`. I have kept` _create_table` because it creates the table scheme and I feel having this here makes sense. Making a general purpose function for this is outside of the scope of what I'd like to spend my time on right now. So if we could perhaps let it be, I'd appreciate it.

src/graphnet/data/pipeline.py
RasmusOrsoe: moved

src/gnn_reco/data/pipeline.py
```diff
@@ -0,0 +1,122 @@
+from abc import ABC, abstractmethod
```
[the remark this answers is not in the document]
RasmusOrsoe: removed

RasmusOrsoe:
@kaareendrup I've introduced a chunk_size argument that lets you control the memory. A database is sliced into np.ceil(n_events/chunk_size) dataloaders. Currently chunk_size defaults to 1 million events - but you can put it to whatever you want. Memory scales linearly with the number of GNN modules - so 5 GNN modules means approx. chunk_size*5 predictions in memory. Hope this helps.

@asogaard this latest commit changed things a bit fyi.

Rasmus

asogaard (approved):
Brilliant! 👍

PR #181, opened 2022-03-29
Fixing build issue

asogaard:
For the past week or two, there has been an issue of automatic build jobs failing due to an import error. This has been traced back to the change in version from `torch-cluster==1.5.9` to `1.6.0`. As a result, I have explicitly specified all package versions in the setup.py file and in the GitHub Actions scripts that run the automatic jobs.

FYI: @BozianuLeon @MortenHolmRep @kaareendrup


RasmusOrsoe (approved):
Yay!

PR #182, opened 2022-04-06
Functionality for pulse level noise cleaning

kaareendrup:
This PR concerns the pulse level noise classification used by me and @MortenHolmRep and is meant as an illustration/discussion point on how to best implement this functionality as well making it avaiable to the rest of you.

Pulse level classification is enabled by creating a new dataset class that queries the database for truth flags, and providing a version of the DynEdge architecture where the aggregation across nodes has been removed. An alternative `get_predictions` function is also implemented for node level predictions.

An example script is provided to showcase the functionality.

NOTE: This example script currently only works when the dataset class used in ['graphnet.models.training.utils.make_dataloader'](https://github.com/icecube/graphnet/blob/21f4e36ef17c765a04cde0b2e34d5f802a988055/src/graphnet/models/training/utils.py#L35) is changed from `SQLiteDataset` to `SQLiteDatasetCleaning`.

I would like to hear you opinions on how to implement this in the most meaningful way.

Best
Kaare :)

@RasmusOrsoe @asogaard


RasmusOrsoe:
@kaareendrup thank you very much!

One of the major things to discuss here I think is that:
**_Certain new applications of our techniques seem to be most naturally implemented by customizing or editing the current dataset class_**

One such example is this PR, but also a switch from a pulse-wise graph representation to a DOM-wise graph representation is (in my opinion) achieved most naturally this way too. 

From the looks of it, we could modify the current dataset class to accept an argument `include_pulse_truth_flags` that would default to `False`. If this is set to `True`, a SQL-check could be run to check if a truth flag table exists in the database, and if so, your `self._add_truth_flag(i, graph)` could be run. 

What  do you think @asogaard ?

kaareendrup:
Thanks for looking at it so quickly guys, and for your clever thoughts! 

Regarding 1) I agree that the node level truth noise flags are very similar to the other variables and very few changes are required to make them function. And I can see the benefits of both of both of your proposals @asogaard but I'm not sure I'm guite able to grasp the consequences of choosing either solution. And I of couse apologize for the nastiness of hard-coding the truth_flag column name, I would of course never do that in a PR I expected to pass without major changes :)

Concerning 2) I totally agree that this would probably be easy, not very dangerous, and generally a good functionality to have, to just let `get_predicitons` infer that it's note level by comparing the sizes of the arrays

MortenHolmRep:
I have just pushed the latest update on my own git with recent changes, herein we can filter feature types pre graph construction, with slight modification this can be done for all features. Currently, it is only done for dom_types and string as an example. However, @RasmusOrsoe was sick of me being sick, so he made a more efficient string selector than the one I've implemented that filters the data out already at the query level. Which is a lot better IMHO.

Please correct me if I am wrong on this, but for usability, we are interested in changing as few parameters as possible. when we use `Features.DEEPCORE` we signal that we want to use DEEPCORE exclusively. With Rasmus' implementation, it automatically filters out strings even if the dataset includes more strings in the dataset. Therefore defining the method={DEEPCORE, UPGRADE, ..} early on to only include relevant parts of the IceCube data would be more consistent.
With the addition of CLI we could end up with flags such as `--DEEPCORE --node` or `--DEEPCORE --event` or `--truth_flags True` per Rasmus' suggestion.

kaareendrup:
Closed, following #190

PR #184, opened 2022-04-07
Filters In truth table, osc_weight in retro table

BozianuLeon:
This is a small pull request that simply includes 5 of the collaborations filters that are useful for LE, oscillation analyses and neutrino sources . We extract these at the same time as the I3EventHeader etc. (they are calculated only for InIceSplit P frames) and will be located in the truth table of future SQLite dbs. Having these filters is very important for the one day in data project.


BozianuLeon:
Also correcting an earlier PR error, a variable name missing an 's' on the end.

RasmusOrsoe:
@BozianuLeon if you could find the time to also include the "passed_lvlx" booleans, (x being 2, 3, 4, 5 etc) it would be quite nice.

BozianuLeon:
This PR was actually specifically targeting real data i3-files, which all contain FilterMask, I've also checked the level 2 simulation and this all contains the same filters (or at least an older subset of them). This shouldn't require a try_get method (which I'd be happy to implement) unless one of the IceCube WG changes the filters the require to be run on all lvl2 data.

I'd be happy to make a passed_levelx boolean, but this does not currently exist (to my knowledge) in low-level i3 files, when I have a little bit more time. This was really just to get the DC filter on raw level 2 data.

BozianuLeon:
Add in the oscillation_weight for noise Monte Carlo, for later calculation of event rates.

src/graphnet/data/i3extractor.py
```diff
@@ -380,27 +392,26 @@ def __call__(self, frame) -> dict:
...
-            #    'L7_PIDClassifier_FullSky_ProbTrack': frame["L7_PIDClassifier_FullSky_ProbTrack"].value,
-            #})
 
         if frame_is_montecarlo(frame):
-            if frame_contains_retro(frame):
+            if frame_is_noise(frame):
                 output.update({
-                    'osc_weight': frame["I3MCWeightDict"]["weight"],
+                    'osc_weight': frame["noise_weight"]["weight"],
                 })
```
asogaard, line 400: @RasmusOrsoe, does this change affect you? Perhaps it would be better to introduce a new output variable called `noise_weight` or similar?

asogaard (approved):
Hi @BozianuLeon,

Sorry for taking so long to get around to this. I think this PR looks perfectly fine, provided @RasmusOrsoe is happy about the change in the weight variable. :)

src/graphnet/data/i3extractor.py
RasmusOrsoe: Hey @asogaard. This looks suspicious, so thanks for pointing it out! 

**However**, I think this is a change following an offline conversation I had with Leon. The `I3MCWeightDict` weight weighs events to physical rates. If I remember correctly, this weight is not stored in `["I3MCWeightDict"]` for pure noise events, and I think this is why Leon is adding it from a different field for noise events. @BozianuLeon can you confirm?

BozianuLeon: I can confirm what you are saying @RasmusOrsoe, this is oscNext's way of handling weights of noise events, just storing it somewhere differently. I do not think we lose any information by overwriting osc_weight here

PR #189, opened 2022-04-19
Dom wise graph representation

RasmusOrsoe:
Adds additional functionality to the sqlite dataset class that lets the user specify the "node representation" in graphs. The node representation defaults to "pulse" - our current choice.

When node_representation = "dom", the SQLite dataset class groups same-pmt pulses to the same node, and aggregate the pulse information by taking min, max, mean and std of the pulse time and charge.

Rasmus

### DIFF
--- a/src/graphnet/data/sqlite_dataset.py
+++ b/src/graphnet/data/sqlite_dataset.py
@@ -135,15 +135,12 @@ class SQLiteDataset(torch.utils.data.Dataset):
             pulse_statistics[count,0] = torch.min(time)
             pulse_statistics[count,1] = torch.mean(time)
             pulse_statistics[count,2] = torch.max(time)
-            pulse_statistics[count,3] = torch.std(time)
+            pulse_statistics[count,3] = torch.std(time, unbiased=False)
             pulse_statistics[count,4] = torch.min(charge)
             pulse_statistics[count,5] = torch.mean(charge)
             pulse_statistics[count,6] = torch.max(charge)
-            pulse_statistics[count,7] = torch.std(charge)
-            count +=1
-        #print(unique_doms.shape)
-        #print(n_pulses_pr_dom.shape)
-        #print(pulse_statistics.shape)      
+            pulse_statistics[count,7] = torch.std(charge, unbiased=False)
+            count +=1   
         data.x = torch.cat((unique_doms, n_pulses_pr_dom.unsqueeze(1), pulse_statistics), dim = 1)
         return data
 
--- a/src/graphnet/data/sqlite_dataset.py
+++ b/src/graphnet/data/sqlite_dataset.py
@@ -38,6 +38,8 @@ class SQLiteDataset(torch.utils.data.Dataset):
         assert isinstance(features, (list, tuple))
         assert isinstance(truth, (list, tuple))
 
+        assert node_representation.lower() in ['pulse', 'dom']
+
         self._database = database
         self._pulsemaps = pulsemaps
         self._features = [index_column] + features
@@ -124,14 +126,20 @@ class SQLiteDataset(torch.utils.data.Dataset):
         return torch.unique(tensor, return_counts = True, return_inverse=True, dim=0)
 
     def _make_dom_wise_representation(self,data):
-        unique_doms, inverse_idx, n_pulses_pr_dom = self._get_unique_positions(data.x[:,[0,1,2,5,6]])
+        unique_doms, inverse_idx, n_pulses_pr_dom = self._get_unique_positions(data.x[:,[self._features.index('dom_x')   -1,
+                                                                                        self._features.index('dom_y')    -1,
+                                                                                        self._features.index('dom_z')    -1,
+                                                                                        self._features.index('rde')      -1,
+                                                                                        self._features.index('pmt_area') -1]])
         unique_inverse_indices = torch.unique(inverse_idx)
         count = 0
         pulse_statistics = torch.zeros(size = (len(unique_doms), 8))
         #'dom_x','dom_y','dom_z','dom_time','charge','rde','pmt_area'
+        time_idx = self._features.index('dom_time')-1
+        charge_idx = self._features.index('charge')-1
         for unique_inverse_idx in unique_inverse_indices:
-            time   = data.x[inverse_idx == unique_inverse_idx,3]
-            charge = data.x[inverse_idx == unique_inverse_idx,4]
+            time   = data.x[inverse_idx == unique_inverse_idx,time_idx]
+            charge = data.x[inverse_idx == unique_inverse_idx,charge_idx]
             pulse_statistics[count,0] = torch.min(time)
             pulse_statistics[count,1] = torch.mean(time)
             pulse_statistics[count,2] = torch.max(time)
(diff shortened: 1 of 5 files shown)


src/graphnet/data/sqlite_dataset.py
```diff
@@ -19,6 +19,7 @@ def __init__(
         truth_table: str = 'truth',
         selection: Optional[List[int]] = None,
         dtype: torch.dtype = torch.float32,
+        node_representation = 'pulse',
```
asogaard, line 22: This would probably be best suited for an "enum" or similar type of structure (i.e., to have a fixed set of options), but this is perfectly fine for now.

src/graphnet/data/sqlite_dataset.py
```diff
@@ -118,6 +120,30 @@ def _get_dbang_label(self, truth_dict):
         except:
             return -1
 
+    def _get_unique_positions(self, tensor):
+        return torch.unique(tensor, return_counts = True, return_inverse=True, dim=0)
+
+    def _make_dom_wise_representation(self,data):
+        unique_doms, inverse_idx, n_pulses_pr_dom = self._get_unique_positions(data.x[:,[0,1,2,5,6]])
```
asogaard, line 127: I would be hesitant to hardcode position, dom number, and string number variables as integer indices. It would probably be more clear to do something like https://github.com/icecube/graphnet/blob/main/src/graphnet/models/detector/icecube.py#L116  (although I admit that hardcoding these as strings is only slightly better than doing so as integers). Also, I don't think that indices `5` and `6` refer to the dom and string number, if that was the intention?
```suggestion
        unique_doms, inverse_idx, n_pulses_pr_dom = self._get_unique_positions(torch.stack((
            data['dom_x'],
            data['dom_y'],
            data['dom_z'],
            data['dom_number'],
            data['string'],
        ), dim=1))
```

src/graphnet/data/sqlite_dataset.py
```diff
@@ -161,12 +187,25 @@ def _create_graph(self, features, truth):
...
+            n_pulses = torch.tensor(len(x), dtype=torch.int32)
+            graph = Data(
+                x=x,
+                edge_index= None
+            )
+            graph = self._make_dom_wise_representation(graph)
+        else:
+            print('WARNING: node representation %s not recognized!'%self._node_representation)
+            
+
```
asogaard, line 208: DRY. Also, the check for `node_representation in ['pulse', 'node']` should probably be done in the constructor, such that we can assume that it has a standard value in all other methods. I'll assume that here to simplify.
```suggestion
        x = torch.tensor(data, dtype=self._dtype)
        n_pulses = torch.tensor(len(x), dtype=torch.int32)
        graph = Data(
            x=x,
            edge_index= None
        )
            
        if self._node_representation == 'dom':
            graph = self._make_dom_wise_representation(graph)
```

src/graphnet/data/sqlite_dataset.py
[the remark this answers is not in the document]
asogaard: If you stick to strings, though, I would add a `assert node_representation in ['pulse', 'node']` check in the constructor (see also comment below).

src/graphnet/models/detector/icecube.py
```diff
@@ -40,6 +40,75 @@ def _forward(self, data: Data) -> Data:
...
+        #pulse_statistics[count,0] = torch.min(time)
+        #pulse_statistics[count,1] = torch.mean(time)
+        #pulse_statistics[count,2] = torch.max(time)
+        #pulse_statistics[count,3] = torch.std(time)
+        #pulse_statistics[count,4] = torch.min(charge)
+        #pulse_statistics[count,5] = torch.mean(charge)
+        #pulse_statistics[count,6] = torch.max(charge)
+        #pulse_statistics[count,7] = torch.std(charge)
+        #unique_doms, n_pulses_pr_dom.unsqueeze(1), pulse_statistics
+
```
asogaard, line 78:
```suggestion
```

src/graphnet/models/detector/icecube.py
```diff
@@ -40,6 +40,75 @@ def _forward(self, data: Data) -> Data:
 
         return data
 
+class IceCube86_v2(Detector):
```
asogaard, line 43: We probably ought to make it clear that this takes in different data than `IceCube86` and therefore is not just a version/alternative of this.
```suggestion
class IceCube86_DOM(Detector):
```

src/graphnet/models/detector/icecube.py
```diff
@@ -40,6 +40,75 @@ def _forward(self, data: Data) -> Data:
...
+    """`Detector` class for IceCube-86 with nodes as doms."""
+
+    # Implementing abstract class attribute
+    features = FEATURES.ICECUBE86
+
+    @property
+    def nb_inputs(self) -> int:
+        return len(self.features) + 7
```
asogaard, line 51: Is this true? Does it add an additional 7 features compared to what's in `FEATURES.ICECUBE86`?

src/graphnet/models/detector/icecube.py
```diff
@@ -40,6 +40,75 @@ def _forward(self, data: Data) -> Data:
...
+
+    @property
+    def nb_outputs(self):
+        return self.nb_inputs
```
asogaard, line 55:
```suggestion
        return self.nb_inputs
        
```

src/graphnet/models/detector/icecube.py
```diff
@@ -40,6 +40,75 @@ def _forward(self, data: Data) -> Data:
...
+        data.x[:,8] *= 20
+
+        data.x[:,9] /= 1.05e+04  # dom_time
+        data.x[:,9] -= 1.
+        data.x[:,9] *= 20
+
+        data.x[:,10] /= 1.  # charge
+        data.x[:,11] /= 1.  # charge
+        data.x[:,12] /= 1.  # charge
+        data.x[:,13] /= 1.  # charge
```
asogaard, line 110: These do not seems to align with the 8 features being output from `_make_dom_wise_representation` above?

src/graphnet/models/graph_builders.py
```diff
@@ -3,8 +3,8 @@
```
asogaard, line 3: The changes to this file seem to be accidental? Perhaps they should be reverted?

asogaard (changes requested):
Hi @RasmusOrsoe,

I have gone through the PR now and added a few comment! :)

src/graphnet/data/sqlite_dataset.py
```diff
@@ -118,6 +120,30 @@ def _get_dbang_label(self, truth_dict):
         except:
             return -1
 
+    def _get_unique_positions(self, tensor):
+        return torch.unique(tensor, return_counts = True, return_inverse=True, dim=0)
+
+    def _make_dom_wise_representation(self,data):
```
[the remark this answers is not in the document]
RasmusOrsoe: It's fairly performant, but it is slower than the pulse representation. I agree with your suggestion and would welcome input on how to vectorize this. I thought about it a bit and sought inspiration from your upgrade code, but I could not see a direct application for this. So I'd be quite interested to see the solution.

src/graphnet/data/sqlite_dataset.py
RasmusOrsoe: [0,1,2,5,6] corresponds to ['dom_x', 'dom_y', 'dom_z', 'rde', 'pmt_area' ]. I include 5,6 because they are then included in the `unique_doms` variable, thus limiting the subsequent for loop to only match pulse-related info with the DOM.  The input features are not saved as fields in the graph, so your suggestion will not work. But I will find a solution that removes the hard-coded indicies.

src/graphnet/models/detector/icecube.py
RasmusOrsoe: Yes. These are the aggregated pulse information.  See https://github.com/icecube/graphnet/blob/b7420b8163c8eb4f30a4dd0b3784582fe4dcb974/src/graphnet/data/sqlite_dataset.py#L132

src/graphnet/models/detector/icecube.py
RasmusOrsoe: The graph has 14 entries, not 8. Not sure what you mean?

src/graphnet/models/graph_builders.py
RasmusOrsoe: I do not understand what this comment

RasmusOrsoe: I do not understand what this comment refers to @asogaard - Could you help me out?

src/graphnet/models/detector/icecube.py
RasmusOrsoe: Yeah, let's do that. But at the same time, I think it's a little unintuitive to give this a different "detector" class. Perhaps we should rethink this structure a little in the future.

src/graphnet/models/graph_builders.py
asogaard: It's just that the only changes to this file seem to be whitespace changes and adding a few imports that are not used? (E.g., `from graphnet.components.pool import group_identical`) If these are indeed unintentional, I was just asking if it would be better to just not modify this file at all in this PR.

asogaard:
NB: Have opened #196 as an alternative implementation.

asogaard:
@RasmusOrsoe I figure this can be closed following #196?

PR #190, opened 2022-04-19
Pulse noise training

RasmusOrsoe:
This is my take on a 'beautified' version of #182 according to the comment thread.

- Adds functionality to include node-level truth (generalizes to any number of truths from any node truth table)
- Lets the user specify a range of strings (so-called 'string_selection'). If a string selection is made, only pulses seen by DOMs on these strings are returned by the SQLite query.

- `get_predictions_pulse_level` is still ugly. Would like input on how to do this generally @asogaard 

@kaareendrup @MortenHolmRep  fyi

### DIFF
--- a/src/graphnet/data/sqlite_dataset.py
+++ b/src/graphnet/data/sqlite_dataset.py
@@ -85,36 +85,6 @@ class SQLiteDataset(torch.utils.data.Dataset):
             graph[node_truth_column] = torch.tensor(self._get_node_truth(i, node_truth_column)).reshape(-1)
         return graph
 
-    def _get_node_truth(self, i, node_truth_column):
-        """Query SQLite database for node truth information.
-        """
-        if self._database_list == None:
-            index = self._indices[i]
-        else:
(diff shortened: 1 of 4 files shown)


RasmusOrsoe:
@asogaard Thanks for all the comments, and its 100% easier for the workflow when I can simply commit your changes directly! Thank you!

I've committed all comittable changes and I'll try to get the last few things dinged out tomorrow.

RasmusOrsoe:
@asogaard should have addressed everything now. Checked that the code runs.

asogaard (approved):
Brilliant, thanks so much for the excellent work, @RasmusOrsoe! 🥳    When this is merged, I figure we can go ahead and close #182?

kaareendrup:
This looks very good guys! Great work @RasmusOrsoe . With the thesis deadline approaching rapidly, I was hoping you would jump on this and allow me to focus on other things, and you certainly came through! I will happily close #182 and speak to @MortenHolmRep about implementing the DOM selections

PR #196, opened 2022-04-28
DOM-wise coarsening

asogaard:
Have:
* Tried to rework #189 to take the coarsening functionality out of the `SQLiteDataset` class and into a set of small, dedicated classes, inheriting from `Coarsening`. 
* Implemented vectorised local pooling in addition to loop-based. Have validated that the two yield the same results (up to the order of the output columns) and that the vectorised version is approx. x3 faster.

The loop-based pooling should be taken out before merging. The unit tests are also incomplete currently.

Tagging @RasmusOrsoe

### DIFF
--- a/.github/actions/install-torch-cpu/action.yml
+++ b/.github/actions/install-torch-cpu/action.yml
@@ -14,8 +14,8 @@ runs:
   steps:
     - name: "Install dependencies"
       run: |
-        pip install --upgrade pip 
-        pip install wheel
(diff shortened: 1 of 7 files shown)


asogaard:
Have updated the PR to ensure that `setuptools==59.5.0` is installed, both in the default conda environment and on the GitHub Actions runners. This should close #197.

asogaard:
Ping @RasmusOrsoe

PR #198, opened 2022-04-28
Fixed call of get_predictions(...)

timephy:
This fix is needed because the function signature of `get_predictions(...)` was changed recently.

The 'old' signature was:
```python
def get_predictions(trainer, model, dataloader, prediction_columns, additional_attributes=None):
```
But the new one is:
```python
def get_predictions(trainer, model, dataloader, prediction_columns, node_level = False, additional_attributes=None):
```

This problem might likely occur in many places, since `get_predictions` is probably used in many places.
It might make sense to make `additional_attributes` as well as `node_level` kw-only-args in order to produce helpful error messages.

Suggestion:
```python
def get_predictions(trainer, model, dataloader, prediction_columns, *, node_level = False, additional_attributes=None):
```

### DIFF
--- a/examples/train_model.py
+++ b/examples/train_model.py
@@ -127,7 +127,7 @@ def main():
         model,
         validation_dataloader,
         [config["target"] + '_pred'],
-        [config["target"], 'event_no'],
+        additional_attributes=[config["target"], 'event_no'],
     )
 
     save_results(config["db"], run_name, results, archive, model)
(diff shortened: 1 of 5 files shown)


asogaard:
Hi @timephy,
I think that is an excellent observation and suggestion! Would you mind including the proposed change in this PR, then we can fix this issue straight away. :)

timephy:
No problem, here you go! :)

timephy:
Also just fixed all calls of `get_prediction()` which I found via `Control+F`.

asogaard (approved):
🙌

PR #204, opened 2022-05-06
PR review of Cluster submission implementation

MortenHolmRep:
Initial integration of tom stuttards cluster submission tool.
an example script called train_model_cluster.py has been included in the `graphnet/examples` folder.

running `train_model_cluster.py` executes the `train_model_cluster.sh` script, ensuring that the environment we defined is pulled along into the cluster, which in turn calls the ordinary `train_model.py` script.
We, therefore, do not need to alter the current workflow to submit our scripts to the cluster and only have to deal with the submission parameters.

The current iteration should be able to run on NBI, NPX, DESY and GRID. Although Tom has warned about the extra hurdles of setting it up on GRID. It is however something we may need later down the road.
The submission script automatically detects which cluster you have access to and submits the job to the given cluster. Meaning, that if we have cloned the git repo to hep, it will use the hep cluster for the job.

Any ideas for changes, reductions, etc. are welcome

PR #206, opened 2022-05-10
CI/CD fixes

asogaard:
As mentioned in #204 this PR:
* Removes `lint` job in GitHub Actions
* Drops all manually generated badges in favour of ones from [CodeClimate](https://codeclimate.com/github/asogaard/graphnet/badges)
* Remove pytorch and PyG from conda env file to avoid problems switching between versions

Additionally:
* Adds `black` job to GitHub Actions to automatically check compliance
* Adds custom `black` configuration file (`black.toml`) to set max line length to 79 characters and to exclude files from versioneer, and formats the existing python files in the repo accordingly.


MortenHolmRep (approved):
Black performs a few weird formatting options but overall it looks fine.

PR #212, opened 2022-06-02
PR new setup

MortenHolmRep:
An attempt at a more streamlined setup. 
The new setup should work by running `pip install -r requirements.txt -e .[develop]`, where all needed packages should be installed. 

The setup has only been tested locally as hep is very slow today (atleast for me).
Currently, the new setup needs to be tested on hep:
- [x] installation from scratch, new python (not sure of version, but higher than 3.8), no packages
- note to above: seems to work with 3.8 now
- [x] run examples/train_model.py


.gitignore
```diff
@@ -138,4 +138,7 @@ lightning_logs/
 wandb/
 **.png
 **.pth
-**.ipynb
\ No newline at end of file
+**.ipynb
+datasets/dev_lvl7_robustness_muon_neutrino_0000.db
+examples/train_model.py
+examples/train_model.py
```
asogaard, line 144: I would not add `examples/train_model.py` to .gitignore, as we might reasonably want to update it. I think it would be better to just not add local changes to this file to unrelated commits. Wrt. the other file, I think it it too specific to add to the .gitignore. I would suggest `**.db` or similar, if we don't ever want to add SQLite databases to the repo (which is probably pretty reasonable).
```suggestion
**.ipynb
**.db

```

examples/train_model.py
```diff
@@ -33,7 +33,7 @@
     project="example-script",
```
asogaard, line 33: I would revert changes to this file, so as to not change it as part of this PR unless there are functional changes that you think should be included.

setup.py
```diff
@@ -1,36 +1,35 @@
+try:
+    from setuptools import setup, find_packages
+except ImportError:
+    from distutils.core import setup
+
```
asogaard, line 5: Why is this needed? Or: why has this not been an issue before? Is it due to the updated `setuptools` version?

setup.py
```diff
@@ -1,36 +1,35 @@
...
+    "wandb>=0.12",
+    "matplotlib>=3.5",
+    "scikit_learn~=1.1.1",
+    "scipy>=1.8",
+    "torch~=1.11",
+    "torch-cluster==1.6.0",
     "torch-scatter==2.0.9",
-    "torch-sparse==0.6.12",
+    "torch-sparse==0.6.13",
     "torch-spline-conv==1.2.1",
```
asogaard, line 30: I figure we should aim to specify version _ranges_ (e.g., `torch_cluster>=1.6.0` or similar) rather than specific versions, or — ideally — not specify versions here at all, unless it is strictly necessary for compatibility.

setup.py
```diff
@@ -1,36 +1,35 @@
...
-    "torch-geometric==2.0.1",
-    "pytorch-lightning==1.5.6",
-    "dill",
-    "wandb",
-    "matplotlib",
+    "torch-geometric==2.0.4",
```
asogaard, line 31: See comment above about version range.

MortenHolmRep:
> Thanks for the PR, @MortenHolmRep! I am trying to run the two checks above. Will drop a few comments on the PR in the meantime. :) 

Thank you, my python 3.10.4 download failed on hep twice, at ~90% done.

setup.py
```diff
@@ -1,36 +1,35 @@
...
-    "torch-cluster==1.5.9",
+    "sqlalchemy>=1.4.37",
+    "pandas>=1.4",
+    "numpy>=1.22",
+    "timer>=0.2",
+    "tqdm>=4.64",
+    "dill>=0.3",
+    "wandb>=0.12",
+    "matplotlib>=3.5",
+    "scikit_learn~=1.1.1",
```
asogaard, line 24: Is there a reason, like for `torch` below, for using `~=` rather than  `>=`? For `torch` I figure it has to do with the CUDA version, but would it be a problem if someone manually (for whatever reason) wanted to go with `torch=1.10+cu113` rather than `torch=1.11+cu115`, as long as the pytorch code itself is compatible?

setup.py
```diff
@@ -45,15 +44,18 @@ def install(package):
...
-# Ensure pytorch is already installed (see e.g.
-# https://github.com/pyg-team/pytorch_geometric/issues/861#issuecomment-566424944)
-try:
-    import torch  # pyright: reportMissingImports=false
-except ImportError:
-    install("torch==1.9.0")
+# https://pypi.org/classifiers/
+CLASSIFIER = {
+    "Development Status :: 3 - Alpha",
+    "Programming Language :: Python :: 3.7"
```
asogaard, line 53: Right now we are testing for python versions 3.6-3.9; but I have considered going to 3.7-3.10. For now, I think we should add 3.6 to this list, and then we can always updated if we decide to support a different set of versions.

.gitignore
MortenHolmRep: This one is wholly related to my git inexperience, thinking it was just what was ignored locally on my end and what you write makes great sense.

setup.py
MortenHolmRep: The specification of version for the pytorch geometric related packages (and pytorch geometric itself) are based on past experience, where the geometric packages are not very flexible.
Further, it seems the torch packages in general often fall into paradigms, where for example torch 1.10 and 1.11 are specifically needed for versions of the torch geometric and cu*** installations.
So the strict specification was more to make sure that the version was always correct.

setup.py
MortenHolmRep: I thought of it mostly as flexibility. If for whatever reason the setuptools import failed, that an alternative setup was called.
But I see now that the way we have find_packages, makes this moot.

README.md
```diff
@@ -9,23 +9,45 @@
...
 $ cd graphnet
-$ conda env create -f envs/gnn_py38.yml
-$ conda activate gnn_py38
-(gnn_py38) $ pip install -e .[develop]
+$ conda create --name graphnet python=3.8 gcc_linux-64 gxx_linux-64 libgcc -y
+$ conda activate graphnet
+(graphnet) $ pip install -r requirements/torch_[gpu/cpu].txt -e .[develop,torch]
 ```
 
 This should allow you to e.g. run the scripts in [examples/](./examples/) out of the box.
```
asogaard, line 25:
```suggestion
This should allow you to e.g. run the scripts in [examples/](./examples/) out of the box. Here, we have installed recent C++ compilers using conda (`gcc_linux-64 gxx_linux-64 libgcc`), but if your system already have recent versions (`$gcc --version` should be > 5, at least) you should be able to omit these from the setup.
```

README.md
```diff
@@ -9,23 +9,45 @@
...
 
-You can also install the package in a python virtual environment, or in your system python, but then you will have to contend with C++ compiler versions; the non-standard interplay between [pytorch](https://pytorch.org/) and [pytorch-geometric](https://pytorch-geometric.readthedocs.io/en/latest/) (see e.g. [here](https://github.com/pyg-team/pytorch_geometric/issues/861#issuecomment-566424944)), which `graphnet` uses internally; etc.
+### :gear: Database creation and Installation on IceTray
```
asogaard, line 27:
```suggestion
### Installing with IceTray
```

README.md
```diff
@@ -9,23 +9,45 @@
...
+
+Within the IceTray enviroment, the Graphnet module is used to create SQL databases from I3 files, using the provided 'convert_i3_to_sqlite.py' script. The recommended set up is as follows:
```
asogaard, line 29:
```suggestion
In some instances, you might want `graphnet` to be able to interface with IceTray, e.g., when converting I3 files to an intermediate file format for training GNN models(e.g., SQLite or parquet), as shown in the [examples/convert_i3_to_sqlite.py](examples/convert_i3_to_sqlite.py) script, or when running GNN inference as part of an IceTray chain. In these cases, you need to install `graphnet` in a python runtime that has IceTray installed.

To achieve this, we recommend running the following commands in a clean bash shell:

File: README.md [hunk 11]
```diff
@@ -9,23 +9,45 @@
...
+
+```bash
+$ cd graphnet
+$ nano cvmfs.sh
+```
+
+Within the 'cvmfs.sh' copy the following and press 'ctrl+X'
```
asogaard, line 36:
```suggestion
```

README.md
```diff
@@ -9,23 +9,45 @@
...
+```bash
+eval `/cvmfs/icecube.opensciencegrid.org/py3-v4.1.0/setup.sh`
+/cvmfs/icecube.opensciencegrid.org/py3-v4.1.0/RHEL_7_x86_64/metaprojects/combo/stable/env-shell.sh
```
asogaard, line 39:
```suggestion
$ eval `/cvmfs/icecube.opensciencegrid.org/py3-v4.1.0/setup.sh`
$ /cvmfs/icecube.opensciencegrid.org/py3-v4.1.0/RHEL_7_x86_64/metaprojects/combo/stable/env-shell.sh
```

README.md
```diff
@@ -9,23 +9,45 @@
...
+```
+
```
asogaard, line 41:
```suggestion
Optionally, you can alias these commands or save them as a bash script for convenience, as you will have to run these commands every time you want to use IceTray (with `graphnet`) in a clean shell.

```

README.md
```diff
@@ -9,23 +9,45 @@
...
+Next, launch it and install the Graphnet module at a user level.
```
asogaard, line 42:
```suggestion
With the IceTray environment active, you can now install `graphnet` at a user level. In the example below, we are installing a light-weight version of `graphnet` without the `torch` extras, i.e., without the machine learning packages (pytorch and pytorch-geometric). This is useful when you just want to convert data from I3 files to, e.g., SQLite, and won't be running inference on I3 files later on. In this case, you don't need to specify a requirements file, compared to the example below.
```

README.md
```diff
@@ -9,23 +9,45 @@
...
+```bash
+$ source cvmfs.sh
+$ conda create --name graphnet_icetray
+$ conda activate graphnet_icetray
+$ pip install --user -e .[develop]
+```
 
+This should allow you to run 'convert_i3_to_sqlite.py' in [examples/](./examples/) with your preferred I3 files.
```
asogaard, line 50:
```suggestion
This should allow you to run the [examples/convert_i3_to_sqlite.py](examples/convert_i3_to_sqlite.py) script with your preferred I3 files.
```

asogaard (commented):
Hi @MortenHolmRep,

Thanks for the massive work done here! I have added a few last comments, mainly textual stuff for the README. Let me know once you have had a chance to look them over, then I'll approve this puppy! 🥳

examples/train_model.py
MortenHolmRep: File has been reverted

asogaard (approved):
🎆

Issue #214, opened 2022-06-03
Add option to read all fields from I3 files


asogaard:
Rather than specifying pulse maps, features, etc. similar to existing (in icetray) HDF5 converter. Could be done by making an `EverythingExtractor`.

RasmusOrsoe:
@asogaard  I'll take a look at this

PR #215, opened 2022-06-03
ParquetDataConverter and refactoring DataConverter

asogaard:
FYI @RasmusOrsoe


examples/convert_i3_to_parquet.py
```diff
@@ -0,0 +1,74 @@
...
+def main_icecube86():
+    """Main script function."""
+    paths = [
+        "/groups/icecube/asogaard/data/i3/i3_to_sqlite_workshop_test/level7_v02.00"
+    ]
+    pulsemap = "SRTInIcePulses"
+    gcd_rescue = "resources/GeoCalibDetectorStatus_AVG_55697-57531_PASS2_SPE_withScaledNoise.i3.gz"
+    outdir = "/groups/icecube/asogaard/temp/parquet_test_ic86"
+
+    converter = ParquetDataConverter(
```
RasmusOrsoe, line 28: Would be really cool, and probably preferable, if we could create an extractor `extract_everything_in_frame` that would just loop over frames. It would make the user experience a lot nicer.

RasmusOrsoe (commented):
Hey! I think this looks really nice.

If we want the "full" advantage of the parquet format, we should probably consider to just extract everything available in the i3 files. We could simplify the conversion code in the repo by a lot if we chose to do that. Then we could retire sqlite, or rewrite the sqlite conversion to work from parquet files, as that would be easier too.

asogaard:
Yes, agreed! I think that extracting everything from the I3 files is separate from the format to which we subsequently save the data, right? Because I remember all data also being saved to HDF5. That is why I have created #214, which could then be used as the sole extractor, that takes everything from the I3 files and dumps it in a parquet file (or SQLite database, for that matter). Does that agree with what you have in mind? :)

RasmusOrsoe:
> Yes, agreed! I think that extracting everything from the I3 files is separate from the format to which we subsequently save the data, right? Because I remember all data also being saved to HDF5. That is why I have created #214, which could then be used as the sole extractor, that takes everything from the I3 files and dumps it in a parquet file (or SQLite database, for that matter). Does that agree with what you have in mind? :)

Almost. I think it would be very challenging to make the SQLite format extend to _whatever_ is in the files.

asogaard:
Hi @RasmusOrsoe,

I have _finally_ prepared the PR for review. It's a massive one, so I think it could make sense for us to go over it in 1:1/in-person. The most substantial changes are:

* Rewrite of `DataConverter` class
* Rewrite of `SQLiteDataConverter` class to properly subclass `DataConverter`
* Addition of `ParquetDataConverter` class
* Addition of `Dataset` class
* Rewrite of `SQLiteDataset` class to subclass `Dataset`
* Addition of `ParquetDataset` class
* Restructuring `graphnet.data` module to make it easier to navigate
* Addition of example to compare the two sets of `DataConverter`/`Dataset` classes
* Unit tests for both "backend" file formats

examples/convert_i3_to_sqlite.py
```diff
@@ -60,8 +60,10 @@ def main_icecube_upgrade():
         ],
         outdir,
         gcd_rescue,
-        db_name=db_name,
         workers=workers,
+        nb_files_to_batch=1000,
+        # sequential_batch_pattern="temp_{:03d}",
+        input_file_batch_pattern="[A-Z]{1}_[0-9]{5}*.i3.zst",
```
RasmusOrsoe, line 66: Can we add a comment to explain this pattern? Not clear to me

RasmusOrsoe (approved):
Hey -

I think this is a nice refactoring. Only thing I could spot was the file pattern (see comment)

Rasmus

examples/convert_i3_to_sqlite.py
asogaard: Added to the `DataConverter` docstring.

PR #217, opened 2022-06-14
Parquet interface

RasmusOrsoe:
Here is a draft PR for a simple parquet dataloader. @asogaard fyi


asogaard (approved):
Thanks for this PR, @RasmusOrsoe! As an initial interface, while we're working on 
#215, I think this looks perfect! I say we merge it and then I'll make sure that it plays nicely with a ParquetDataConverter once we get there.

PR #218, opened 2022-06-15
Code for Marc Jacquart's analysis

### LINKED ISSUE #216: Make dataset class fail gracefully in case of missing truth variables
As mentioned in #169, there is a clear chance that e.g. SQLite database produced before a change in the I3TruthExtractor will not contain all of the expected/nominal truth variables listed in `src/graphnet/data/constants.py`. This is most often not an issue, but currently the run fails in a truth variable is missing. Instead, we should have the dataset produce a warning about missing variables but not break by default. _If_ the missing variable is being referenced downstream, e.g. in the training script, then an appropriate error will be produced there.

asogaard:
Initial version of code for @marcjacquart 's NMO sensitivity analysis in IceCube-Upgrade.

* Refactors the `I3Extractor` classes from `src/graphnet/data/i3extractor.py` into separate files in ``src/graphnet/data/extractors/` to make them more manageable. Should also have updated all references to these in database-generating scripts, etc.
* Add reproducible analysis workflow in `studies/upgrade_nmo_sensitivity/` for:
    * keeping track of data provenance (`datasets/upgrade_nmo_sensitivity.yml`)
    * generating database using `GraphSAGE`-cleaned pulses (courtesy of @kotania) in IceCube-Upgrade `nu_{mu}` events.
    * training energy, track-energy, zenith, and inelasticity regression models
    * resolution plotting (not tested)
* Adds `InelasticityReconstruction` task.
* Adds a bit of functionality to perform train/test split based on the modulus of `event_no` and save these and (sub)run/event IDs to file for keeping track of the data split in down-stream performance studies.
* Adds general (R)MSE losses, and renames `EuclideanDistance` -> `EuclideanDistanceLoss` to be consistent with naming.
* Discontinues the `frame_has_key(frame, key)` function in favour of the more transparent `key in frame` syntax.
* Closes #216 by warning-of-then-ignoring missing features rather than throwing an error. Means that we don't have to do `truth = features.TRUTH[:-1]` for missing `interaction_time`, etc.


RasmusOrsoe (approved):
Quite a mouthful to go through!

_"Adds a bit of functionality to perform train/test split based on the modulus of event_no and save these and (sub)run/event IDs to file for keeping track of the data split in down-stream performance studies."_
 Sounds interesting! Maybe you can elaborate on this for the next meeting?

_"Discontinues the frame_has_key(frame, key) function in favour of the more transparent key in frame syntax."_
**Shakes**

_"Closes #216 "_ 
Nice!

asogaard:
I have furthermore added an example script that shows how to use the upgrade models for inference in icetray, and saved the model in a "cleain" way, i.e. such that they don't accidentally carry around information on the pytorch-lightning trainer, dataloader, etc. used for training (which they, bafflingly, do if they are saved straight to file after training). Will look into torchscript soon as a way to make inference many times easier...

@RasmusOrsoe, did you want to discuss this PR at Tuesday's meeting, or is it "safe" to merge?

PR #234, opened 2022-07-21
Weighted Loss

RasmusOrsoe:
This PR lets us weigh the loss in training.

The `_forward` is changed to `_forward(self, prediction: Tensor, target: Tensor, weights: Tensor ) -> Tensor:` for all current functions, where `weights` is the weights. The output of the loss functions is altered such that `return elements*weights`. 

Because loss functions doesn't have access to the full `Data` object, the `weights` tensor is passed to the loss functions from the Task objects instead. The Task will pass the weights from `Data[self._loss_weight_column]` if `self._loss_weight_column` is specified (defaults to None) and if not specified,  `weights` is passed as `weights = torch.repeat(1, len(pred),dtype=torch.int).reshape(-1, 1)` - so weights of 1. 

In this PR, the ` Data[self._loss_weight_column]` is assigned in the `SQLiteDataset` which has three new inputs introduced; 

```
loss_weight_table: str = None,
loss_weight_column: str = None,
loss_weight_padding_value: float = 1.0
```

if both `loss_weight_table` and `loss_weight_column` is specified, the SQLiteDataset will query `loss_weight_table`  for `loss_weight_column` and add it to the graph as `Data[loss_weight_column] = weight`. If SQLiteDataset queries `loss_weight_table` but finds no weights for the specific event, the weight is set to `loss_weight_padding_value` as `Data[loss_weight_column] = loss_weight_padding_value`

In a future PR, I think we should add the ability to pass a `weight_function(features,truth)` to `SQLiteDataset` such that the above logic is overwritten if such a function is specified, and the weight for an event could then be set as `Data[loss_weight_column] = weight_function(features,truth)`


src/graphnet/data/sqlite_dataset.py
```diff
@@ -30,6 +30,9 @@ def __init__(
         node_truth: Optional[List[str]] = None,
```
asogaard, line 30: I think this file has moved due to #215. Could you merge this feature branch with `main` to make sure that we are not adding a second `sqlite_dataset.py` file, much like what happened with the I3Extractors, cf. the discussion with Kunal?

src/graphnet/models/task/task.py
```diff
@@ -64,6 +64,7 @@ def __init__(
         transform_target: Optional[Callable] = None,
         transform_inference: Optional[Callable] = None,
         transform_support: Optional[tuple] = None,
+        loss_weight_column: Optional[str] = None,
```
asogaard, line 67: I would suggest that we don't use "column" in the naming. As I see it, "column" is something in a database, whereas here it is an attribute of a graph (a `Data` object).
```suggestion
        loss_weight: Optional[str] = None,
```

src/graphnet/models/task/task.py
```diff
@@ -78,6 +79,7 @@ def __init__(
         self._target_labels = target_labels
         self._loss_function = loss_function
         self._inference = False
+        self._loss_weight_column = loss_weight_column
```
asogaard, line 82:
```suggestion
        self._loss_weight = loss_weight
```

src/graphnet/models/task/task.py
```diff
@@ -118,7 +120,18 @@ def compute_loss(self, pred: Union[Tensor, Data], data: Data) -> Tensor:
             [data[label] for label in self._target_labels], dim=1
         )
         target = self._transform_target(target)
-        loss = self._loss_function(pred, target) + self._regularisation_loss
+        if self._loss_weight_column is not None:
+            weights = data[self._loss_weight_column]
```
asogaard, line 124:
```suggestion
        if self._loss_weight is not None:
            weights = data[self._loss_weight]
```

src/graphnet/models/task/task.py
```diff
@@ -118,7 +120,18 @@ def compute_loss(self, pred: Union[Tensor, Data], data: Data) -> Tensor:
...
+        else:  # if no weights are given, applies weights of 1.
+            weights = (
+                torch.ones(len(pred), dtype=torch.int)
+                .reshape(-1, 1)
+                .to(pred.device)
+            )
```
asogaard, line 130: Default behaviour should be handled by `LossFunction`, I figure?
```suggestion
        else:
            weights = None
```

src/graphnet/components/loss_functions.py
```diff
@@ -31,6 +31,7 @@ def forward(
         self,
```
asogaard, line 31: In this module, I would suggest changing where the weights are multiplied from the within `_forward` method of the classes inheriting from `LossFunction` to the `forward` method, i.e., [here](https://github.com/icecube/graphnet/blob/main/src/graphnet/components/loss_functions.py#L48). If you change `LossFunction.forward` [here](https://github.com/icecube/graphnet/blob/main/src/graphnet/components/loss_functions.py#L30-L35) to have syntax `(self, prediction: Tensor, target: Tensor, weights: Optional[Tensor] = None, return_elements: bool = False)` and handle the default behaviour of `weights=None` to create a Tensor of all-ones, and then multiply these weights onto the output of `_forward`, you can leave all of the other classes unchanged. This is possible because `_forward` always return the per-sample losses. I think that would be a lot cleaner. And having a sensible default behaviour for `weight=None` implemented here means that you don't have to do this elsewhere, e.g. in the `Task` class.

asogaard (changes requested):
Hi Rasmus,

Thanks for preparing this! This is a _highly_ overdue feature, and I really appreciate you taking the lead here! 🥳 

I have made a few suggestions that I hope you will take a look at. Let me know what you make of it!

RasmusOrsoe:
@asogaard have updated the code following your comments. Code runs

src/graphnet/components/loss_functions.py
```diff
@@ -45,15 +47,19 @@ def forward(
             Tensor: Loss, either averaged to a scalar (if `return_elements = False`)
                 or elementwise terms with shape [N,] (if `return_elements = True`).
         """
-        elements = self._forward(prediction, target)
+        elements = self._forward(prediction, target)  # , weights)
```
asogaard, line 50:
```suggestion
        elements = self._forward(prediction, target)
```

src/graphnet/components/loss_functions.py
```diff
@@ -45,15 +47,19 @@ def forward(
...
             dim=0
         ), "`_forward` should return elementwise loss terms."
 
         return elements if return_elements else torch.mean(elements)
 
     @abstractmethod
-    def _forward(self, prediction: Tensor, target: Tensor) -> Tensor:
+    def _forward(
+        self, prediction: Tensor, target: Tensor, weights: Tensor
+    ) -> Tensor:
```
asogaard, line 62: I suppose this should be reverted given the rewrite?
```suggestion
    def _forward(self, prediction: Tensor, target: Tensor) -> Tensor:
```

src/graphnet/data/dataset.py
```diff
@@ -30,6 +29,9 @@ def __init__(
         string_selection: Optional[List[int]] = None,
         selection: Optional[List[int]] = None,
         dtype: torch.dtype = torch.float32,
+        loss_weight_table: str = None,
+        loss_weight_column: str = None,
+        loss_weight_padding_value: float = 1.0,
```
asogaard, line 34: It seems that `loss_weight_padding_value` is only used if `loss_weight_column is None`, i.e. as a default value. Given that `LossFunction.forward` already has a sensible default behaviour (all samples weights are one when `weights=None` is passed) I suggest we remove the default-behaviour-handling code from here so as to avoid duplication. My suggestions below reflect this.
```suggestion
```

src/graphnet/data/dataset.py
```diff
@@ -46,6 +48,7 @@ def __init__(
         self._truth = [index_column] + truth
         self._index_column = index_column
         self._truth_table = truth_table
+        self._loss_weight_padding_value = loss_weight_padding_value
```
asogaard, line 51:
```suggestion
```

src/graphnet/data/dataset.py
```diff
@@ -232,13 +248,23 @@ def _query(
...
             node_truth = None
-        return features, truth, node_truth
+
+        if self._loss_weight_column is not None:
+            if self._loss_weight_table is not None:
+                loss_weight = self._query_table(
+                    self._loss_weight_table, self._loss_weight_column, index
+                )
+        else:
+            loss_weight = [self._loss_weight_padding_value]
```
asogaard, line 258:
```suggestion
            loss_weight = None
```

src/graphnet/data/dataset.py
```diff
@@ -232,13 +248,23 @@ def _query(
...
+
+        return features, truth, node_truth, loss_weight
 
     def _create_graph(
         self,
         features: List[Tuple[Any]],
         truth: List[Tuple[Any]],
         node_truth: Optional[List[Tuple[Any]]] = None,
+        loss_weight: float = None,
```
asogaard, line 267:
```suggestion
        loss_weight: Optional[float] = None,
```

src/graphnet/data/sqlite/sqlite_dataset.py
```diff
@@ -59,6 +59,8 @@ def _query_table(
         # Query table
         self._establish_connection(index)
         try:
+            # a = f"SELECT {columns} FROM {table} WHERE {self._index_column} = {index} and {selection}"
+            # print(a)
```
asogaard, line 63:
```suggestion
```

src/graphnet/data/dataset.py
```diff
@@ -283,6 +310,16 @@ def _create_graph(
         graph = Data(x=x, edge_index=None)
         graph.n_pulses = n_pulses
         graph.features = self._features[1:]
+        # Write add loss weight to graph.
+        if loss_weight is not None and self._loss_weight_column is not None:
+            if len(loss_weight) == 0:
+                graph[self._loss_weight_column] = torch.tensor(
+                    self._loss_weight_padding_value, dtype=self._dtype
+                ).reshape(-1, 1)
```
asogaard, line 318: I am not sure what this guards against, if `loss_weight` is either None or a float, see the type hints and docstring?

asogaard (commented):
Thanks for the update, @RasmusOrsoe! I think this looks very clean. I have added just a few suggestions to possible clean this up just a bit more, but I think this is very close to merging. 🥳

src/graphnet/data/dataset.py
RasmusOrsoe: Not entirely. The padding value is assigned as weight for events that has no weights in the weight table. So if the query is empty.

asogaard: Brilliant, following our discussion there are two potential use cases/options:

1. You either specify a sample weight for all events in your dataset. In this case a missing weight would be a sign of an error, and we would want to alert the user to this.
2. You knowingly only specify a sample weight for a subset of your data (e.g., only for neutrino MC events when training on a mix of neutrino MC and real data). In this case a missing weight is expected and not _necessarily_ a sign of an error.

I would always recommend option (1.), but I acknowledge that it might be useful to leave option (2.) open for those that are brave enough to implicitly ignore potential missing weights.

In this case, I thing we should rename the variable (I understand "padding" to be something different, at least) and make it optional. If the variable is `None` and missing weights are found, this should raise an exception. This ought to be the default behaviour. If the variable is not `None`, but instead a float, this number should be used as the weight for all events with a "missing" entry in the weight column.

```suggestion
        loss_weight_default_value: Optional[float] = None,
```

src/graphnet/data/dataset.py
```diff
@@ -283,6 +310,16 @@ def _create_graph(
         graph = Data(x=x, edge_index=None)
         graph.n_pulses = n_pulses
         graph.features = self._features[1:]
+        # Write add loss weight to graph.
```
asogaard, line 313:
```suggestion

        # Add loss weight to graph.
```

src/graphnet/data/dataset.py
```diff
@@ -283,6 +310,16 @@ def _create_graph(
...
+        if loss_weight is not None and self._loss_weight_column is not None:
+            if len(loss_weight) == 0:
```
asogaard, line 315:
```suggestion
            # No loss weight was retrieved, i.e., it is missing for the current event
            if len(loss_weight) == 0:
```

src/graphnet/data/dataset.py
asogaard, line 318:
```suggestion
                if self._loss_weight_default_value is None:
                    raise ValueError(
                        "At least one event is missing an entry in "
                        f"{self._loss_weight_column} "
                        "but loss_weight_default_value is None."
                    )
                graph[self._loss_weight_column] = torch.tensor(
                    self._loss_weight_default_value, dtype=self._dtype
                ).reshape(-1, 1)
```

asogaard (commented):
Thanks for the update, @RasmusOrsoe! I think this looks very clean. I have added just a few suggestions following our discussion to possible clean this up just a bit more, but I think this is very close to merging. 🥳

RasmusOrsoe:
@asogaard thanks for comments. Have added your suggestions

asogaard (approved):
🥳

PR #237, opened 2022-08-02
Pisa weight fitter

RasmusOrsoe:
adds the ability to use pisa to fit physical weights to each events in a database


src/graphnet/pisa/fitting.py
```diff
@@ -15,6 +15,7 @@
 import io
 from configupdater import ConfigUpdater
 from contextlib import contextmanager
+from graphnet.data.utils import run_sql_code, save_to_sql
```
asogaard, line 18: Due to refactor in #215 
```suggestion
from graphnet.data.sqlite import run_sql_code, save_to_sql
```

src/graphnet/pisa/fitting.py
```diff
@@ -74,6 +75,174 @@ def config_updater(
                 updater.write(configfile)
 
 
+class WeightFitter:
+    def __init__(
```
asogaard, line 79: A docstring here, or in `fit_weights`, would be helpful to explain which weights are being fitted, exactly.

src/graphnet/pisa/fitting.py
```diff
@@ -74,6 +75,174 @@ def config_updater(
...
+        self,
+        database_path,
```
asogaard, line 81: Here and elsewhere we are assuming that databases (as opposed to some other intermediate file format) are being used to interface with PISA. Is that something we are committing to, or just something that works for the time being but might be expanded on down the line?

asogaard (commented):
Hi Rasmus,
I don't have a good handle on the PISA portion of the repo, so I am happy to merge this PR without much back-and-forth. I have made one suggestion though, to update an import that was changed on a recent PR, that I suggest we add before merging. I have also suggested adding just a bit of documentation to explain the intent behind the `WeightFitter` class, as this is not super clear to me at present, but I'll leave that up to you. :)

asogaard (approved):

src/graphnet/pisa/fitting.py
RasmusOrsoe: This is something we can expand upon down the line, I think. Currently, GraphNeT is only interfaced with PISA via SQL. To make it work with a new format, we need to update PISA as it relies on a dataloader for each format. See https://github.com/icecube/pisa/tree/master/pisa/stages/data for reference. They are easy to write, so it shouldn't be a big thing.

PR #243, opened 2022-08-09
Close SQLite connection after construction to avoid problem with multiprocessing

asogaard:
After #215 the `SQLiteDataset` was unintentionally created with the `_conn` member variable set, since `_remove_missing_variables` established a connection to the SQLite database, but didn't close it again, resulting in the error:
```
sqlite3.DatabaseError: database disk image is malformed
```
when using more than one worker process in `DataLoader`. By now calling `SQLiteDataset._close_connection()` after initialisation this problem should be fixed.

Also:
* Renamed `_initialise` to `_init` to consistency with new method
* Wrapped a few lines to get them within the 79 character limit

### DIFF
--- a/src/graphnet/data/dataset.py
+++ b/src/graphnet/data/dataset.py
@@ -74,8 +74,8 @@ class Dataset(ABC, torch.utils.data.Dataset, LoggerMixin):
 
         self._dtype = dtype
 
-        # Implementation specific initialisation.
-        self._initialise()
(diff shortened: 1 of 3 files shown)


RasmusOrsoe (approved):
Thanks for looking into this @asogaard

RasmusOrsoe:
@asogaard fyi, can confirm this indeed fixes the issue

Issue #262, opened 2022-08-23
Make training scripts simpler


RasmusOrsoe:
In `/examples/train_model.py` the MWE of a training script is 150 lines of code. If we introduce a default configuration for each model (`model.config`), a `Session` class we could simplify this greatly. Here is a bit of conceptual code

```
from graphnet.models.training.utils import TrainingSession, InferenceSession, make_dataloaders
from graphnet.models.gnn import DynEdge
from graphnet.components.loss_functions import LogCoshLoss
from graphnet.models.task.reconstruction import EnergyReconstruction
from graphnet.models.detector.icecube import IceCubeDeepCore

# Main function definition
def main(model_config = None):
    
    archive = "/groups/icecube/asogaard/gnn/results/"
    run_name = "dynedge_{}_example".format(config["target"])
    selection = None
    
    model = DynEdge()
    # Configuration
    if model_config is None:
        model.config['loss_function'] = LogCoshLoss
        model.config['task'] = EnergyReconstruction
        model.config['data_path'] = "/groups/icecube/asogaard/data/sqlite/dev_lvl7_robustness_muon_neutrino_0000/data/dev_lvl7_robustness_muon_neutrino_0000.db"
        model.config['pulsemap'] = 'SRTTWOfflinePulsesDC'
        model.config['detector'] = IeCubeDeepCore
        model.config['selection'] = selection
    else:
        model.config = model_config
    
    # Make DataLoaders
    train_dataloader, valid_dataloader, test_dataloader = make_dataloaders(model)

    # Setup Sessions
    training_session = TrainingSession(model, archive, run_name)
    inference_session = InferenceSession(model, archive, run_name)
        
    # training session
    trained_model = training_session.start(train_dataloader = train_dataloader, validation_dataloader = valid_dataloader)
    
    # inference session
    if trained_model.converged:
        validation_results = inference_session.start(model = trained_model, dataloader = valid_dataloader, archive, run_name, tag = 'validation')
        test_results = inference_session.start(model = trained_model, dataloader = valid_dataloader, archive, run_name, tag = 'test')


    # save results to csv
    save_results(validation_results, trained_model, run_name, archive)
    save_results(test_results, trained_model, run_name, archive)


# Main function call
if __name__ == "__main__":
    main()
```

asogaard:
[...]

PR #266, opened 2022-08-25
Implement I3GenericExtractor

### LINKED ISSUE #214: Add option to read all fields from I3 files

asogaard:
`I3Extractor` class to automatically extract all information from I3Frame objects. Useful for when you just want to convert an I3 file to an intermediate format (e.g., Parquet, SQLite) without having to tinker too much. Nested/non-trivial datatypes (e.g. lists-of-lists) are serialised to JSON to allow for saving. Checks:
- [x] The extraction itself works on typical P-frames
- [x] Extracted data can be saved to both Parquet and SQLite format.
- [x] Extracted data is exactly the same as when using a regular `I3FeatureExtractor` and/or `I3TruthExtractor`
- [x] Saved data works with `ParquetDataset` and `SQLiteDataset`.

Closes #214


examples/convert_i3_files.py
```diff
@@ -0,0 +1,95 @@
...
+        # nb_files_to_batch=10,
+        # sequential_batch_pattern="temp_{:03d}",
+        # input_file_batch_pattern="[A-Z]{1}_[0-9]{5}*.i3.zst",
+        icetray_verbose=1,
+    )
+    converter(paths)
+
+
+if __name__ == "__main__":
+    backend = "parquet"
```
RasmusOrsoe, line 92: Thats sexy!

src/graphnet/data/extractors/i3genericextractor.py
```diff
@@ -0,0 +1,205 @@
...
+    )  # pyright: reportMissingImports=false
+except ImportError:
+    logger.warning("icecube package not available.")
+
+
+GENERIC_EXTRACTOR_NAME = "<GENERIC>"
+
+
+class I3GenericExtractor(I3Extractor):
+    """Dynamically and generically extract all information from frames."""
```
RasmusOrsoe, line 31: Last we spoke, there were indeed a few keys we could not extract generically. If this is still the case, it should be clear from this docstring which fields we leave unextracted.

src/graphnet/data/extractors/i3genericextractor.py
```diff
@@ -0,0 +1,205 @@
...
+            exclude_keys = [exclude_keys]
+
+        # Reference to frame currently being processed
+        self._keys: List[str] = keys
+        self._exclude_keys: List[str] = exclude_keys
+
+        super().__init__(GENERIC_EXTRACTOR_NAME)
+
+    def _get_keys(self, frame: "icetray.I3Frame") -> List[str]:
+        """Get the effective list of keys to be queried from `frame`."""
```
RasmusOrsoe, line 58: what does "effective list of keys" mean? Would be in favor of making this docstring clearer

RasmusOrsoe: "Removes i3 fields that the user specified not to be extracted from the list of i3 fields" or similar

src/graphnet/data/extractors/i3genericextractor.py
```diff
@@ -0,0 +1,205 @@
...
+        if self._keys is None:
+            keys = list(frame.keys())
+            if self._exclude_keys is not None:
+                keys = [key for key in keys if key not in self._exclude_keys]
+        else:
+            keys = self._keys
+        return keys
+
+    def __call__(self, frame: "icetray.I3Frame") -> dict:
+        """Extract all possible data from `frame`."""
```
RasmusOrsoe, line 68: If some fields are not extractable by this class, perhaps repeating them here is a good idea.

src/graphnet/data/extractors/i3genericextractor.py
```diff
@@ -0,0 +1,205 @@
...
+                result = cast_pulse_series_to_pure_python(
+                    frame, key, self._calibration, self._gcd_dict
+                )
+
+                if result is None:
+                    self.logger.debug(
+                        f"Pulse series map {key} didn't return anything."
+                    )
+
+            # -- Per-pulse map
```
RasmusOrsoe, line 105: "event-level information" ?

src/graphnet/data/extractors/i3genericextractor.py
```diff
@@ -0,0 +1,205 @@
...
+                    self.logger.debug(
+                        f"Per-pulse map {key} didn't return anything."
+                    )
+
+                else:
+                    # If we get a per-pulse map, which isn't a
+                    # "I3RecoPulseSeriesMap*", we don't care about area,
+                    # direction, orientation, and position -- we only care
+                    # about the OM index for future reference. We therefore
+                    # only keep these indices and the associated mapping value.
```
RasmusOrsoe, line 129: I do not understand this. What is a per-pulse map?

src/graphnet/data/extractors/utilities/collections.py
```diff
@@ -0,0 +1,61 @@
...
+
+        return dict(items)
+
+    # Other non-dict like (list, str, float, int, etc.)
+    else:
+        return {parent_key: obj}
+
+
+def serialise(obj: Union[Dict, Any]) -> Union[Dict, Any]:
+    if isinstance(obj, MutableMapping):
```
RasmusOrsoe, line 32: Could we add a doc string here? Not clear what problem this solves

src/graphnet/data/extractors/utilities/types.py
```diff
@@ -0,0 +1,243 @@
...
+    if return_discarded:
+        return valid_member_variables, discarded_member_variables
+
+    return valid_member_variables
+
+
+def cast_object_to_pure_python(obj: Any) -> Any:
+    """Extracts all possible pure-python data from `obj`."""
+
+    logger.debug(f"Value: {obj}")
```
RasmusOrsoe, line 101: Could we expand on this doc string a bit? Just to make the recursive nature of this function a bit clearer

RasmusOrsoe: (Seems to be very central to the extractor, so putting in a nice fat doc string here is probably worthwhile)

RasmusOrsoe (changes requested):
Hey! I think this is a really nice contribution to the repo!!

I have added comments in a few places - mostly related to docstrings. 

Here are a few questions:

- Can you elaborate on how the SQLite table structure looks like when using this extractor? Do each flattened entry in `results` get their own table?
- What does the unticked boxes in your PR comment mean?

I think we should refactor `__call__` in  `I3GenericExtractor` to make the special cases easier to read. Below is a conceptual take on how that could look like. It would be a miracle if we didn't have to tinker with these parts in the future, so having them expcilitly written in private methods makes maintenence easier, i'd say. 
```

def __call__(self, frame: "icetray.I3Frame") -> dict:
    """Extract all possible data from `frame`. The following keys are not extracted : ..."""

    results = {}
    for key in self._get_keys(frame):
        # Extract object from frame
        try:
            obj = frame[key]
        except RuntimeError:
            self.logger.debug(
                f"Key {key} in frame not supported. Skipping."
            )
        except KeyError:
            if self._keys is not None:
                self.logger.warning(f"Key {key} not in frame. Skipping")
            continue

        # Special case(s)
        # -- Pulse series mao
        if isinstance(
            obj,
            (
                dataclasses.I3DOMLaunchSeriesMap,
                dataclasses.I3RecoPulseSeriesMap,
                dataclasses.I3RecoPulseSeriesMapMask,
                dataclasses.I3RecoPulseSeriesMapUnion,
            ),
        ):
            result = self._extract_pulsemap(args)
            
        # -- Per-pulse map (is this pulsemap attributes??)
        elif isinstance(
            obj,
            (
                dataclasses.I3MapKeyUInt,
                dataclasses.I3MapKeyDouble,
                dataclasses.I3MapKeyVectorInt,
                dataclasses.I3MapKeyVectorDouble,
            ),
        ):
            result = self._extract_pulsemap_attributes(args)

        # -- MC Tree
        elif isinstance(obj, dataclasses.I3MCTree):
            result = self._extract_mc_tree(args)

        # -- Triggers
        elif isinstance(obj, dataclasses.I3TriggerHierarchy):
            result = self._extract_trigger(args)

        # -- Generic case
        else:
            result = self._generic_extraction(args)

        # Skip empty extractions
        if result is None:
            continue

        # Flatten Result
        
        result = self._flatten_result(args)

    # Serialise list of iterables to JSON
    results = {key: serialise(value) for key, value in results.items()}

    return results
```

asogaard:
Hi @RasmusOrsoe,
Thanks for the detailed comments! I have tried to accommodate all of your suggestions, and have refactored the main `__call__` method according to your suggestion. Please check that you are happy with the changes. I am now comparing the results from the generic extractor and the current feature/truth extractors to make sure that they agree exactly. Once that's done, I'll add some unit tests to this PR to confirm. (See also checklist in description above.)

RasmusOrsoe:
Looks really nice now @asogaard!

asogaard:
@RasmusOrsoe I have had to remove the `_get_labels` and `_get_dbang_label` methods in `Dataset` because they rely on hard-coded attribute names that we can no longer expect with `I3GenericExtractor`. I have added an `add_label` method that instead. See all changes since you last looked [here](https://github.com/icecube/graphnet/pull/266/files/b44a04c3b2a7e59f41a753d8d3a57703503bd4df..e7e0b30a3b8a9f75862df5b84b933296fdcedd7f).

RasmusOrsoe (commented):
Hey Andreas! Thanks for the changes. 

I understand that with a varying name convention between files, the label creation method is difficult to handle. This is one of the issues I had to deal with in the parquet stuff. By solving this with a public method, we put the responsibility to the user, which I think is a nice degree of freedom for niche cases. However, I think the replacement of  `_get_labels()` with a public method is a step we should consider carefully. On the user experience side, it complicates the training quite a bit as we would have to adopt `make_dataloader` and `make_train_validation_loaders` to accept multiple of such functions, and the user would have to know in advance how to make the labels - which in all honesty requires detailed knowledge of how we process data and what's called what. It would add several lines of code to training scripts that are already a bit too complicated. 

We could pack required label creation methods in the `Model` class in line with #262, perhaps even associate `Task`'s with labels creation methods, but I think that is a bit beyond the scope of this PR - but a future rework that I believe might be necessary. 

If we choose to proceed down this path, I think we should adjust `dataset`,  `make_dataloader` and `make_train_validation_dataloaders` to accept arguments `interaction_type = {table_name: column_name}` and `pid = {table_name: column_name}` such that we can create the minimal `muon` , `neutrino` and `track` labels in the dataset by default. We require these fields to be available in the parquet guidelines, so I think we can assume that they are. This way, the PR doesn't remove our classification functionality. 

And then those of us who want more exotic labels can use the public method. 

What do you think @asogaard  ?

asogaard:
Alright, @RasmusOrsoe, I have tried to work through the PR to create an `I3GenericExtractor` that does what it promises: It is able to read out all data from P-frames in I3Files and write them to SQLite and Parquet. I'll add a small unit test, but otherwise I think this is ready for review/discussion.

**One outstanding issue** which I'm inclined to leave open, is that this `I3GenericExtractor` does not provide all of the capabilities that the `I3TruthExtractor` does. This is because the former returns the parsed results of the `I3MCTree` which has a variable number of entries (for particles and primaries), whereas the latter return per-event quantities. For the purposes of this PR, I think it is OK that the `I3GenericExtractor` just does the extraction it promises to do. It then leaves open the question of how we want to transform the output of the `I3MCTree` into proper targets downstream. The good thing is that the `I3GenericExtractor` can be used seamlessly with the `I3TruthExtractor`, and I have updated the `covert_i3_files.py` example to reflect this. So, if we're happy with the `I3TruthExtractor` (which probably ought to be refactored into a number of smaller, task-specific extractors) we have what we need.

RasmusOrsoe (approved):
@asogaard Looks good. We can merge this if you want. But I'd like to point out that the changes made to the labels means that we remove out-of-the-box classification functionality from the repo. 

If we merge this, I think we should put it on high priority to follow this PR up with a `training/labels.py` that at the very minimum includes labels for muon/neutrino and track/cascade.

asogaard:
Hi @RasmusOrsoe,
Let's set aside some time to go over this before I merge. :)

To your last point: We can still do classification just like before using `I3TruthExtractor` alongside `I3GenericExtractor`. But you're right that the latter isn't (yet) a silver bullet. So I think we can keep using `I3TruthExtractor` and then think about the best way to either implement something like `training/labels.py` or tiny, single-purpose classes inheriting from `I3TruthExtractor` that basically just define the label(s) that a specific analysis needs.

PR #272, opened 2022-08-30
Add feature to cluster by time windows

mhaminh:
I added a Coarsening class to cluster by time windows. Basically you first cluster by DOMs, and then use DBSCAN to cluster in time. There's also a small hacky part in there so that everything can be clustered in one go, maybe there's a nicer way to do this. Also you can pass the size of the time window at the initialization of the coarsening class.

### DIFF
--- a/src/graphnet/components/pool.py
+++ b/src/graphnet/components/pool.py
@@ -107,6 +107,22 @@ def group_pulses_to_pmt(data: Data) -> Data:
     return data
 
 
+def time_window_clustering(data: Union[Data, Batch], cluster_method, time_window: float):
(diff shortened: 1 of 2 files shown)


asogaard (changes requested):
Hi @mhaminh,

Thanks for this PR! I think it is a nice feature to have. :)  I have added a few suggestions regarding the location of the code.

RasmusOrsoe:
@mhaminh  ping

mhaminh:
Sorry, I've been caught up in a few things, I'll push the updated changes soon

asogaard (approved):
:dab:

PR #277, opened 2022-08-31
Parquet to sqlite converter

RasmusOrsoe:
This PR brings a new class to graphnet.data.util.parquet_to_sqlite called `ParquetToSQLiteConverter`. It converts parquet file(s) to a SQLite database with indexation. It works for any parquet files that follow our guidelines. Code runs and verified that it indeed works. 

It even has graphnet-themed progress bars! 

billede


src/graphnet/data/utilities/parquet_to_sqlite.py
```diff
@@ -0,0 +1,227 @@
...
+            if parquet_path.endswith(".parquet"):
+                files = [parquet_path]
+            else:
+                files = glob.glob(f"{parquet_path}/*.parquet")
+        elif isinstance(parquet_path, list):
+            files = []
+            for path in parquet_path:
+                files.extend(glob.glob(f"{path}/*.parquet"))
+        assert len(files) > 0, f"No Files Found in {parquet_path}"
+        return files
```
asogaard, line 53:
```suggestion
    def _find_parquet_files(self, paths: Union[str, List[str]]) -> List[str]:
        if isinstance(paths, str):
            if paths.endswith(".parquet"):
                files = [paths]
            else:
                files = glob.glob(f"{paths}/*.parquet")
        elif isinstance(paths, list):
            files = []
            for path in paths:
                files.extend(self._find_parquet_files(path))
        assert len(files) > 0, f"No files found in {paths}"
        return files
```

src/graphnet/data/utilities/parquet_to_sqlite.py
```diff
@@ -0,0 +1,227 @@
...
+
+    def _create_table(
+        self,
+        df: pd.DataFrame,
+        field_name: str = None,
+        outdir: str = None,
+        database_name: str = None,
+        n_events_in_file: int = None,
+    ):
+        """Creates a table.
```
asogaard, line 160: This appears duplicate of https://github.com/graphnet-team/graphnet/blob/main/src/graphnet/data/sqlite/sqlite_dataconverter.py#L120-L146. Since `self` isn't being referenced, I figure we could convert this into a utility function in `graphnet.data.sqlite.sqlite_utilities` and use that method both here and in `SQLiteDataConverter`,

src/graphnet/data/utilities/parquet_to_sqlite.py
```diff
@@ -0,0 +1,227 @@
...
+            )
+            self._created_tables.append(field_name)
+            self._submit_to_database(outdir, database_name, field_name, df)
+
+    def _make_df(
+        self,
+        ak_array: awkward.Array = None,
+        field_name: str = None,
+        n_events_in_file: int = None,
+    ):
```
asogaard, line 109:
```suggestion
    def _convert_to_dataframe(
        self,
        ak_array: awkward.Array,
        field_name: str,
        n_events_in_file: int,
    ) -> pd.DataFrame:
```

src/graphnet/data/utilities/parquet_to_sqlite.py
```diff
@@ -0,0 +1,227 @@
...
+        df = pd.DataFrame(ak.to_pandas(ak_array[field_name]))
```
asogaard, line 110: Are both `pd.DataFrame(...)` and `ak.to_pandas(...)` necessary?

src/graphnet/data/utilities/parquet_to_sqlite.py
```diff
@@ -0,0 +1,227 @@
...
+        if len(df.columns) == 1:
+            if df.columns == ["values"]:
+                df.columns = [field_name]
+        if (
+            len(df) != n_events_in_file
+        ):  # if true, the dataframe contains more than 1 row pr. event (e.g. Pulsemap).
+            event_nos = []
+            for event_no in range(n_events_in_file + self._event_counter):
```
asogaard, line 118: Should this be something like `range(self._event_counter, n_events_in_file + self._event_counter)`? Right now it seems like all pulse maps are assigned event numbers `0, 1, ..., n_events_in_file` whereas I guess the desired behaviour is `self._event_counter, self._event_counter + 1, ..., self._event_counter + n_events_in_file - 1`.

asogaard (changes requested):
Hi @RasmusOrsoe,

Thanks for this PR, I think this is very nice functionality to have around! :)  I have proposed a few changes, and noted two places where I see code duplication of `SQLiteDataConverter` which I think is best handled by turning the relevant code into utility functions and calling it both places.

src/graphnet/data/utilities/parquet_to_sqlite.py
RasmusOrsoe: In my experience, yes.

src/graphnet/data/utilities/parquet_to_sqlite.py
RasmusOrsoe: Thank you for pointing this out. I have fixed it now.

RasmusOrsoe:
@asogaard I think this is everything

asogaard (approved):
Looks good! 👍

PR #279, opened 2022-09-02
Uniform weight fitter

RasmusOrsoe:
Adds a new module called weight_fitting with UniformWeightFitter that calculates pr. event weights such that the distribution of a chosen truth variable is uniform.


asogaard:
Hi @RasmusOrsoe,

Just a few questions here before I get into reviewing:
* Right now the code is written assuming a SQLite backend. If we want to support multiple (I think that is a good idea, at least for the time being) we should consider how that would interact with a weight fitter class. Should we instead take a `Dataset` class as argument, and iterate over that to fill the histograms?
    * Corollary: We should also consider what the best way is to handle weights. Should we have them as separate files (CSV or backend-specific)? Should we append them to the files being fitted? I think I am in favour of the latter, as it means that there are that fewer things to keep track of which is also nice fir reproducibility. But that means that we should have a generic interface to write to intermediate files, e.g., through the `Dataset` class.
* I figure we might want to use multiple different weighting schemes. Would it make sense to structure these under a `src/graphnet/models/training/weighting/` directory? Or should we "promote" training to `src/graphnet/training` and have it contain weighting schemes, training callbacks, loss functions, etc., to avoid too many nested directories?

RasmusOrsoe:
_Right now the code is written assuming a SQLite backend. If we want to support multiple (I think that is a good idea, at least for the time being) we should consider how that would interact with a weight fitter class. Should we instead take a Dataset class as argument, and iterate over that to fill the histograms?_
**I think that is a good idea. But it would also complicate development quite a bit. If we really want to support both sqlite/parquet in parallel, then I think your suggestion is the only sensible way forward  (NOT in this PR though). But if we choose to operate purely in sqlite for developers (so parquet -> sqlite via the new converter) then I think it's unnecessary. So there is a design choice here that needs to be taken. **
        _Corollary: We should also consider what the best way is to handle weights. Should we have them as separate files (CSV or backend-specific)? Should we append them to the files being fitted? I think I am in favour of the latter, as it means that there are that fewer things to keep track of which is also nice fir reproducibility. But that means that we should have a generic interface to write to intermediate files, e.g., through the Dataset class._
**CSV files does not scale well. Each worker will need a copy of the fully-in-memory file and that explodes quite quickly. So I opted for creating tables in the sqlite database and extract them in the dataset class. We could make a separate database with these weights and pass a database path. I know @mhaminh prefers to pass a list of weights to the training (csv-like) and I think that he has implemented this on a local branch. Perhaps we could allow for weights passing this way also, and let the user choose what they prefer. **
    _I figure we might want to use multiple different weighting schemes. Would it make sense to structure these under a src/graphnet/models/training/weighting/ directory? Or should we "promote" training to src/graphnet/training and have it contain weighting schemes, training callbacks, loss functions, etc., to avoid too many nested directories?_
**I would prefer to keep the number of directories at a minimum. I often struggle to find the proper files as is :-D . So I think a /training/weighting.py should do the trick**

asogaard:
Hi @RasmusOrsoe,

* I think it is fine to assume a SQLite backend for now, and have an idea of how a more generic fitter would be implemented.
* I also think it is fine to go with CSV-files now, and have an idea that this may not be the optimal solution.
* Brilliant! If you create `src/graphnet/training/` and move `weight_fitter.py` there, I will try to refactor the other relevant modules over there.

I'll do a quick review, then, but it looks like it should be quite minimal! :)

RasmusOrsoe:
@asogaard ready

asogaard (approved):
Hi @RasmusOrsoe,
Thanks so much for this! I see that I had a few comments pending, but so do with those what you want. :)

RasmusOrsoe:
@asogaard believe this is everything. Am i right?

src/graphnet/training/weight_fitting.py
```diff
@@ -0,0 +1,93 @@
...
+            query_columns.append(f"{column} {type_}")
+        query_columns = ", ".join(query_columns)
+
+        code = (
+            "PRAGMA foreign_keys=off;\n"
+            f"CREATE TABLE {table_name} ({query_columns});\n"
+            "PRAGMA foreign_keys=on;"
+        )
+        run_sql_code(database, code)
+        return
```
asogaard, line 93: This seems to be duplicate of similar methods in [`graphnet.data.pipeline`](https://github.com/graphnet-team/graphnet/blob/main/src/graphnet/data/pipeline.py#L208-L229) and [`graphnet.data.sqlite.sqlite_dataconverter`](https://github.com/graphnet-team/graphnet/blob/main/src/graphnet/data/sqlite/sqlite_dataconverter.py#L120-L152). I think it would be nice to promote this to a utility function.

src/graphnet/training/weight_fitting.py
```diff
@@ -0,0 +1,93 @@
...
+        if weight_name is None:
+            weight_name = variable + "_uniform_weight"
+        truth[weight_name] = sample_weights
+        truth = truth.drop(columns=[variable])
+        if add_to_database:
+            self._create_table(self._database_path, weight_name, truth)
+            save_to_sql(truth, weight_name, self._database_path)
+        return truth.sort_values("event_no").reset_index(drop=True)
+
+    def _get_truth(self, variable, selection):
```
asogaard, line 63:
```suggestion
    def _get_truth(self, variable: str, selection: Optional[List[int]] = None):
        """"Return truth `variable`, optionally only for `selection` event nos.""""
```

src/graphnet/training/weight_fitting.py
```diff
@@ -0,0 +1,93 @@
...
+        Args:
+            bins (numpy.array): An array containing the bins. E.g. numpy.arange(0,180,1)
+            variable (str): the name of the truth variable in truth_table
+            weight_name (str, optional): the assigned column name of the weights if add_to_database = True. Defaults to None.
+            add_to_database (bool, optional): If True, the weights are written to the database as a new table named weight_name . Defaults to False.
+            selection (List, optional): a List containing the event_no's to which the weights are calculated. Defaults to None.
+
+        Returns:
+            _type_: pandas.DataFrame
+        """
```
asogaard, line 42:
```suggestion
    def fit_weights(
        self,
        bins: np.ndarray,
        variable: str,
        weight_name: Optional[str] =None,
        add_to_database: bool =False,
        selection: Optional[List[int]] = None,
    ) -> pd.DataFrame:
        """Calculate per-event weights to make `variable` distribution uniform.
        
        If selection is None, the weights are calculated using all events in the 
        database.

        Args:
            bins (numpy.array): An array containing the bins. E.g. 
                numpy.arange(0,180,1)
            variable (str): the name of the truth variable in truth_table
            weight_name (str, optional): the assigned column name of the weights 
                if add_to_database = True. Defaults to None.
            add_to_database (bool, optional): If True, the weights are written 
                to the database as a new table named weight_name. Defaults to 
                False.
            selection (List, optional): a List containing the event_no's to 
                which the weights are calculated. Defaults to None.

        Returns:
            pandas.DataFrame
        """
```

asogaard (commented):
Hi @RasmusOrsoe,

I have added three (final) comments: Two are about documentation, the third about duplicate code. Let me know what you think. :)

asogaard:
🎉

PR #295, opened 2022-09-20
Direction example

RasmusOrsoe:
Contains an example of training a direction reconstruction ala what we showed for the bruxelles collab meeting.


asogaard (commented):
Hi @RasmusOrsoe,

I think this looks perfectly fine. It might be worth it to develop this into an actual analysis over in https://github.com/graphnet-team/analyses, but it's find to start out here. At some point, we should clean up our examples, but not right now. :)

asogaard (approved):

Issue #281, opened 2022-09-07
Error when saving results in train_model.py


MoustHolmes:
**Describe the bug**
When I try to run train_model.py everything runs except the last line where we save the results using the function save_results.
If I out-comment the last line the whole script runs without problems.
the error seams to occurs when we use the pytorch function model.save() within the save_results function which won't pickle a "sqlite3.Connection" ?

**Traceback**
```
Traceback (most recent call last):
  File "train_model.py", line 155, in <module>
    main()
  File "train_model.py", line 150, in main
    save_results(config["db"], run_name, results, archive, model)
  File "/lustre/hpc/icecube/moust/work/graphnet/src/graphnet/models/training/utils.py", line 206, in save_results
    model.save(path + "/" + tag + "_model.pth")
  File "/lustre/hpc/icecube/moust/work/graphnet/src/graphnet/models/model.py", line 147, in save
    torch.save(self.cpu(), path, pickle_module=dill)
  File "/groups/icecube/moust/miniconda3/envs/graphnet/lib/python3.8/site-packages/torch/serialization.py", line 380, in save
    _save(obj, opened_zipfile, pickle_module, pickle_protocol)
  File "/groups/icecube/moust/miniconda3/envs/graphnet/lib/python3.8/site-packages/torch/serialization.py", line 589, in _save
    pickler.dump(obj)
  File "/groups/icecube/moust/miniconda3/envs/graphnet/lib/python3.8/site-packages/dill/_dill.py", line 620, in dump
    StockPickler.dump(self, obj)
  File "/groups/icecube/moust/miniconda3/envs/graphnet/lib/python3.8/pickle.py", line 487, in dump
    self.save(obj)
```

After this, there are like 50 lines of slight variations of the last line with this being the last line

```
TypeError: cannot pickle 'sqlite3.Connection' object
```

MortenHolmRep:
I updated the framework to the recent iteration and get somewhat the same error as @MoustHolmes when running inference.
```
Traceback (most recent call last):
  File "/groups/icecube/qgf305/anaconda3/envs/graphnet/lib/python3.8/runpy.py", line 194, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/groups/icecube/qgf305/anaconda3/envs/graphnet/lib/python3.8/runpy.py", line 87, in _run_code
    exec(code, run_globals)
[...]

PR #297, opened 2022-09-21
update of direction example

RasmusOrsoe:
updates direction example such that it's compatible with main

### DIFF
--- a/examples/train_direction_reco.py
+++ b/examples/train_direction_reco.py
@@ -1,0 +1,203 @@
+import os
+from pytorch_lightning import Trainer
+from pytorch_lightning.callbacks import EarlyStopping
+from pytorch_lightning.loggers import WandbLogger
+import torch
+from torch.optim.adam import Adam
+import pandas as pd
+from graphnet.components.loss_functions import VonMisesFisher2DLoss
+from graphnet.data.constants import FEATURES, TRUTH
+from graphnet.data.sqlite.sqlite_selection import (
+    get_equal_proportion_neutrino_indices,
+)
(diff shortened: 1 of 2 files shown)


examples/train_direction_reco.py
```diff
@@ -48,13 +49,22 @@
 )
 
 
+def save_results(db, tag, results, archive):
+    db_name = db.split("/")[-1].split(".")[0]
+    path = archive + "/" + db_name + "/" + tag
+    os.makedirs(path, exist_ok=True)
+    results.to_csv(path + "/results.csv")
+    print("Results saved at: \n %s" % path)
+
```
asogaard, line 58: This seems duplicate of the function by the same name in [src/graphnet/models/training/utils.py](https://github.com/graphnet-team/graphnet/blob/97110e0d1286fdc9477cd49ef77f2b40963efcf0/src/graphnet/models/training/utils.py)

asogaard (changes requested):
Hi @RasmusOrsoe,
I noticed one seemingly duplicate function. Otherwise, I don't have any strong opinions on the example code. :)

examples/train_direction_reco.py
RasmusOrsoe: I know. I started getting the pickle error that Moust mentioned. Using this removed the error.

asogaard: In that case I vote that we resolve #281 first, before adding more technical debt. 😅

RasmusOrsoe:
@asogaard closing because of messy commits. See #303 which is clean.

PR #299, opened 2022-09-23
Update GCD-file finding

asogaard:
* Update `is_gcd_file` to match more GCD files
* Don't case `filename` to lowercase in `is_i3_file`
* Allow for I3- and GCD-files with different suffixes/compression methods in the same folder in `find_i3_files`


RasmusOrsoe:
Have we checked that this works as intended? (doesn't break on stuff it worked on before?)

RasmusOrsoe (approved):

PR #300, opened 2022-09-23
Pointing task

MortenHolmRep:
Two new features
1) Task for symmetrical uncertainty estimation for both zenith and azimuth.
2) Loss function that calls the existing framework with the torch function for [Gaussian negative log likelihood loss](https://pytorch.org/docs/stable/generated/torch.nn.GaussianNLLLoss.html)

### DIFF
--- a/src/graphnet/components/loss_functions.py
+++ b/src/graphnet/components/loss_functions.py
@@ -60,6 +60,19 @@ class LossFunction(_WeightedLoss):
     def _forward(self, prediction: Tensor, target: Tensor) -> Tensor:
         """Syntax similar to `.forward` for implentation in inheriting classes."""
 
+class GaussLoss(LossFunction):
+    """Gaussian negative log likelihood loss."""
+
+    def _forward(self, prediction: Tensor, target: Tensor) -> Tensor:
+        """Implementation of loss calculation."""
+        # Check(s)
+        assert prediction.dim() == 2
+        assert prediction.size() == target.size()
+        variance = 1/prediction[:,2]
+
+        loss = torch.nn.GaussianNLLLoss()
+
+        return loss(prediction, target, variance, eps=1e-06, reduction='mean')
 
 class MSELoss(LossFunction):
     """Mean squared error loss."""
--- a/src/graphnet/components/loss_functions.py
+++ b/src/graphnet/components/loss_functions.py
@@ -60,17 +60,19 @@ class LossFunction(_WeightedLoss):
     def _forward(self, prediction: Tensor, target: Tensor) -> Tensor:
         """Syntax similar to `.forward` for implentation in inheriting classes."""
 
+
 class GaussLoss(LossFunction):
     """Gaussian negative log likelihood loss."""
 
     def _forward(self, prediction: Tensor, target: Tensor) -> Tensor:
         """Implementation of loss calculation."""
 
-        variance = 1/prediction[:,2]
+        variance = 1 / prediction[:, 2]
 
         loss = torch.nn.GaussianNLLLoss()
 
-        return loss(prediction, target, variance, eps=1e-06, reduction='mean')
+        return loss(prediction, target, variance, eps=1e-06, reduction="mean")
+
 
 class MSELoss(LossFunction):
     """Mean squared error loss."""
--- a/src/graphnet/components/loss_functions.py
+++ b/src/graphnet/components/loss_functions.py
@@ -65,9 +65,7 @@ class GaussLoss(LossFunction):
 
     def _forward(self, prediction: Tensor, target: Tensor) -> Tensor:
         """Implementation of loss calculation."""
-        # Check(s)
-        assert prediction.dim() == 2
-        assert prediction.size() == target.size()
+
         variance = 1/prediction[:,2]
 
         loss = torch.nn.GaussianNLLLoss()
--- a/src/graphnet/models/task/reconstruction.py
+++ b/src/graphnet/models/task/reconstruction.py
@@ -4,6 +4,25 @@ import torch
 from graphnet.models.task import Task
 from graphnet.utilities.maths import eps_like
 
+class ZenithAndAzimuthReconstructionWithKappa(Task):
+    """Reconstructs azimuthal and zenith angle and associated kappa (1/var)"""
+
+    # Requires three features: untransformed points in (x,y,z)-space.
+    nb_inputs = 3
+    #0 = x, 1 = y, 2 = z
+
+    def _forward(self, x):
+        # Transform outputs to angle and prepare prediction
+        kappa = torch.linalg.vector_norm(x, dim=1) + eps_like(x)
+        
+        zenith = torch.atan2(torch.linalg.vector_norm(x[:,:2]), x[:, 2])
+
+        azimuth = torch.atan2(x[:, 1], x[:, 0])
+        azimuth = torch.where(
+            azimuth < 0, azimuth + 2 * np.pi, azimuth
+        )  # atan(y,x) -> [-pi, pi]
+
+        return torch.stack((zenith,azimuth, kappa), dim=1)
 
 class AzimuthReconstructionWithKappa(Task):
     """Reconstructs azimuthal angle and associated kappa (1/var)."""
@@ -178,3 +197,4 @@ class InelasticityReconstruction(Task):
     def _forward(self, x):
         # Transform output to unit range
         return torch.sigmoid(x)
+
--- a/src/graphnet/models/task/reconstruction.py
+++ b/src/graphnet/models/task/reconstruction.py
@@ -4,25 +4,27 @@ import torch
 from graphnet.models.task import Task
 from graphnet.utilities.maths import eps_like
 
+
 class ZenithAndAzimuthReconstructionWithKappa(Task):
     """Reconstructs azimuthal and zenith angle and associated kappa (1/var)"""
 
     # Requires three features: untransformed points in (x,y,z)-space.
     nb_inputs = 3
-    #0 = x, 1 = y, 2 = z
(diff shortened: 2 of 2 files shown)


src/graphnet/components/loss_functions.py
```diff
@@ -61,6 +61,19 @@ def _forward(self, prediction: Tensor, target: Tensor) -> Tensor:
         """Syntax similar to `.forward` for implentation in inheriting classes."""
 
 
+class GaussLoss(LossFunction):
```
asogaard, line 64:
```suggestion
class GaussianNLLLoss(LossFunction):
```

src/graphnet/components/loss_functions.py
```diff
@@ -61,6 +61,19 @@ def _forward(self, prediction: Tensor, target: Tensor) -> Tensor:
...
+    """Gaussian negative log likelihood loss."""
+
+    def _forward(self, prediction: Tensor, target: Tensor) -> Tensor:
+        """Implementation of loss calculation."""
+
+        variance = 1 / prediction[:, 2]
+
+        loss = torch.nn.GaussianNLLLoss()
+
+        return loss(prediction, target, variance, eps=1e-06, reduction="mean")
```
asogaard, line 74: Making sure that we are returning the loss "element"/term for each example in the batch rather than averaging over them. (This is necessary for applying per-event weights.)
```suggestion
        loss = torch.nn.GaussianNLLLoss()

        variance = 1.0 / prediction[:, 2]
        elements = loss(prediction, target, variance, eps=1e-06, reduction="none")
        elements = torch.sum(elements, dim=-1)  # Sum along coordinate dimension
        return elements
```

src/graphnet/models/task/reconstruction.py
```diff
@@ -5,6 +5,27 @@
...
+        zenith = torch.atan2(torch.linalg.vector_norm(x[:, :2]), x[:, 2])
+
+        azimuth = torch.atan2(x[:, 1], x[:, 0])
+        azimuth = torch.where(
+            azimuth < 0, azimuth + 2 * np.pi, azimuth
+        )  # atan(y,x) -> [-pi, pi]
+
+        return torch.stack((zenith, azimuth, kappa), dim=1)
+
+
```
asogaard, line 28:
```suggestion
class PointingReconstructionWithKappa(Task):
    """Reconstructs azimuthal and zenith angle and associated kappa (1/var)"""

    # Requires three features: untransformed points in (x,y,z)-space.
    nb_inputs = 3

    def _forward(self, x):
        # Transform outputs to angle and prepare prediction
        kappa = torch.linalg.vector_norm(x, dim=1) + eps_like(x)

        zenith = torch.atan2(torch.linalg.vector_norm(x[:, :2]), x[:, 2])

        azimuth = torch.atan2(x[:, 1], x[:, 0])
        azimuth = torch.where(
            azimuth < 0, azimuth + 2 * np.pi, azimuth
        )  # atan(y,x) -> [-pi, pi]

        return torch.stack((zenith, azimuth, kappa), dim=1)


```

asogaard (changes requested):
Hi @MortenHolmRep,

I think this looks great! I have suggested one change regarding the "reduction" of the `GaussianNLLLoss` and on one the naming of the reconstruction task.

As this is a pretty important and common couple of classes, would you mind adding one unit test for each class?
* The `GaussianNLLLoss` unit test should take a handful of example predictions and targets, and you should check that the returned loss has the shape and values you expect. See https://github.com/graphnet-team/graphnet/blob/main/tests/ml/test_loss_functions.py#L41-L63 for how it has been done for the `LogCoshLoss`. You wouldn't be comparing two implementations, though, rather calculated vs. expected values.
* The `PointingReconstructionWithKappa` unit test should similarly take a handful of examples in the input `(x,y,z)` parametrisation and you should check that it gives the expected values for  zenith, azimuth, and kappa/variance. You could add the test here [ZenithAndAzimuthReconstructionWithKappa ](https://github.com/graphnet-team/graphnet/blob/main/tests/ml/test_task.py) and you would only need to check the `forward`/`_forward` method.

MortenHolmRep:
The requested changes have been made :)

> As this is a pretty important and common couple of classes, would you mind adding one unit test for each class?

I will make sure to create unit tests. I expect them to be done in the first half of week 40. I will report back with the results of the unit test and we can see if further adjustments are required. I am assigning myself to the PR until then.

MortenHolmRep:
@asogaard Before I go use too much time on any given thing, Is this the kind of [test](https://github.com/graphnet-team/graphnet/pull/300/commits/c9978cae75a2b33c2b4fb156e4c36fc50e8c21ab) you had in mind?

tests/ml/test_loss_functions.py
```diff
@@ -62,6 +67,30 @@ def test_log_cosh(dtype=torch.float32):
...
+def test_GaussianNLLLoss(dtype=torch.float32):
+    # Prepare test data, zenith, azimuth and kappa
+    prediction = torch.empty((7,3),dtype=dtype) # Shape [N, 3]
+    angles = torch.tensor([-pi, -pi/2, -pi/4, 0, pi/4, pi/2, pi])
+    kappa = abs(angles) + eps_like(angles)
+    prediction[:,0] = angles
+    prediction[:,1] = angles
+    prediction[:,2] = kappa
+
+    target = 0.0 * prediction.clone()  # Shape [N,3]
```
asogaard, line 79: Currently, this "only" tests cases where zenith == azimuth == kappa. That is probably fine, but I think it would be healthy to also test other combinations.
```suggestion
    # Prepare test data for prediction (zenith, azimuth and kappa) and target
    angles = torch.tensor([-pi, -pi/2, -pi/4, 0, pi/4, pi/2, pi])  # Shape: (7,)
    angles_matrix = angles.unsqueeze(1).tile(angles.size())  # (7,7)
    zenith_azimuth = torch.stack([
        angles_matrix.ravel(),
        angles_matrix.T.ravel(),
    ]).T  # (49,2)
    nb_examples = zenith_azimuth.size(0)
    kappa = torch.logspace(-2, 2, nb_examples)  # (49,)
    
    prediction = torch.hstack([zenith_azimuth, kappa.unsqueeze(1)])  # (49, 3)
    
    target = torch.zeros(nb_examples, 2)  # (zenith, azimuth)
    
```

asogaard (commented):
Hi @MortenHolmRep,
As requested, I have make some comments/suggestions to the unit test for `GaussianNLLLoss`. :)

asogaard:
> @asogaard Before I go use too much time on any given thing, Is this the kind of [test](https://github.com/graphnet-team/graphnet/pull/300/commits/c9978cae75a2b33c2b4fb156e4c36fc50e8c21ab) you had in mind?

Hi @MortenHolmRep,
Yes, I think this is exactly the type of test we want to do. I have made some suggestions to test a few more tests and to make the test data a little more broad. Let me know what you make of it. :)

asogaard:
Hi @MortenHolmRep,
I see that you are actively working on this. Please ping when you're ready for a second set of eyes. :)

MortenHolmRep:
With regards to testing the task; how would this work in practice?
if I am not mistaken, the other task that is being tested performs a loading of the task and a transform test.

Would I load test data, akin to the loss function test, into the task and run a small training pipeline to test the task properly?

tests/models/test_task.py
```diff
@@ -63,14 +65,17 @@ def test_PointingReconstructionWithKappa():
...
-    # Test task
-    # test_transform_prediction_and_target()
-    test_PointingReconstructionWithKappa()
+def is_function_local(object):
+    return isinstance(object, types.FunctionType) and object.__module__ == __name__
+
+
+def main():
+    # Test tasks in script
+    function_names = [name for name, _ in getmembers(sys.modules[__name__], predicate=is_function_local)]
```
MortenHolmRep, line 76: I have added a for loop that tests all (meaningful) functions in the script.
I thought it could help give helpful information when running the unit test script.
Thoughts?

asogaard: We're using `pytest` for our unit testing suite (see [here](https://github.com/graphnet-team/graphnet/blob/main/setup.py#L37) and [here](https://github.com/graphnet-team/graphnet/blob/main/.github/workflows/build-matrix.yml#L29-L33)), so you should be able to run all of the relevant functions (i.e., the ones named `test_*`) using:
```
$ pytest tests/models/test_task.py
```
I can update the [contributing guide](https://github.com/graphnet-team/graphnet/blob/main/CONTRIBUTING.md) to reflect this, if you think that would be helpful? :)

asogaard:
> With regards to testing the task; how would this work in practice? if I am not mistaken, the other task that is being tested performs a loading of the task and a transform test.
> 
> Would I load test data, akin to the loss function test, into the task and run a small training pipeline to test the task properly?

We haven't done a particularly good job of testing `Task`s in the past, you're right. The only other test we have is just testing different combinations of transforms. I think the best thing to do would be to create some toy data with appropriate dimensions and characteristics (like you do for the Gaussian NLL loss unit test), pass that through the `Task` and verify that the output looks as expected (e.g., the right dimensions, values within the expected bounds, and such).

tests/models/test_task.py
MortenHolmRep: I think it is important to be transparent about the functions used in the framework and that they live up to a predictable output. Moving this transparency more "in front" would solidify the validity of the analysis results.
Correct me if I am wrong but [this](https://codeclimate.com/github/graphnet-team/graphnet) shows how much of the code/functions are currently tested. if not we should have something that shows how many of the functions pass tests..
Taking it further, it could be a "check item" (I think?) that requires a function to be testable and have a test before it can be submitted as a valid PR. But that is more in your ballpark of doability.

PR #302, opened 2022-10-04
Restructure models

### LINKED ISSUE #294: Clean up dynedge.py
the file contains three versions of dynedge; `dynedge` (low energy / default), `dynege_V2` (Upgrade / High energy), `dynedge_V3` (pulsemap cleaning) each representing different requirements for the model in domain specific applications. 

The naming is confusing and it's not clear to the user which version to choose.

We should gather these three use cases into a single model (default to `dynedge`). We could introduce an argument `node_aggregations: List[str] = ['min', 'max', 'avg', 'sum']` where the user can specify which many-to-one projections are kept. This allows us to merge `dynedge` and `dynedge_V2`. Then we can come up with an elegant way of toggle on/off the "node aggregation" and "read-out" sections based on either user arguments or something else. Perhaps @asogaard has an elegant idea.

asogaard:
First attempt at rewriting DynEdge to better parametrise it and thus avoid duplication.

As I understand the current models:
* DynEdge_V2 vs DynEdge:
  * Add global variables before convolutions rather than after global pooling
  * Changing global variable n_pulses -> log10(n_pulses)
  * Computing, adding average features as global variable
  * Dropping sum-global pooling
* DynEdge_V3 vs DynEdge:
  * No global pooling, i.e. for node-level predictions

How to reproduce the existing models (`DynEdgeNew` == proposed class)
```
nb_inputs = 8

# DynEdge_V1
dynedge = DynEdge(nb_inputs)
dynedge_new = DynEdgeNew(
    nb_inputs,
    add_global_variables_after_pooling=True,
    global_pooling_schemes=["min", "max", "mean", "sum"],
)
# NB: Not exact:
# - Global pooling now happens *after* a LeakyReLU, rather than beofre
# - Uses log10(n_pulses) rather than n_pulses as global variable
# - Adds average features as global variables

# DynEdge_V2
dynedge = DynEdge_V2(nb_inputs)
dynedge_new = DynEdgeNew(
    nb_inputs,
    global_pooling_schemes=["min", "max", "mean"],
)

# DynEdge_V3
dynedge = DynEdge_V3(nb_inputs)
dynedge_new = DynEdgeNew(nb_inputs)
```
Closes #294


RasmusOrsoe:
@asogaard I only see DynEdgeJinst in these two files?

asogaard:
> @asogaard I only see DynEdgeJinst in these two files?

Which two files do you mean? The examples use the updated DynEdge while DynEdgeJINST is mainly kept for reference.

RasmusOrsoe (changes requested):
Hey @asogaard. I think this is great. All I think is missing is a good docstring for the new DynEdge that explains the new inputs!

asogaard:
> Hey @asogaard. I think this is great. All I think is missing is a good docstring for the new DynEdge that explains the new inputs!

A taste of my own medicine, I see... 😅  But excellent suggestion, and that should be done now!

RasmusOrsoe (approved):

PR #303, opened 2022-10-04
Direction example update -v2

RasmusOrsoe:
I messed up the commits on #297 . This is a clean PR.

### DIFF
--- a/examples/train_direction_reco.py
+++ b/examples/train_direction_reco.py
@@ -24,6 +24,7 @@ from graphnet.models.training.callbacks import ProgressBar, PiecewiseLinearLR
 from graphnet.models.training.utils import (
     get_predictions,
     make_train_validation_dataloader,
+    save_results,
 )
 from graphnet.utilities.logging import get_logger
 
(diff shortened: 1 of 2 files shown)


src/graphnet/models/gnn/dynedge.py
```diff
@@ -170,7 +170,12 @@ def forward(self, data: Data) -> Tensor:
 
 
 class DynEdge_V2(GNN):
-    def __init__(self, nb_inputs, layer_size_scale=4):
```
asogaard, line 173: The `Model` class already has the option to do coarsening, and the implementation in `DynEdge` and this proposed change to `DynEdge_V2` does the coarsening before any actual convolutional operations, so I think we could leave this functionality in the `Model` class. (That's at least what I am proposing in #302.) Of course, if we had a `GNN` model that relied on coarsening "in the middel" of the convolutional operations, that would be a different story, but that's not the case here, as far as I can tell.

examples/train_direction_reco.py
```diff
@@ -76,14 +79,18 @@ def train(config):
...
-    else:
-        gnn = DynEdge(
-            nb_inputs=detector.nb_outputs,
+    if config["node_pooling"] == "dom_coarsening":
+        gnn = DynEdge_V2(
+            nb_inputs=detector.nb_outputs, node_pooling=DOMCoarsening()
         )
+    else:
+        gnn = DynEdge_V2(nb_inputs=detector.nb_outputs)
+
```
asogaard, line 93: See comment below about leaving the coarsening to `Model`.

asogaard (commented):
Hi @RasmusOrsoe,
I have made one comment/suggestion about where we implement coarsening in the models. Let me know what you think. :)

PR #306, opened 2022-10-11
Changed i3featureextractor.py to also extract the event datetime from frame I3EventHeader.

Peterandresen12:
I have changed the i3featureextractor.py to also extract the datetime of the events, which can for instance be used to identify the moon position at the time of each event. The datetime is required for the `icecube.astro.I3GetMoonDirection()`

https://docs.icecube.aq/icetray/main/projects/astro/index.html

### DIFF
--- a/src/graphnet/data/extractors/i3featureextractor.py
+++ b/src/graphnet/data/extractors/i3featureextractor.py
@@ -57,14 +57,13 @@ class I3FeatureExtractorIceCube86(I3FeatureExtractor):
         if "CalibrationErrata" in frame:
             calibration_errata = frame.Get("CalibrationErrata")
 
+        event_time = frame["I3EventHeader"].start_time.mod_julian_day_double
+
         for om_key in om_keys:
             # Common values for each OM
             x = self._gcd_dict[om_key].position.x
             y = self._gcd_dict[om_key].position.y
             z = self._gcd_dict[om_key].position.z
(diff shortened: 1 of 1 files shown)


src/graphnet/data/extractors/i3featureextractor.py
```diff
@@ -61,6 +62,7 @@ def __call__(self, frame) -> dict:
             x = self._gcd_dict[om_key].position.x
             y = self._gcd_dict[om_key].position.y
             z = self._gcd_dict[om_key].position.z
+            event_time = frame["I3EventHeader"].start_time.mod_julian_day_double
```
asogaard, line 65: I would add this above the `for om_key ...` loop, as the value is the same for all OMs in the event.

asogaard (changes requested):
Hi @Peterandresen12,

Thanks for this PR! 🚀   While I think it is a little inefficient for us to add the event time to all pulses, rather than just once per event, I realise that this is the best/only way to do it currently, so I'm not going to push back on that. I have added on suggestion but otherwise I think this is fine to merge. :)

src/graphnet/data/extractors/i3featureextractor.py
Peterandresen12: Great suggestion. I have changed it.

asogaard (approved):

PR #308, opened 2022-10-11
3d von mises fisher [draft]

RasmusOrsoe:
This PR contains:

1. DirectionReconstructionWithKappa Task
2. 3DVonMisesFisherLoss
3. General Label-class
    a Direction label class for use with the new dataset syntax. We could pass a list of Label-classes and use Label.key to simplify the arguments for the new dataset syntax (right now asks for a dict{key:function} ) 

Let me know what you think @asogaard. Would have liked to tag this as a "draft" given the new Label-class stuff, but I can't figure out how.

Rasmus

### DIFF
--- a/src/graphnet/models/task/reconstruction.py
+++ b/src/graphnet/models/task/reconstruction.py
@@ -39,6 +39,22 @@ class AzimuthReconstruction(AzimuthReconstructionWithKappa):
         return angle
 
 
+class DirectionReconstructionWithKappa(Task):
+    """Reconstructs Direction with kappa from The 3DVonMisesFisher Distribution."""
+
+    # Requires 4 features: untransformed points in (x,y,z)-space + kappa estimation.
+    nb_inputs = 4
+
+    def _forward(self, x):
+        # Transform outputs to angle and prepare prediction
+        norm = torch.sqrt(x[:, 0] ** 2 + x[:, 1] ** 2 + x[:, 2] ** 2)
+        kappa = torch.abs(x[:, 3]) + eps_like(x)
+        vec_x = x[:, 0] / norm
+        vec_y = x[:, 1] / norm
+        vec_z = x[:, 2] / norm
+        return torch.stack((vec_x, vec_y, vec_z, kappa), dim=1)
+
+
 class PassOutput1(Task):
     """Passes 1 output without interference."""
 
--- a/src/graphnet/models/task/reconstruction.py
+++ b/src/graphnet/models/task/reconstruction.py
@@ -42,16 +42,15 @@ class AzimuthReconstruction(AzimuthReconstructionWithKappa):
 class DirectionReconstructionWithKappa(Task):
     """Reconstructs Direction with kappa from The 3DVonMisesFisher Distribution."""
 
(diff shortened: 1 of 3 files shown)


src/graphnet/models/task/reconstruction.py
```diff
@@ -39,6 +39,22 @@ def _forward(self, x):
...
+    nb_inputs = 4
+
+    def _forward(self, x):
+        # Transform outputs to angle and prepare prediction
+        norm = torch.sqrt(x[:, 0] ** 2 + x[:, 1] ** 2 + x[:, 2] ** 2)
+        kappa = torch.abs(x[:, 3]) + eps_like(x)
+        vec_x = x[:, 0] / norm
+        vec_y = x[:, 1] / norm
+        vec_z = x[:, 2] / norm
+        return torch.stack((vec_x, vec_y, vec_z, kappa), dim=1)
```
asogaard, line 55: In the proposed implementation, the normalisation of the `(x,y,z)`-vector is arbitrary, meaning that the task is "overparametrised" (i.e., there is an output, `x[:,3]`, that the model can freely change with no impact on the actual prediction). Would it make sense to take a note from the `AzimuthReconstructionWithKappa` task and instead take the norm of the `(x,y,z)`-vector to represent kappa, like they do in the original paper, I believe? Since kappa is then a norm of a vector, it is automatically non-negative, meaning that we don't need to introduce the discontinuous `torch.abs(...)`.
```suggestion
    # Requires three features: untransformed points in (x,y,z)-space.
    nb_inputs = 3

    def _forward(self, x):
        # Transform outputs to angle and prepare prediction
        kappa = torch.linalg.vector_norm(x, dim=1) + eps_like(x)
        vec_x = x[:, 0] / kappa
        vec_y = x[:, 1] / kappa
        vec_z = x[:, 2] / kappa
        return torch.stack((vec_x, vec_y, vec_z, kappa), dim=1)
```

src/graphnet/training/labels.py
```diff
@@ -0,0 +1,40 @@
+from abc import ABC, abstractmethod
+from typing import Any, Callable, Dict, List, Optional, Tuple, Union
+import numpy as np
+import torch
+from torch_geometric.data import Data
+from graphnet.utilities.logging import LoggerMixin
+
+
+class Label(ABC, LoggerMixin):
```
asogaard, line 9: I really like this, @RasmusOrsoe; I think it's quite elegant. We will have to consider whether we want _primarily_ to add these labels as part of the database production (e.g., trough similarly small `I3TruthExtractor` subclasses, like `I3TruthDirectionExtractor`), or primarily add them on-the-fly during training. But either way, I think it's nice to have both options, so I'd like to see this included. :)

asogaard (approved):
Hi @RasmusOrsoe,
I really like these additions! I have added one suggestion regarding the parametrisation of  `DirectionReconstructionWithKappa` that you can do with what you want. Otherwise I really like the proposed `Label`, `Direction`, and `VonMisesFisher3DLoss` classes! :)

src/graphnet/models/task/reconstruction.py
RasmusOrsoe: @asogaard Could we perhaps zoom about this?

Issue #226, opened 2022-06-28
Add model and data config files


asogaard:
By setting up a standardised syntax for model and data config files (e.g. using pydantic), building off what little is currently in https://github.com/icecube/graphnet/tree/main/datasets makes it easier to reproduce results, make their provenance transparent, and share configurations between users.

- [ ] Data/DataConverter config
- [X] Dataset config (#353)
- [X] Model config (#309)

PR #309, opened 2022-10-12
Implementing model configuration and adding abstract base Model class

asogaard:
* Adds `ModelConfig` class for configuring models directly from file, and similarly for documenting model configurations as part of experimentation and hyperparameter optimisation
* Renames `Model` -> `StandardModel` to signify that, which the way we typically construct models (`model = StandardModel(detector=..., gnn=..., tasks=[...])`) is the standard approach, it is possibly to construct GraphNeT models entirely freely, like any other pytorch model
* Introduces a base `Model` class that provides some configuration-related interface, `LoggerMixin` for all inheriting classes, etc.
* Changes the `LoggerMixin` class such that inheriting classes can use the `.{critical,error,warning,info,debug}()` methods directly, i.e., `self.debug(msg)` rather than having to do, e.g., `self.logger.info(msg)`. This was done to avoid a conflict with the `.logger` attribute also being set/overwritten by pytorch-lightning. I also think it's a bit cleaner and more intuitive.

Addresses first part of #226 — `DataConfig` still not implemented


src/graphnet/models/task/task.py
```diff
@@ -55,11 +60,12 @@ class Task(LightningModule):
     def nb_inputs(self) -> int:
         """Number of inputs assumed by task."""
 
+    @save_config
     def __init__(
         self,
         hidden_size: int,
         target_labels: Union[str, List[str]],
-        loss_function: LossFunction,
+        loss_function: "LossFunction",
```
RasmusOrsoe, line 68: Is this intentional?

tests/models/test_config.py
```diff
@@ -0,0 +1,144 @@
...
+    assert isinstance(loaded_config, ModelConfig)
+    assert loaded_config == model.config
+
+    # Construct model
+    constructed_model = Model.from_config(loaded_config)
+    assert constructed_model.config == model.config
+    assert repr(constructed_model) == repr(model)
+
+
+def test_complete_config(path: str = "/tmp/complete.yml"):
```
RasmusOrsoe, line 65: Thanks for adding this!

src/graphnet/training/loss_functions.py
```diff
@@ -13,17 +15,19 @@ def final(f):  # Identity decorator
...
 import torch
 from torch import Tensor
-from torch.nn.modules.loss import _WeightedLoss
+
+from graphnet.models.config import save_config
+from graphnet.models.model import Model
 
 
-class LossFunction(_WeightedLoss):
+class LossFunction(Model):
```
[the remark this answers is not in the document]
asogaard: Yes, I think you're right!

src/graphnet/models/task/task.py
asogaard: Yes, this is to avoid a cyclic dependency caused by exposing `StandardModel` directly from `graphnet.models`. It has no practical effect, but is just a way to add type hinting for classes that are not directly imported.

RasmusOrsoe (approved):

PR #310, opened 2022-10-14
Improve code quality and checking

### LINKED ISSUE #109: Add docstrings
We should aim to document all non-trivial methods/functions and modules, since these are used to automatically generate the API documentation at [https://icecube.github.io/graphnet](https://icecube.github.io/graphnet). While this is not something that need to be done in one go, we should aim to have all new PRs include sufficient documentation, and to add documentation when reworking parts of the code where it's missing.

### LINKED ISSUE #225: Add docstring checking to pre-commit hooks
This ensures that all functions, modules, etc. that are added have docstrings with a standard convention (https://peps.python.org/pep-0257/)

### LINKED ISSUE #227: Enforce all pre-commit hooks in CI/CD pipeline
Currently, only black is checked during automated jobs in the CI/CD pipeline. Flake isn't checked, and if we're adding additional quality requirements (see #225, #229) we should also make sure that these requirements are enforced before merging code to main.

### LINKED ISSUE #229: Add type-hint requirements and/or checking to pre-commit
This ensures that all functions have type hints, and that these are mutually compatible (https://peps.python.org/pep-0585/, https://mypy.readthedocs.io/en/stable/)

asogaard:
* Adds automatic formating and checking of docstrings
* Adds automatic checking of static typing
* Updates all files to be compliant with these
* Changes call to `_make_df` to `_convert_to_dataframe` in `ParquetToSQLiteConverter`, as the former was renamed but the corresponding call was not update to reflect this.
* Adds new set of `Code quality` GitHub actions for checking each of the pre-commit hooks, replacing `black` which was the only thing we checked for before.
* Renames and regroups GitHub Actions to, e.g. ,`Build`, for clarity.
* Renames example script to a similar, imperative form.
* Added `configupdater` as setup dependency, as it is used in `graphnet.pisa`.

After this PR the recommended way to document functions is like:
```python
def my_function(arg: int, other: str = "a") -> None:
    """Do something.

    A more detailed explanation of the logic behind `my_function` goes here.

    Args:
        arg: This thing.
        other: This other thing.
    """
```
That is, types and default values are document in the code itself. Any explanation of the logic is done in the docstring. For trivial function, the `Args` and `Returns` section on the docstring can be omitted. This should reduce duplication of e.g. type information, and the typing hinting allows for static type checking using `mypy`.

Closes #109 
Closes #225 
Closes #227 
Closes #229


src/graphnet/data/extractors/i3featureextractor.py
```diff
@@ -122,14 +149,24 @@ def _get_relative_dom_efficiency(self, frame, om_key):
 
 
 class I3FeatureExtractorIceCubeDeepCore(I3FeatureExtractorIceCube86):
-    """..."""
+    """Class for extracting reconstructed features for IceCube-DeepCore."""
```
RasmusOrsoe, line 152: I'm not sure an IceCuber will recognize what is meant with "reconstructed features" here. I think "Class for extracting unfolded waveform pulses and associated OM information, such as position, orientation, etc." is more descriptive. 

I don't think you necessarily have to change all the doc strings accordingly in this PR. Perhaps it's best if we discuss language first and iterate on doc string wording in a separate PR in the future.

RasmusOrsoe (approved):
@asogaard really nice! I've added a single comment on the feature extractors; but as I mention, I think we can save that for later, if you prefer. 

On another note; how does the automatic checks on doc strings and type hints affect our future PR's? Will the checks update the doc strings and type hints like black formatting does? Or will we have to consult output logs?

asogaard:
@RasmusOrsoe, if we do implement these checks as part of the automated PR review (which I think we should but haven't done yet, then likely there will be a bit of automatic reformatting, but mostly it will throw errors that the developer would need to implement. I could look into automatically posting comments to the PR with the relevant parts of the error log so as to save some digging through log files manually.

src/graphnet/data/extractors/i3featureextractor.py
asogaard: I think this is a good point, and this PR might be a good time to review our language as it will update all docstrings in the codebase. I'll try to keep this in mind going forward!

asogaard:
_... It's done..._ 😮‍💨

RasmusOrsoe (approved):
Well done!!

asogaard:
As with all PRs of this size, there is a small change of a bug sneaking in, but as the vast majority of changes are purely to the "frosting", I figure it is still okay to merge.

Please see the original, updated comments for notes on the major changes. Please also see the slightly updated [contributing guidelines](https://github.com/graphnet-team/graphnet/blob/main/CONTRIBUTING.md). Finally, people who have been holding off on using pre-commit hooks ought to set it up now. I will also talk about this at next week's meeting.

PR #311, opened 2022-10-14
Bug-fix for dynedge Refactor

RasmusOrsoe:
#302 introduced a nice refactor of dynedge. 

However, a small bug was introduced where the concatenation of global variables failed because of dimensions. This is fixed here.


asogaard (approved):
👏

Issue #111, opened 2021-12-14
Implement multi-class classification


asogaard:
Currently, the repo support binary graph-level classification. However, for tasks like per-flavour neutrino classification (challenging as it might be), graph-level multi-class would be relevant.

PR #329, opened 2022-10-27
Multi-class classification implementation

MortenHolmRep:
Implementation of multi-class classification predictions as per #111 , where the new operational structure would imply 
`Raw Data -> [noise, muon, neutrino]`

In the current implementation, this could also easily be expanded to 
`Raw Data -> [noise, muon, nu_e, nu_mu, nu_tau]`
But a two-step classification, first on `[noise, muon, neutrino]` and then a second classification, with a neutrino-specific task, training with the identified neutrinos on `[nu_e, nu_mu, nu_tau]` is probably preferred for neutrino classifications.
comments on an approach or alteration to this multi-class classification implementation are welcome.

Training with the task using `examples/train_model.py` works with the following alterations, where the target are `pid`:
```
task = MulticlassificationTask(
        hidden_size=gnn.nb_outputs,
        target_labels=config["target"],
        loss_function=MultiClassificationCrossEntropyLoss(),
    )
```

and for the results
```
results = get_predictions(
        trainer,
        model,
        validation_dataloader,
        [config["target"] + "_noise_pred", config["target"] + "_muon_pred", config["target"]+ "_neutrino_pred"],
        additional_attributes=[config["target"], "event_no"],
    )
```

Implementation by @Peterandresen12 and I


Peterandresen12 (approved):
Looks good to me :-)

src/graphnet/models/task/reconstruction.py
```diff
@@ -167,6 +167,13 @@ def _forward(self, x):
         # Leave as it is
         return x
 
+class MulticlassificationTask(Task):
+    # requires three features: probability of being noise, muon or neutrino.
+    nb_inputs = 3
+
+    def _forward(self, x):
+        # transform probability of being noise, muon or neutrino
+        return torch.sigmoid(x)
```
asogaard, line 176: This hard-codes the `Task` to three-way classification, which does not align with the class name. I think we need to do something a bit more dynamical, see (untested) sketched suggestion below. You also need to use softmax activation for multi-class classification to ensure that outputs are in [0,1] and sum to one. (The last condition is not satisfied by sigmoid.)
```suggestion
class MulticlassClassificationTask(Task):
    # Requires the same number of features as the number of classes being predicted
    @property
    def nb_inputs(self):
        return self._nb_inputs

    @save_config
    def __init__(self, nb_inputs, *args, **kwargs):
        self._nb_inputs = nb_inputs
        self._softmax = torch.nn.Softmax()
        super().__init__(self, *args, **kwargs)

    def _forward(self, x):
        # Transform latent features into probabilities.
        return self._softmax(x)
```

src/graphnet/training/loss_functions.py
```diff
@@ -111,11 +111,27 @@ def _forward(self, prediction: Tensor, target: Tensor) -> Tensor:
...
+
+class MultiClassificationCrossEntropyLoss(LossFunction):
+    """Computes cross entropy for a matrix [num_class,N] of predictions (between 0 and 1),
+    targets [1,N] should be a particle id for noise (1,-1) muon (13,-13) and neutrino (12,-12,14,-14,16,-16) respectively.
+    loss should be reported elementwise, so set reduction to none.
+    """
+
+    def _forward(self, prediction: Tensor, target: Tensor) -> Tensor:
+        # transformer for multiple pid types to total of three classes
+        device = torch.device('cuda' if torch.cuda.is_available() else 'cpu')
```
asogaard, line 123:
```suggestion
        device = prediction.device
```

src/graphnet/training/loss_functions.py
```diff
@@ -111,11 +111,27 @@ def _forward(self, prediction: Tensor, target: Tensor) -> Tensor:
         return elements
 
 
+
+class MultiClassificationCrossEntropyLoss(LossFunction):
```
asogaard, line 115:
```suggestion
class CrossEntropyLoss(LossFunction):
```

src/graphnet/training/loss_functions.py
```diff
@@ -111,11 +111,27 @@ def _forward(self, prediction: Tensor, target: Tensor) -> Tensor:
...
+    """Computes cross entropy for a matrix [num_class,N] of predictions (between 0 and 1),
+    targets [1,N] should be a particle id for noise (1,-1) muon (13,-13) and neutrino (12,-12,14,-14,16,-16) respectively.
+    loss should be reported elementwise, so set reduction to none.
+    """
```
asogaard, line 119: We shouldn't hard-code specific PIDs into a generic cross-entropy loss function.
```suggestion
    """Compute cross entropy loss.
    
    Predictions are an [N, num_class]-matrix  of values between 0 and 1, and
    targets are an [N,1]-matrix with integer values in (0, num_classes - 1).
    """
```

src/graphnet/training/loss_functions.py
```diff
@@ -111,11 +111,27 @@ def _forward(self, prediction: Tensor, target: Tensor) -> Tensor:
...
+
+    def _forward(self, prediction: Tensor, target: Tensor) -> Tensor:
+        # transformer for multiple pid types to total of three classes
+        device = torch.device('cuda' if torch.cuda.is_available() else 'cpu')
+        pid_transform = {1:0, 12:2, 13:1, 14:2, 16:2}
+        target_new = one_hot(torch.tensor([pid_transform[np.abs(int(value))] for value in target]), 3).to(device)
```
asogaard, line 125: As mentioned above, we need to find a way to make the loss function more generic, i.e. not hard-coding specific PID values. I suggest we implement the `__init__()` function to take an argument like `class_options: List[int]`; alternatively `num_classes: int` and then require the target attribute to be in [0, num_classes[. (I probably like the second option better.) And then dynamically implement a one-hot encoding that allows for an arbitrary number of classes.

src/graphnet/training/loss_functions.py
```diff
@@ -111,11 +111,27 @@ def _forward(self, prediction: Tensor, target: Tensor) -> Tensor:
...
+        return cross_entropy(
+            prediction.float(), target_new.float(), reduction="none"
+        )
+
 class BinaryCrossEntropyLoss(LossFunction):
     """Computes binary cross entropy for a vector of predictions (between 0 and 1),
     targets should be 0 and 1 for muon and neutrino respectively
     where prediction is prob. the PID is neutrino (12,14,16)
-    loss should be reported elementwise, so set reduction to None
+    loss should be reported elementwise, so set reduction to none.
```
asogaard, line 134:
```suggestion
```

src/graphnet/training/loss_functions.py
```diff
@@ -111,11 +111,27 @@ def _forward(self, prediction: Tensor, target: Tensor) -> Tensor:
...
+        pid_transform = {1:0, 12:2, 13:1, 14:2, 16:2}
+        target_new = one_hot(torch.tensor([pid_transform[np.abs(int(value))] for value in target]), 3).to(device)
+        return cross_entropy(
+            prediction.float(), target_new.float(), reduction="none"
+        )
+
 class BinaryCrossEntropyLoss(LossFunction):
     """Computes binary cross entropy for a vector of predictions (between 0 and 1),
     targets should be 0 and 1 for muon and neutrino respectively
     where prediction is prob. the PID is neutrino (12,14,16)
```
asogaard, line 133:
```suggestion
    """Compute binary cross entropy loss.
    
    Predictions are vector probabilities (i.e., values between 0 and 1), and 
    targets should be 0 and 1.
```

asogaard (changes requested):
Hi @MortenHolmRep and @Peterandresen12,

Thanks for this PR! It would make a nice addition to the repo. 🙂 

I have added a few suggestions and comments that center on the fact that, currently, the proposed task and loss function hard-code assumption about the number of classes, and the PIDs of the particles being classified. We should try to be more general if we want to implement generic multi-class classification.

MortenHolmRep:
> Hi @MortenHolmRep and @Peterandresen12,
> 
> Thanks for this PR! It would make a nice addition to the repo. 🙂 
> 
> I have added a few suggestions and comments that center on the fact that, currently, the proposed task and loss function hard-code assumption about the number of classes, and the PIDs of the particles being classified. We should try to be more general if we want to implement generic multi-class classification.

I will look into a rework tonight, based on the suggestions 👍

src/graphnet/models/task/reconstruction.py
MortenHolmRep: This is true. The predictions that I ran gave results that did not add up to 1, i.e. 
predictions of {noise, muon, neutrino} = {7.072934220832261e-16, 5.221908991911821e-14, 1.0}

src/graphnet/training/loss_functions.py
MortenHolmRep: upon testing, `one_hot()` automatically identifies the number of classes based on the range of the input.
i.e. a list containing the elements of [0,1,2] becomes 3 classes, similarly [0,1,2,3,4] becomes 5 classes.

as long as the target is transformed prior. Should this be implemented as a custom transformer that for example takes in the dict in line 124?

src/graphnet/models/task/reconstruction.py
MortenHolmRep: can you elaborate on the `@save_config` decorator?
and would the implementation of the `__init__()` imply that the task would have two additional callables? i.e.
```
task = MulticlassClassificationTask(
        nb_inputs=?,
        softmax= <default torch.nn.Softmax(),
        hidden_size=gnn.nb_outputs,
        target_labels=config["target"],
        loss_function=CrossEntropyLoss(),
    )
```

MortenHolmRep: When Softmax() is called in:
```
def __init__(self, nb_inputs, *args, **kwargs):
    self._nb_inputs = nb_inputs
    self._softmax = torch.nn.Softmax()
    super().__init__(self, *args, **kwargs)
```
The following error occour
```
Exception has occurred: AttributeError
cannot assign module before Module.__init__() call
  File "/lustre/hpc/icecube/qgf305/workspace/graphnet/src/graphnet/models/task/reconstruction.py", line 177, in __init__
    self._softmax = torch.nn.Softmax()
  File "/lustre/hpc/icecube/qgf305/workspace/graphnet/src/graphnet/models/config.py", line 242, in wrapper
    ret = init_fn(self, *args, **kwargs)
  File "/groups/icecube/qgf305/workspace/analyses/multi_classification_on_stop_and_track_muons/modelling/train_multiclass_classification_model.py", line 197, in main
    task = MulticlassClassificationTask(
  File "/groups/icecube/qgf305/workspace/analyses/multi_classification_on_stop_and_track_muons/modelling/train_multiclass_classification_model.py", line 261, in <module>
    main()
```
[Consulting stack](https://stackoverflow.com/a/43080779), it should be a matter of calling the super().__init__() correctly, but I haven't been able to construct it in such a way that would fix the issue.

asogaard: Try moving the `super().__init__(...)` call up as the first statement in the constructor. 🙂

MortenHolmRep: > Try moving the `super().__init__(...)` call up as the first statement in the constructor. 🙂

I tried to reorder it, but then It complains about `hidden_size`
`Argument of type "Self@MulticlassClassificationTask" cannot be assigned to parameter "hidden_size" of type "int" in function "__init__"`
It gives approximately the same error in the traceback when I try to run it.
`hidden_size` can also not be assigned as a `self.` statement in the init

asogaard: > This is true. The predictions that I ran gave results that did not add up to 1, i.e. predictions of {noise, muon, neutrino} = {7.072934220832261e-16, 5.221908991911821e-14, 1.0}

I think this would still qualify as summing to 1, since the first two numbers are probably below `eps` for the type of float you're using. But the softmax point should still apply. :)

asogaard: > > Try moving the `super().__init__(...)` call up as the first statement in the constructor. 🙂
> 
> I tried to reorder it, but then It complains about `hidden_size` `Argument of type "Self@MulticlassClassificationTask" cannot be assigned to parameter "hidden_size" of type "int" in function "__init__"` It gives approximately the same error in the traceback when I try to run it. `hidden_size` can also not be assigned as a `self.` statement in the init

Try 
```
def __init__(self, nb_inputs, *args, **kwargs):
    self._nb_inputs = nb_inputs
    super().__init__(self, *args, **kwargs)
    self._softmax = torch.nn.Softmax()
```
Otherwise, could you paste, or point to, the current version of the relevant code snippet along with the full error message?

src/graphnet/training/loss_functions.py
asogaard: * Since we're using batching, and since each batch may not contain examples of all classes (unlikely, but not impossible), I think we need to specify `one_hot(target, num_classes=num_classes)`. (I also don't think you need to set device explicitly?)
* I think we should allow for either:
    * `target` is an integer in `[0, num_classes[` and thus one_hot can be used directly.
    * `target` is not that, and an optional list of classes will be provide to `MulticlassClassificationTask(int: nb_classes, Optional[List[Any]]: classes = None)`, and then in the `CrossEntropyLoss` class, do something like:

```
        if self._classes is not None:
            assert isinstance(self._classes, Tensor)  # Cast to tensor in the constructor
            target_integer = torch.nonzero(target == self._classes.unsqueeze(dim=1), as_tuple=True)[0]
        else:
            assert target.dtype in [torch.int32, torch.int64]
            assert torch.all(target >= 0)
            assert torch.all(target < self._nb_classes)
            target_integer = target

        target_one_hot = one_hot(target_integer)
        return cross_entropy(
            prediction.float(), target_one_hot.float(), reduction="none"
        ) 
```

asogaard:
Hi @MortenHolmRep,

Thanks for the quick iteration! :)  I have added comments to a few threads. Please let me know if there's anything else you'd like to discuss there. Otherwise, feel free to re-request a review once you think the code in the PR is ready for a second look. :)

src/graphnet/models/task/reconstruction.py
MortenHolmRep: I am somewhat comforted in you trying all the same things I've already tried :D
```
def __init__(self, nb_inputs, *args, **kwargs):
    super().__init__(self, *args, **kwargs)
    self._nb_inputs = nb_inputs
    self._softmax = softmax()
```
The above code has the following trace:
```
Exception has occurred: TypeError
__init__() got multiple values for argument 'hidden_size'
  File "/lustre/hpc/icecube/qgf305/workspace/graphnet/src/graphnet/models/config.py", line 242, in wrapper
    ret = init_fn(self, *args, **kwargs)
  File "/lustre/hpc/icecube/qgf305/workspace/graphnet/src/graphnet/models/task/reconstruction.py", line 178, in __init__
    super().__init__(self, *args, **kwargs)
  File "/lustre/hpc/icecube/qgf305/workspace/graphnet/src/graphnet/models/config.py", line 242, in wrapper
    ret = init_fn(self, *args, **kwargs)
  File "/groups/icecube/qgf305/workspace/analyses/multi_classification_on_stop_and_track_muons/modelling/train_multiclass_classification_model.py", line 206, in main
    task = MulticlassClassificationTask(
  File "/groups/icecube/qgf305/workspace/analyses/multi_classification_on_stop_and_track_muons/modelling/train_multiclass_classification_model.py", line 270, in <module>
    main()
```
```
def __init__(self, nb_inputs, *args, **kwargs):
    super(Task, self).__init__(self, *args, **kwargs)
    self._nb_inputs = nb_inputs
    self._softmax = softmax()
```
The above code initially removes the hidden_size message in pylance but results in the following trace
```
Exception has occurred: TypeError
__init__() got an unexpected keyword argument 'hidden_size'
  File "/lustre/hpc/icecube/qgf305/workspace/graphnet/src/graphnet/models/task/reconstruction.py", line 178, in __init__
    super(Task,self).__init__(self, *args, **kwargs)
  File "/lustre/hpc/icecube/qgf305/workspace/graphnet/src/graphnet/models/config.py", line 242, in wrapper
    ret = init_fn(self, *args, **kwargs)
  File "/groups/icecube/qgf305/workspace/analyses/multi_classification_on_stop_and_track_muons/modelling/train_multiclass_classification_model.py", line 206, in main
    task = MulticlassClassificationTask(
  File "/groups/icecube/qgf305/workspace/analyses/multi_classification_on_stop_and_track_muons/modelling/train_multiclass_classification_model.py", line 270, in <module>
    main()
```
```
def __init__(self, nb_inputs, *args, **kwargs):
    self._nb_inputs = nb_inputs
    super().__init__(self, *args, **kwargs)
    self._softmax = softmax()
```
The above yields the following trace
```
Exception has occurred: TypeError
__init__() got multiple values for argument 'hidden_size'
  File "/lustre/hpc/icecube/qgf305/workspace/graphnet/src/graphnet/models/config.py", line 242, in wrapper
    ret = init_fn(self, *args, **kwargs)
  File "/lustre/hpc/icecube/qgf305/workspace/graphnet/src/graphnet/models/task/reconstruction.py", line 179, in __init__
    super().__init__(self, *args, **kwargs)
  File "/lustre/hpc/icecube/qgf305/workspace/graphnet/src/graphnet/models/config.py", line 242, in wrapper
    ret = init_fn(self, *args, **kwargs)
  File "/groups/icecube/qgf305/workspace/analyses/multi_classification_on_stop_and_track_muons/modelling/train_multiclass_classification_model.py", line 206, in main
    task = MulticlassClassificationTask(
  File "/groups/icecube/qgf305/workspace/analyses/multi_classification_on_stop_and_track_muons/modelling/train_multiclass_classification_model.py", line 270, in <module>
    main()
```

Not that I expected a different outcome but I tested it.
```
def __init__(self, nb_inputs, *args, **kwargs):
    self._nb_inputs = nb_inputs
    super(Task, self).__init__(self, *args, **kwargs)
    self._softmax = softmax()
```
```
Exception has occurred: TypeError
__init__() got an unexpected keyword argument 'hidden_size'
  File "/lustre/hpc/icecube/qgf305/workspace/graphnet/src/graphnet/models/task/reconstruction.py", line 179, in __init__
    super(Task,self).__init__(self, *args, **kwargs)
  File "/lustre/hpc/icecube/qgf305/workspace/graphnet/src/graphnet/models/config.py", line 242, in wrapper
    ret = init_fn(self, *args, **kwargs)
  File "/groups/icecube/qgf305/workspace/analyses/multi_classification_on_stop_and_track_muons/modelling/train_multiclass_classification_model.py", line 206, in main
    task = MulticlassClassificationTask(
  File "/groups/icecube/qgf305/workspace/analyses/multi_classification_on_stop_and_track_muons/modelling/train_multiclass_classification_model.py", line 270, in <module>
    main()
```


The code lives [here](https://github.com/MortenHolmRep/graphnet/tree/multiclass)
The loss function [here](https://github.com/MortenHolmRep/graphnet/blob/multiclass/src/graphnet/training/loss_functions.py)
The task [here](https://github.com/MortenHolmRep/graphnet/blob/multiclass/src/graphnet/models/task/reconstruction.py)

src/graphnet/training/loss_functions.py
MortenHolmRep: > * Since we're using batching, and since each batch may not contain examples of all classes (unlikely, but not impossible), I think we need to specify `one_hot(target, num_classes=num_classes)`. (I also don't think you need to set device explicitly?)

This is a really good point about classes that I did not take into account.
The device issue was just what stopped the code from running when using hot. I will try with and without the device statement in your suggestion.

> * I think we should allow for either:
>   
>   * `target` is an integer in `[0, num_classes[` and thus one_hot can be used directly.
>   * `target` is not that, and an optional list of classes will be provide to `MulticlassClassificationTask(int: nb_classes, Optional[List[Any]]: classes = None)`, and then in the `CrossEntropyLoss` class, do something like:
> 
> ```
> def _forward(self, prediction: Tensor, target: Tensor) -> Tensor:
>         if self._classes is not None:
>             assert isinstance(self._classes, Tensor)  # Cast to tensor in the constructor
>             target_integer = torch.nonzero(target == self._classes.unsqueeze(dim=1), as_tuple=True)[0]
>         else:
>             assert target.dtype in [torch.int32, torch.int64]
>             assert torch.all(target >= 0)
>             assert torch.all(target < self._nb_classes)
>             target_integer = target
> 
>         target_one_hot = one_hot(target_integer)
>         return cross_entropy(
>             prediction.float(), target_one_hot.float(), reduction="none"
>         ) 
> ```

I will give this a go :)

But we still need to address the use case from the dict transformer we used before.
the target `pid` input can be both a positive and negative integer, i.e. noise is `{1,-1}` and the same for muons and neutrinos.
How do we transform the `pid` according to the training we want? for example, the static case that was first introduced where we have a classification of whether we have noise, muons or neutrinos, regardless of flavor?

MortenHolmRep:
Thanks for your review @asogaard! I have added some comments for now. The suggestion you made I'll have to test later today.

MortenHolmRep:
Sorry in advance for the messy upload. I had to adhere to the new pre-commit hooks and still learning the type hinting stuff :)

The implementation we talked about is done. However, we still do not get the correct predictions out in the end.
The current implementation in our static multiclass branch, use the prediction transformation. Do you have a suggestion or should use the transformation?

MortenHolmRep:
Despite comparing my working branch with a static 3 classification classes, I have not been able to resolve the generic one in this branch.

In the following, I only print the process for the first 5 elements at each step, but that training itself was done on a sufficient set.
In testing the outputs of the various steps we get in `src/graphnet/models/task/classification.py`:
line 29: 
```
x = tensor([[ 1.0707,  0.5128, -0.0091],
        [ 1.1699,  0.1510, -0.2159],
        [ 0.8537,  0.4584, -0.3231],
        [ 2.7514,  1.1381,  0.1574],
        [ 2.4920,  0.5496, -0.7172]])
```

in `src/graphnet/training/loss_functions.py`
line 158 `nb_classes=3`
line 160 `target_integer = tensor([2, 2, 2, 1, 1])`
line 167:
```
target_new = tensor([[0, 0, 1],
        [0, 0, 1],
        [0, 0, 1],
        [0, 1, 0],
        [0, 1, 0]])
```
line 169 `loss = tensor([1.7280, 1.8627, 1.8607, 1.8555, 2.1111])`

returning to the classification.py script
```
x = tensor([[ 1.3362,  0.5268, -0.0676],
        [ 2.1126,  0.8355, -0.1429],
        [ 1.5445,  0.3405, -0.2234],
        [ 1.4625,  0.4801, -0.1694],
        [ 1.1249,  0.4071, -0.1924]], grad_fn=<SliceBackward0>)
```
and again for loss
`loss = tensor([1.3346, 1.6018, 0.3857, 0.4511, 1.8802], grad_fn=<SliceBackward0>)`

at some point, we then get to `get_predictions()` which outputs a result, here shown for the first 25, that all predictions just falls back to one class dominating.
```
    pid_noise_pred  pid_muon_pred  pid_neutrino_pred   pid    event_no
0     8.702452e-05       0.956136       4.377704e-02  14.0  84440754.0
1     1.593457e-05       0.967339       3.264529e-02  12.0  91827257.0
2     2.576510e-04       0.944390       5.535277e-02  14.0  33468682.0
3     3.354662e-09       0.998212       1.787505e-03  13.0  98059030.0
4     4.872288e-09       0.997223       2.777350e-03  13.0   5995775.0
5     8.474670e-05       0.955719       4.419648e-02  -1.0  41854280.0
6     1.647413e-04       0.943988       5.584761e-02 -12.0  97405864.0
7     1.869076e-06       0.980326       1.967184e-02  -1.0  61830022.0
8     5.690738e-07       0.990746       9.253046e-03  13.0  69901250.0
9     1.532002e-08       0.997376       2.623657e-03  13.0  70110229.0
10    1.437597e-04       0.952118       4.773793e-02  14.0  59292859.0
11    6.329699e-12       0.999704       2.955984e-04  13.0  48470148.0
12    5.159412e-07       0.991909       8.090075e-03  13.0  67834372.0
13    1.397524e-21       1.000000       2.277882e-07  12.0  59692614.0
14    1.294005e-07       0.993294       6.705615e-03  13.0  36965785.0
15    9.180514e-10       0.998568       1.431756e-03  13.0  97338461.0
16    1.874697e-08       0.996092       3.907592e-03  16.0  53355077.0
17    1.468242e-04       0.929612       7.024118e-02  16.0  22959794.0
18    4.150671e-06       0.973754       2.624215e-02  -1.0   8894064.0
19    1.136094e-06       0.984150       1.584838e-02  13.0  70706937.0
20    8.326127e-06       0.953914       4.607764e-02 -12.0    946169.0
21    1.563201e-05       0.969463       3.052181e-02  14.0  20603481.0
22    7.137892e-15       0.999975       2.511165e-05  13.0  36216144.0
23    3.502722e-08       0.993866       6.133591e-03  -1.0  59089881.0
24    1.807485e-08       0.997106       2.894209e-03  13.0  50010669.0
```

When I run the same training on my static, 3 classes-only branch, I get the correct results. At some point in `get_predictions()`, the result changes.
I have tried a myriad of variations such as, but not limited to:

- not having the `transform_inference=lambda x: softmax(x, dim=-1),` in line 101 of `/graphnet/examples/train_classification_model.py`
- in `src/graphnet/models/task/classification.py`, line 20 tried `self._softmax = Softmax(dim=-1)` with line 29 as `return self._softmax(x)`
- in `src/graphnet/models/task/classification.py`, line 20 tried `self._softmax = None` with line 29 as `x`
- in `src/graphnet/models/task/classification.py`, line 20 tried `self._softmax = None` with line 29 as `torch.nn.functional.softmax(x)` // as functional.softmax() cannot be pre instantiated in line 20

MortenHolmRep:
@asogaard Can you review the PR again?
I have attempted to clean the code, config and unused packages. The generic classification now works and produces good results.
The build fails for the `LogCMK `class in loss_functions, which is unrelated to my implementation, I do not know why I am getting this error.

MortenHolmRep (commented):
Should be resolved individually

configs/datasets/classification_last_one_lvl3MC.yml
```diff
@@ -0,0 +1,43 @@
+path: /groups/icecube/petersen/GraphNetDatabaseRepository/Leon2022_DataAndMC_CSVandDB_StoppedMuons/last_one_lvl3MC.db
```
asogaard, line 1: If you add this, and the ditto model config file, I figure you should use these in your example script. Otherwise I would just keep these in the relevant analysis repo.

configs/models/dynedge_classification_noise_muon_neutrino.yml
```diff
@@ -0,0 +1,44 @@
+arguments:
```
asogaard, line 1: See comment to dataset config file.

examples/train_classification_model.py
```diff
@@ -0,0 +1,199 @@
...
+        validation_dataloader,
+    ) = make_train_validation_dataloader(
+        cast(str, config["db"]),
+        train_selection,
+        cast(str, config["pulsemap"]),
+        features,
+        truth,
+        batch_size=cast(int, config["batch_size"]),
+        num_workers=cast(int, config["num_workers"]),
+    )
```
asogaard, line 86: See comment regarding dataset config file.

examples/train_classification_model.py
```diff
@@ -0,0 +1,199 @@
...
+        loss_function=CrossEntropyLoss(options=config["class_options"]),
+        transform_inference=lambda x: softmax(x, dim=-1),
+    )
+    model = StandardModel(
+        detector=detector,
+        gnn=gnn,
+        tasks=[task],
+        optimizer_class=Adam,
+        optimizer_kwargs={"lr": 1e-04, "eps": 1e-03},
+    )
```
asogaard, line 109: See comment regarding model config file.

examples/train_classification_model.py
```diff
@@ -0,0 +1,199 @@
...
+        callbacks=callbacks,
+        log_every_n_steps=1,
+        logger=wandb_logger,
+    )
+
+    try:
+        trainer.fit(model, training_dataloader, validation_dataloader)
+    except KeyboardInterrupt:
+        logger.warning("[ctrl+c] Exiting gracefully.")
+        pass
```
asogaard, line 133: I would suggest you use the new `Model.fit(...)` method.

examples/train_classification_model.py
```diff
@@ -0,0 +1,199 @@
...
+        trainer,
+        model,
+        validation_dataloader,
+        [
+            cast(str, config["target"]) + "_noise_pred",
+            cast(str, config["target"]) + "_muon_pred",
+            cast(str, config["target"]) + "_neutrino_pred",
+        ],
+        additional_attributes=[cast(str, config["target"]), "event_no"],
+    )
```
asogaard, line 146: I would suggest you use the new `Model.predict{,_as_dataframe}(...)` method(s).

src/graphnet/models/task/classification.py
```diff
@@ -1,9 +1,33 @@
...
+        self._softmax = None
+
+    @property
+    def nb_inputs(self) -> Tensor:
+        """Output from Classification."""
+        return self._nb_classes
+
+    def _forward(self, x: Tensor) -> Tensor:
+        """Transform latent features into probabilities."""
+        return x
```
asogaard, line 30: Leverage existing class to avoid duplication.
```suggestion
class MulticlassClassificationTask(IdentityTask):
    """General task for classifying any number of classes.

    Requires the same number of input features as the number of classes being
    predicted. Returns the untransformed latent features, which are interpreted
    as the logits for each class being classified.
    """
```

src/graphnet/models/task/classification.py
```diff
@@ -1,9 +1,33 @@
 """Classification-specific `Model` class(es)."""
 
-import torch
-from torch import Tensor
+from torch import Tensor, sigmoid
+from torch.nn import Softmax
+from graphnet.utilities.config.model_config import save_model_config
 
 from graphnet.models.task import Task
+from typing import Any
```
asogaard, line 8: Group imports as (A) builtin, (B) installed, (C) within-package. Sort imports alphabetically.

```suggestion
from typing import Any

import torch
from torch import Tensor

from graphnet.models.task import Task
from graphnet.utilities.config.model_config import save_model_config
```

src/graphnet/models/task/classification.py
```diff
@@ -14,7 +38,7 @@ class BinaryClassificationTask(Task):
 
     def _forward(self, x: Tensor) -> Tensor:
         # transform probability of being muon
-        return torch.sigmoid(x)
+        return sigmoid(x)
```
asogaard, line 41: Change is not pertinent to the goal of the PR.
```suggestion
        return torch.sigmoid(x)
```

src/graphnet/models/task/reconstruction.py
```diff
@@ -141,14 +141,6 @@ def _forward(self, x: Tensor) -> Tensor:
 class TimeReconstruction(Task):
     """Reconstructs time."""
 
-    # Requires one feature, time.
```
asogaard, line 144: I suggest you revert this change — it is not pertinent to the goal of the PR.

src/graphnet/training/utils.py
```diff
@@ -2,7 +2,7 @@
 
 from collections import OrderedDict
 import os
-from typing import Dict, List, Optional, Tuple, Union
+from typing import Dict, List, Optional, Tuple, Union, Any
```
asogaard, line 5: Change is not pertinent to the goal of the PR.
```suggestion
from typing import Dict, List, Optional, Tuple, Union
```

tests/training/test_loss_functions.py
```diff
@@ -6,7 +6,10 @@
 from torch import Tensor
 from torch.autograd import grad
 
-from graphnet.training.loss_functions import LogCoshLoss, VonMisesFisherLoss
+from graphnet.training.loss_functions import (
+    LogCoshLoss,
+    VonMisesFisherLoss,
+)
```
asogaard, line 12: Change is not pertinent to the goal of the PR.
```suggestion
from graphnet.training.loss_functions import LogCoshLoss, VonMisesFisherLoss
```

src/graphnet/models/task/task.py
```diff
@@ -173,9 +174,6 @@ def _validate_and_set_transforms(
             (transform_prediction_and_target is not None)
             and (transform_target is not None)
         ), "Please specify at most one of `transform_prediction_and_target` and `transform_target`"
-        assert (transform_target is not None) == (
-            transform_inference is not None
-        ), "Please specify both `transform_inference` and `transform_target`"
```
asogaard, line 177:
```suggestion
if (transform_target is not None) != (transform_inference is not None):
    self.warning(
        "Setting one of `transform_target` and `transform_inference`, but not "
        "the other."
    )

```

src/graphnet/training/loss_functions.py
```diff
@@ -105,6 +112,65 @@ def _forward(self, prediction: Tensor, target: Tensor) -> Tensor:
...
+            self.error(f"Type {type(self._options)} not supported")
+
+        # pid_transform = {1:0,12:2,13:1,14:2,16:2}
+        device = prediction.device
+        target_new: Tensor = one_hot(target_integer, nb_classes).to(device)
+
+        loss = nn.CrossEntropyLoss(reduction="none")
+        return loss(prediction.float(), target_new.float())
+
+
```
asogaard, line 173:
```suggestion
class CrossEntropyLoss(LossFunction):
    """Compute cross-entropy loss for classification tasks.

    Predictions are an [N, num_class]-matrix of logits (i.e., non-softmax'ed
    probabilities), and targets are an [N,1]-matrix with integer values in 
    (0, num_classes - 1).
    """

    @save_model_config
    def __init__(
        self,
        options: Union[int, List[Any], Dict[Any, int]],
        *args: Any,
        **kwargs: Any,
    ):
        """Construct CrossEntropyLoss."""
        # Base class constructor
        super().__init__(*args, **kwargs)
        
        # Member variables
        self._options = options
        self._nb_classes: int
        if isinstance(self._options, int):
            assert self._options in [torch.int32, torch.int64]
            assert self._options >= 2, (
                f"Minimum of two classes required. Got {self._options}."
            )
            self._nb_classes = options
        elif isinstance(self._options, list):
            self._nb_classes = len(self._options) 
        elif isinstance(self._options, dict):
            self._nb_classes = len(np.unique(list(self._options.values())))
        else:
            raise ValueError(
                f"Class options of type {type(self._options)} not supported"
            )
            
        self._loss = nn.CrossEntropyLoss(reduction="none")

    def _forward(self, prediction: Tensor, target: Tensor) -> Tensor:
        """Transform outputs to angle and prepare prediction."""
        if isinstance(self._options, int):
            # Integer number of classes: Targets are expected to be in 
            # (0, nb_classes - 1).
            
            # Target integers are positive
            assert torch.all(target >= 0)

            # Target integers are consistent with the expected number of class.
            assert torch.all(target < self._options)

            assert target.dtype in [torch.int32, torch.int64]
            target_integer = target

        elif isinstance(self._options, list):
            # List of classes: Mapping target classes in list onto 
            # (0, nb_classes - 1). Example:
            #    Given options: [1, 12, 13, ...]
            #    Yields: [1, 13, 12] -> [0, 2, 1, ...]
            target_integer = torch.tensor(
                [self._options.index(value) for value in target]
            )

        elif isinstance(self._options, dict):
            # Dictionary of classes: Mapping target classes in dict onto 
            # (0, nb_classes - 1). Example:
            #     Given options: {1: 0, -1: 0, 12: 1, -12: 1, ...}
            #     Yields: [1, -1, -12, ...] -> [0, 0, 1, ...]
            target_integer = torch.tensor(
                [self._options[int(value)] for value in target]
            )
            
        else:
            assert False, "Shouldn't reach here."

        target_one_hot: Tensor = (
            one_hot(target_integer, nb_classes)
            .to(prediction.device)
        )
        
        return self._loss(prediction.float(), target_one_hot.float())


```

src/graphnet/training/loss_functions.py
```diff
@@ -15,6 +15,13 @@
 from graphnet.utilities.config import save_model_config
 from graphnet.models.model import Model
 from graphnet.utilities.decorators import final
+from torch import nn
```
asogaard, line 18: I don't think this line is needed?
```suggestion

```

asogaard (changes requested):
Hi @MortenHolmRep,

I have gone through the PR again and made some comments and suggestions. :)  Let me know what you think!

src/graphnet/models/task/classification.py
MortenHolmRep: Makes sense, I will do this moving forward 🙂

tests/training/test_loss_functions.py
MortenHolmRep: This is actually something black did at some point. How do I apply black only to my own code?

src/graphnet/models/task/reconstruction.py
MortenHolmRep: I am really confused about why this happened, as I do not recall changing this function. I was able to locate the change to commit ["33beeb97d86e3afd46bd8b245bd9592504159911" ](https://github.com/MortenHolmRep/graphnet/commit/33beeb97d86e3afd46bd8b245bd9592504159911) during a merge from main.

MortenHolmRep: The function has been restored.

examples/train_classification_model.py
MortenHolmRep: added model from config

examples/train_classification_model.py
MortenHolmRep: added training from config

examples/train_classification_model.py
MortenHolmRep: added a prediction segment, but need help setting it up.

examples/train_classification_model.py
MortenHolmRep: the script has been reconfigured to match the new config setup, but I need help setting up the datasets file.

MortenHolmRep:
I have added a script called "train_classification_model_without_configs.py" that follows the old structure and a new script called "train_classification_model.py" that should adhere to the new structure.
The new setup with configs is not done as I need input on how to set it up properly.

asogaard (approved):
Hi @MortenHolmRep,

I think this looks great! If it runs as expected, I am all for merging this. There is still the problem of the failing unit test that needs to be resolved. I think it is just a matter of deleting [this block](https://github.com/graphnet-team/graphnet/blob/main/tests/models/test_task.py#L39-L52) as it no longer represents expected behaviour. :)

PR #334, opened 2022-11-01
Update README.md

MortenHolmRep:
update to active projects
* Peter and Morten are working on multi class classification
* Rasmus, Kaare and Morten have all worked on the noise cleaning, each with a part of the solution. It is, however, not implemented (to my knowledge) as a usable feature.


asogaard (approved):
🚀

PR #339, opened 2022-11-02
added i3pisaextractor

RasmusOrsoe:
Replaces the stale  #176 with a new extractor class `I3PISAExtractor` . Confirmed to work.


asogaard (approved):
🚀

PR #340, opened 2022-11-02
Fix for I3PulseNoiseTruthFlag extractor issue

### LINKED ISSUE #337: I3PulseNoiseTruthFlagIceCubeUpgrade doesn't work anymore
**Describe the bug**
`I3PulseNoiseTruthFlagIceCubeUpgrade` fails because of inheritance to `I3FeatureExtractorIceCube86 `. 

In `__call__` the lines

```
output_icecube86 = super().__call__(frame)
output.update(output_icecube86)
```
the `I3PulseNoiseTruthFlagIceCubeUpgrade` extractor asks `I3FeatureExtractorIceCube86 ` to retrieve hardware information (DOM position, rde etc) along with pulse information (charge, time, width). However, pulse noise flags are not pulses, and the extraction fails because `pulse.charge` doesn't exist.

**To Reproduce**
1. Download the files: `/home/rorsoe/phd/upgrade/test_data`
2. Run this

```

import logging
import os

from graphnet.utilities.logging import get_logger

from graphnet.data.extractors import (
    I3FeatureExtractorIceCubeUpgrade,
    I3TruthExtractor,
    I3GenericExtractor,
    I3PulseNoiseTruthFlagIceCubeUpgrade,
)

from graphnet.data.sqlite import SQLiteDataConverter

logger = get_logger(level=logging.INFO)

CONVERTER_CLASS = {
    "sqlite": SQLiteDataConverter,
    "parquet": ParquetDataConverter,
}


def main_icecube_upgrade(backend: str):
    """Convert IceCube-Upgrade I3 files to intermediate `backend` format."""
    # Check(s)
    assert backend in CONVERTER_CLASS
    database_name = 'dev_step4_upgrade_0028_numu_nutau_01.11.202_test_debug_v2'
    inputs = ["/home/rorsoe/phd/upgrade/test_data"]
    outdir = f"debug_test/{database_name}"
    workers = 3
    converter = CONVERTER_CLASS[backend](
        [
            I3TruthExtractor(),
          I3FeatureExtractorIceCubeUpgrade(
                "SplitInIcePulses"
            ),
          I3PulseNoiseTruthFlagIceCubeUpgrade(
                "SplitInIcePulses_TruthFlags"
            ),
            I3FeatureExtractorIceCubeUpgrade(
              "SplitIceCubePulsesTWSRT"
            ),
          I3PulseNoiseTruthFlagIceCubeUpgrade(
               "SplitIceCubePulsesTWSRT_TruthFlags"
           ),
          I3FeatureExtractorIceCubeUpgrade(
             "SplitInIcePulses_GraphSage_Pulses"
           ),
          I3PulseNoiseTruthFlagIceCubeUpgrade(
               "SplitInIcePulses_GraphSage_Predictions"
            ),
        ],
        outdir + '/tmp',
        workers=workers,
        icetray_verbose=1,
        gcd_rescue = '/data/sim/IceCubeUpgrade/genie/step4/140028/GeoCalibDetectorStatus_ICUpgrade.v58.mixed.V0.i3.bz2'
    )
    os.makedirs(outdir + '/data/')
    converter(inputs)
    if backend == "sqlite":
        converter.merge_files(outdir + '/data/' + database_name)


if __name__ == "__main__":
    backend = "sqlite"
    main_icecube_upgrade(backend)
```

3. See error

**Expected behavior**
`I3PulseNoiseTruthFlagIceCubeUpgrade` should not ask for pulse information. (should run and complete without issue)

RasmusOrsoe:
Fixes #337 . Tested on upgrade data.

### DIFF
--- a/src/graphnet/data/extractors/i3featureextractor.py
+++ b/src/graphnet/data/extractors/i3featureextractor.py
@@ -185,7 +185,6 @@ class I3PulseNoiseTruthFlagIceCubeUpgrade(I3FeatureExtractorIceCubeUpgrade):
         """Extract features to be used as inputs to GNN models."""
 
         output = {
-            "dom_type": [],
             "truth_flag": [],
         }
 
--- a/src/graphnet/data/extractors/i3featureextractor.py
+++ b/src/graphnet/data/extractors/i3featureextractor.py
@@ -15,9 +15,12 @@ class I3FeatureExtractor(I3Extractor):
 
 
 class I3FeatureExtractorIceCube86(I3FeatureExtractor):
-    def __call__(self, frame, padding_value=-1, include_pulses=True) -> dict:
+    def __call__(self, frame, padding_value=-1) -> dict:
         """Extract features to be used as inputs to GNN models."""
         output = {
+            "charge": [],
+            "dom_time": [],
+            "width": [],
             "dom_x": [],
             "dom_y": [],
             "dom_z": [],
@@ -29,8 +32,6 @@ class I3FeatureExtractorIceCube86(I3FeatureExtractor):
             "is_errata_dom": [],
             "event_time": [],
         }
-        if include_pulses:
-            output.update({"charge": [], "dom_time": [], "width": []})
 
         # Get OM data
         om_keys, data = get_om_keys_and_pulseseries(
@@ -92,10 +93,13 @@ class I3FeatureExtractorIceCube86(I3FeatureExtractor):
             # Loop over pulses for each OM
             pulses = data[om_key]
             for pulse in pulses:
-                if include_pulses:
-                    output["charge"].append(pulse.charge)
-                    output["dom_time"].append(pulse.time)
-                    output["width"].append(pulse.width)
+                output["charge"].append(
+                    getattr(pulse, "charge", padding_value)
+                )
+                output["dom_time"].append(
+                    getattr(pulse, "time", padding_value)
+                )
+                output["width"].append(getattr(pulse, "width", padding_value))
                 output["pmt_area"].append(area)
                 output["rde"].append(rde)
                 output["dom_x"].append(x)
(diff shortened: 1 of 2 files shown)


asogaard (commented):
Hi @RasmusOrsoe,

Thanks for proposing a fix! A few questions:
* `I3PulseNoiseTruthFlagIceCubeUpgrade` inherits from `I3FeatureExtractorIceCube86`, but wouldn't it make more sense for it to inherit from `I3FeatureExtractorIceCubeUpgrade` so as to avoid having to re-derive `string`, `pmt_dir_x`, etc.?
* If the purpose of `I3PulseNoiseTruthFlagIceCubeUpgrade` is to extract a truth label (`truth_flag`) would it make more sense to put it in `i3truthextractor.py` and remove the dependence on `I3FeatureExtractor*` entirely, thereby avoiding the cause of the problem you've encountered altogether?
    * If not, would it then be an option to use e.g. `getattr(pulse, "charge", padding_value)` instead of the `include_pulses` argument to fall back to a default of `padding_value` in case we get a pulse map that doesn't have the `charge` attribute?

RasmusOrsoe:
> Hi @RasmusOrsoe,
> 
> Thanks for proposing a fix! A few questions:
> 
>     * `I3PulseNoiseTruthFlagIceCubeUpgrade` inherits from `I3FeatureExtractorIceCube86`, but wouldn't it make more sense for it to inherit from `I3FeatureExtractorIceCubeUpgrade` so as to avoid having to re-derive `string`, `pmt_dir_x`, etc.?
> 
Yeah, I would think so. I did not change that part, but perhaps I should.
>     * If the purpose of `I3PulseNoiseTruthFlagIceCubeUpgrade` is to extract a truth label (`truth_flag`) would it make more sense to put it in `i3truthextractor.py` and remove the dependence on `I3FeatureExtractor*` entirely, thereby avoiding the cause of the problem you've encountered altogether?

Conceptually, I agree that this would be the best way to go. But technically, it's important to include DOM ids, PMT ids, perhaps even positions, in order to make sure that the flags are sorted correctly according to the pulsemap.        
>       * If not, would it then be an option to use e.g. `getattr(pulse, "charge", padding_value)` instead of the `include_pulses` argument to fall back to a default of `padding_value` in case we get a pulse map that doesn't have the `charge` attribute?
I suspect that would be a much more elegant solution. I did not think of that!

RasmusOrsoe:
@asogaard this should be it, I think

asogaard (approved):
Fantastic!

PR #346, opened 2022-11-10
Fix Logger typing

asogaard:
The use and typing of the `logger` from `get_logger` was slightly mishandled in #310 leading to an error in the `ProgressBar` callback. This PR fixes this.


asogaard:
cc @Aske-Rosted

RasmusOrsoe (approved):

PR #348, opened 2022-11-10
Add dedicated Detector class for IceCube-86/main array

asogaard:
So far, the `IceCube86Detector` and `IceCubeDeepCoreDetector` classes have been identical, but the preprocessing used has been tuned to DeepCore events, meaning that the normalisation of DOM positions and pulse timing is not well-suited for events that span the entire main array. This PR specialises to the two classes to their respective domains.


asogaard:
cc @Aske-Rosted

RasmusOrsoe (approved):

PR #352, opened 2022-11-11
Energy Task preventing over/under-flow

Aske-Rosted:
This task should be useful if experiencing over/underflow errors leading to NaN's in the training process.


src/graphnet/models/task/reconstruction.py
```diff
@@ -90,6 +90,19 @@ class EnergyReconstruction(Task):
     def _forward(self, x: Tensor) -> Tensor:
         # Transform energy
         return torch.pow(10, x[:, 0] + 1.0).unsqueeze(1)
+    
+class EnergyReconstructionPreventFlow(Task):
+        """Reconstructs energy without transforming."""
```
asogaard, line 95:
```suggestion
class EnergyReconstructionWithoutPowerTransform(Task):
        """Reconstructs energy without power-10 transformation."""
```

src/graphnet/models/task/reconstruction.py
```diff
@@ -90,6 +90,19 @@ class EnergyReconstruction(Task):
...
+
+        # Requires one feature: untransformed energy
+        nb_inputs = 1
+
+        def _forward(self, x):
+            # transform to positive energy domain + smallest float 32 ofsett to ensure no -inf in logcosh
+            # prevents overflow and underflow error
+            softplus = torch.nn.Softplus(beta = 0.05, threshold=20)
+            return softplus(x)+np.nextafter(np.float32(0), np.float32(1))
```
asogaard, line 104: * Updating comment to comply with PEP8.
* Wrapping comment to 80 characters.
* Using `softmax` function rather than instantiating class on each call to `forward()`.
* Using `eps_like(x)` which returns the smallest non-zero value ("epsilon") for the type that `x` has.
```suggestion
            # Transform to positive energy domain to ensure no `-inf` in `log10`
            # transform, thereby preventing overflow and underflow error.
            return torch.nn.functional.softplus(x, beta=0.05) + eps_like(x)
```

asogaard (changes requested):
Hi @Aske-Rosted,
Thanks for this contribution! I have made a few "inline" suggestions that you can consider using. Please also consider setting up pre-commit hooks, see [our contributing guide](https://github.com/graphnet-team/graphnet/blob/main/CONTRIBUTING.md) as currently the automated black and flake8 checks are failing. Let me know if it would be beneficial to talk this over at some point. :)

src/graphnet/models/task/reconstruction.py
```diff
@@ -92,6 +92,18 @@ def _forward(self, x: Tensor) -> Tensor:
         return torch.pow(10, x[:, 0] + 1.0).unsqueeze(1)
 
 
+class EnergyReconstructionWithoutPowerTransform(Task):
+    """Reconstructs energy without power-10 transformation."""
+
+    # Requires one feature: untransformed energy
+    nb_inputs = 1
+
+    def _forward(self, x):
```
asogaard, line 101: `mypy` is giving an error because the method is lacking type annotations. See suggestion below.
```suggestion
    def _forward(self, x: Tensor) -> Tensor:
```

asogaard (commented):
Hi @Aske-Rosted,
I have added one suggestion based on the automated `mypy` check. Another point: Perhaps we should remove the exist `EnergyReconstruction` -> `EnergyReconstructionWithPower` and instead just call the `Task` you're proposing `EnergyReconstruction`, also based on @RasmusOrsoe's point?

src/graphnet/models/task/reconstruction.py
Aske-Rosted: seems sensible I'll try and make the commit.

asogaard (approved):
Brilliant!

asogaard:
Psst, @Aske-Rosted ... you're free to merge this anytime you want. :)

Aske-Rosted:
Not sure how to, thought it would happen automatically when approved...
it says "This branch has no conflicts with the base branch
Only those with [write access](https://docs.github.com/articles/what-are-the-different-access-permissions) to this repository can merge pull requests." on my end...

PR #353, opened 2022-11-12
Add DatasetConfig

asogaard:
Similar to `ModelConfig`.

Can be used for dataset configuration as code, thereby making dataset construction more transparent and reproducible.

Examples:
```python
# In one session, do:
>>> dataset = Dataset(...)
>>> dataset.config.dump()
path: (...)
pulsemaps:
  - (...)
(...)
>>> dataset.config.dump("dataset.yml")

# In another session, you can then do:
>>> dataset = Dataset.from_config("dataset.yml")

# Or identically:
>>> dataset = DatasetConfig.load("dataset.yml").construct_dataset()

# Uniquely for `DatasetConfig`, you can also define and load multiple datasets
>>> dataset.config.selection = {
    "train": "event_no % 2 == 0", 
    "test": "event_no % 2 == 1",
}
>>> dataset.config.dump("dataset.yml")
>>> datasets: Dict[str, Dataset] = Dataset.from_config("dataset.yml")
>>> datasets
{
    "train": Dataset(...),
    "test": Dataset(...),
}
```


asogaard:
Hi @RasmusOrsoe,
This should be ready for scrutiny now. :)  The `DatasetConfig` and `ModelConfig` classes will be central to #262, which I am working on in parallel, since it will allow us to do all of the data and model configuration separately from the training scripts, and share these configs between scripts and developers.

RasmusOrsoe (approved):
Really, really cool!

PR #363, opened 2022-11-30
Reweighter

sdebes:
A function based on the gradient boosted reweighting algorithm hep_ml.reweight.GBReweighter, that reweights Monte Carlo data to look like real data.

PR #376, opened 2022-12-15
Limit awkward-array version, due to breaking changes

asogaard:
Fixes failing build jobs reported in #363, and affecting all automated jobs since last week.


asogaard:
cc @sdebes

MortenHolmRep (approved):
I am confident in your solution fixes the mentioned issue.

PR #377, opened 2022-12-15
Support macOS installation

### LINKED ISSUE #373: Installation fails on macOS
**Describe the bug**
As part of openjournals/joss-reviews#4971 I’m trying to install GraphNeT on my local machine (macOS 12.6.1). Since I’m not an IceCube collaborator and thus don’t have access to IceTray, I’m using the stand-alone installation described in the README file. This fails because the requirement `

**To Reproduce**
1. Start clean shell.
2. Run the following commands:
```bash
$ git clone git@github.com:graphnet-team/graphnet.git
$ cd graphnet
$ conda create --name graphnet python=3.8  # already have current gcc version installed, see below
$ conda activate graphnet
$ pip install -r requirements/torch_cpu.txt -e .[develop,torch]
```
3. See error

**Expected behavior**
GraphNeT is installed.

**Full traceback**
```
Looking in links: https://download.pytorch.org/whl/torch_stable.html, https://data.pyg.org/whl/torch-1.11.0+cpu.html
Obtaining file:///Users/jost/[…]/graphnet
  Preparing metadata (setup.py) ... done
ERROR: Could not find a version that satisfies the requirement torch==1.11+cpu (from versions: 1.4.0, 1.5.0, 1.5.1, 1.6.0, 1.7.0, 1.7.1, 1.8.0, 1.8.1, 1.9.0, 1.9.1, 1.10.0, 1.10.1, 1.10.2, 1.11.0, 1.12.0, 1.12.1, 1.13.0)
ERROR: No matching distribution found for torch==1.11+cpu
```

**Additional context**
`requirements/torch_cpu.txt` requires a version `1.11+cpu`, but the macOS wheels included at https://download.pytorch.org/whl/torch_stable.html are tagged `1.11.0` (without the `+cpu` local version specifier); probably because there are no GPU-specific wheels for macOS? After manually installing the appropriate wheel from the list and commenting out the `torch==1.11+cpu` line in `requirements/torch_cpu.txt`, the installation worked fine otherwise.

asogaard:
Closes #373 

Confirmed to work on a 2021 M1 MacBook Pro running macOS Monterey (12.1).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -49,7 +49,7 @@ jobs:
           pip install --upgrade astropy  # Installed version incompatible with numpy 1.23.0 [https://github.com/astropy/astropy/issues/12534]
           pip install --ignore-installed PyYAML  # Distutils installed [https://github.com/pypa/pip/issues/5247]
       - name: Install package
-        uses: ./.github/actions/install-cpu
+        uses: ./.github/actions/install
         with:
           editable: true
       - name: Run unit tests and generate coverage report
@@ -81,9 +81,30 @@ jobs:
         with:
           python-version: ${{ matrix.python-version }}
       - name: Install package
-        uses: ./.github/actions/install-cpu
+        uses: ./.github/actions/install
         with:
           editable: true
+      - name: Run unit tests and generate coverage report
+        run: |
+          set -o pipefail  # To propagate exit code from pytest
+          coverage run --source=graphnet -m pytest tests/ --ignore=tests/data/
+          coverage report -m
+
+
+  build-matrix:
+    name: Unit tests - macOS
+    runs-on: macos-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: Set up Python 3.8
+        uses: actions/setup-python@v4
+        with:
+          python-version: 3.8
+      - name: Install package
+        uses: ./.github/actions/install
+        with:
+          editable: true
+          hardware: "macOS"
       - name: Run unit tests and generate coverage report
         run: |
           set -o pipefail  # To propagate exit code from pytest
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -90,8 +90,7 @@ jobs:
           coverage run --source=graphnet -m pytest tests/ --ignore=tests/data/
           coverage report -m
 
-
-  build-matrix:
+  build-macos:
     name: Unit tests - macOS
     runs-on: macos-latest
     steps:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -103,7 +103,7 @@ jobs:
         uses: ./.github/actions/install
         with:
           editable: true
-          hardware: "macOS"
+          hardware: "macos"
       - name: Run unit tests and generate coverage report
         run: |
           set -o pipefail  # To propagate exit code from pytest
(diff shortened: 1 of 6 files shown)


.github/workflows/build.yml
```diff
@@ -81,11 +81,31 @@ jobs:
...
+
+  build-macos:
+    name: Unit tests - macOS
+    runs-on: macos-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: Set up Python 3.8
+        uses: actions/setup-python@v4
+        with:
+          python-version: 3.8
```
MortenHolmRep, line 101: Does the mac implementation only support python 3.8?
Or would something like
```
strategy:
      matrix:
        python-version: [3.7, 3.8, 3.9, '3.10']
```
still work?

.github/workflows/build.yml
```diff
@@ -81,11 +81,31 @@ jobs:
         with:
           python-version: ${{ matrix.python-version }}
       - name: Install package
-        uses: ./.github/actions/install-cpu
+        uses: ./.github/actions/install
         with:
           editable: true
```
MortenHolmRep, line 86: Should the other with statements also have the hardware specification? i.e.
```
with:
          editable: true
          hardware: "cpu"
```

MortenHolmRep (commented):
Great with a mac support addition!
I have left a few comments on the structure regarding the new hardware statement, to make sure it does not break the old installation.

.github/workflows/build.yml
asogaard: That should work just as well too! The reason I opted for a single, fixed version is:

* We only test for support for python versions 3.8 through 3.10, so I figures it was good to pick one, fixed version to test against rather than using whatever comes with `macos-latest` by default (which could change over time).
* Running on the macOS image is [expensive!](https://docs.github.com/en/billing/managing-billing-for-github-actions/about-billing-for-github-actions#minute-multipliers) On the free tier, we have a fixed number of free GitHub Actions minutes and running on macOS "eats" those minutes x10 faster than running on Ubuntu. Since we're testing for different python versions in another job (the matrix one) I figured it was fine to have this job dedicated just to testing the OS-specific differences. (Assuming, then, that the python version results from the matrix build job would carry over to macOS as well.)

.github/workflows/build.yml
asogaard: The default choice for `hardware` is `"cpu"`; see [here](https://github.com/graphnet-team/graphnet/blob/7124c1090f12f8296f26962a043f411b6e88360a/.github/actions/install/action.yml#L20). So, it would be perfectly possible to specify `hardware: "cpu"` as you mention above, but since there is no _need_ to, and since we will only choose a setting other than `"cpu"` on very rare occasions, I think it's fine to omit.

asogaard:
Thanks for the comments, @MortenHolmRep! (I'll let you resolve them, once you've had a chance to look at them.) The good thing is that the changes to the GitHub Actions (which is most of what's in this PR) are, by definition, automatically checked by the automated checks of this PR. :)  So as long as they're all green, we should be pretty safe.

.github/workflows/build.yml
MortenHolmRep: These are very good reasons, I'll consider them resolved!

.github/workflows/build.yml
MortenHolmRep: Personally, I like to have it for transparency, however, both work and I missed the default when reviewing. I'll consider it resolved :D

MortenHolmRep (approved):
Great addition!

.github/workflows/build.yml
JostMigenda: Just FYI, my understanding is that those 2000 minutes are for private repos; GitHub Actions on public repos are not limited. But I agree that testing once for OS-specific differences should be enough.

PR #375, opened 2022-12-13
Multiclass configs

MortenHolmRep:
[train_classification_model.py](https://github.com/graphnet-team/graphnet/blob/main/examples/train_classification_model.py) was not configured properly, which this PR aimed to fix.

This version was tested on a CPU, using my own dataset and worked.
The current implementation uses the example dataset `dev_lvl7_robustness_muon_neutrino_0000` provided by @asogaard, but there is an issue with writing temporary pytorch lightning logs in the data folder that keeps me from testing the example dataset.

PR #378, opened 2023-01-04
Restructure examples

### LINKED ISSUE #374: Example scripts fail after stand-alone install
As part of https://github.com/openjournals/joss-reviews/issues/4971 I have installed GraphNeT as stand-alone (without IceTray) both on my personal laptop (macOS 12.6.1) and on our university cluster (Ubuntu 20.04) and am trying to run some of the GraphNeT example scripts included in the repo.

Unfortunately, I can use neither of the first four example scripts. On both systems, I get the following errors:
* `compare_sqlite_and_parquet.py`
```bash
$ python examples/compare_sqlite_and_parquet.py 
graphnet: INFO     2022-12-13 17:21:29 - get_logger - Writing log to logs/graphnet_20221213-172129.log
graphnet: WARNING  2022-12-13 17:21:31 - warn_once - `icecube` not available. Some functionality may be missing.
  File "examples/compare_sqlite_and_parquet.py", line 78, in <module>
    convert_data()
  File "examples/compare_sqlite_and_parquet.py", line 45, in convert_data
    SQLiteDataConverter(**opt)(paths)  # type: ignore[arg-type]
  File "/users/k2032402/graphnet/graphnet/src/graphnet/data/dataconverter.py", line 188, in __init__
    icetray.I3Logger.global_logger = icetray.I3NullLogger()
NameError: name 'icetray' is not defined
```
* `convert_i3_files.py` fails with the same `NameError` (also in `data/dataconverter.py`, line 188)
* `contour_fitting.py` contains some hard-coded paths like `/mnt/scratch/rasmus_orsoe/…` or `/home/iwsatlas1/oersoe/…` and additionally fails with a `ModuleNotFoundError`:
```bash
$ python examples/contour_fitting.py
  File "examples/contour_fitting.py", line 3, in <module>
    from graphnet.pisa.fitting import ContourFitter
  File "/users/k2032402/graphnet/graphnet/src/graphnet/pisa/fitting.py", line 17, in <module>
    import pisa
ModuleNotFoundError: No module named 'pisa'
```
* `contour_plotting.py` again contains some hard-coded paths like `/mnt/scratch/rasmus_orsoe/…` or `/home/iwsatlas1/oersoe/…`
* (I’ve stopped trying after those four; but a quick glance at the code indicates that at least the hard-coded paths are present in several other example scripts.)

I understand some example scripts may require IceTray-specific functionality or input files that cannot be included in the GraphNeT repo for various reasons. Still, it is quite frustrating to try several example scripts and run into various errors.
At the very least, it would be useful to include an `examples/README.md` file that lists what examples will run with a stand-alone install and which ones require IceTray and/or non-public input files.

asogaard:
Restructures the example scripts by grouping them and making clear which scripts rely on other packages (IceTray and PISA):

1. Requiring IceTray
2. Data — No other dependencies
3. Weights — No other dependencies
4. Training — No other dependencies
5. Requiring PISA

Closes #374 

cc @JostMigenda


asogaard:
Alright, that should be all tests passing. Let me know what you think, @RasmusOrsoe. 

Please not that I have tried to generalise the training examples, and as a result I have removed:
* `train_classification_model.py`
* `train_classification_model_without_configs.py`
Because the config didn't seem to work (see #375) and because the example data from Prometheus doesn't have data that allow for classification (it's all nu_mu events as far as I can tell). cc @MortenHolmRep 

Similarly I have removed:
* `train_position_reconstruction_model.py`
* `train_direction_reconstruction.py`
But the functionality contained is new covered by `train_classification_model.py`. This is decribed in the new `examples/04_training/README.md` file.

RasmusOrsoe (approved):

MortenHolmRep:
@asogaard Should the multi-classification example be included in some other capacity? Like a separate folder where it is stated that it only works for specific datasets?

asogaard:
> @asogaard Should the multi-classification example be included in some other capacity? Like a separate folder where it is stated that it only works for specific datasets?

I think it should definitely be included, but only as a self-contained example. So we would need some other/additional generic example data that we could use to demonstrate this use case. I am sure there exists some Prometheus data that'll do the job (or maybe the new Kaggle stuff), but until then I'd rather avoid examples that cannot be run out of the box.

MortenHolmRep:
> > @asogaard Should the multi-classification example be included in some other capacity? Like a separate folder where it is stated that it only works for specific datasets?
> 
> I think it should definitely be included, but only as a self-contained example. So we would need some other/additional generic example data that we could use to demonstrate this use case. I am sure there exists some Prometheus data that'll do the job (or maybe the new Kaggle stuff), but until then I'd rather avoid examples that cannot be run out of the box.

For now we could create a ~100 data point example .db that imitates the real one but with gibberish data and add it to datasets. I can create a new issue in Feb and you can remove the multi class example for now. How does that sound?

Issue #381, opened 2023-01-11
f"Pulse series {pulseseries} does not exist in `frame`"


RasmusOrsoe:
**Describe the bug**
Since this [commit](https://github.com/RasmusOrsoe/graphnet/commit/07a040211c4cff601d3d6fbc682e495c8f40593f) an error will be raised if a specified pulsemap is not in the frame. The result of this is that conversion stops if any one pulsemap is not present in any one frame. We used to have an expection in these cases, where the extractor would just pass an empty pulsemap. I believe the current logic is a bug.

**To Reproduce**
Run the i3-to-sqlite example where the feature extractor is asked to extract a pulsemap that is not in the files.

**Expected behavior**
We need the extractors to return an empty pulsemap for pulsemaps that are not in the frame, and to continue extraction for certain projects. One such project is converting Upgrade sim data to sqlite - here the muon/neutrino files have certain pulsemaps that the pure noise events doesn't. As the code stands now, this extraction will not complete because the error is raised for every noise event.

**Full traceback**
Please include the full error message to allow for debugging
```
Traceback (most recent call last):
  File "make_database.py", line 79, in <module>
    main_icecube_upgrade(backend)
  File "make_database.py", line 72, in main_icecube_upgrade
    converter(inputs)
  File "/lustre/hpc/hep/pcs557/github/graphnet/src/graphnet/data/dataconverter.py", line 215, in __call__
    self.execute(filesets)
  File "/lustre/hpc/hep/pcs557/github/graphnet/src/graphnet/data/dataconverter.py", line 286, in execute
    pool = self._iterate_over_individual_files(filesets)
  File "/lustre/hpc/hep/pcs557/github/graphnet/src/graphnet/data/dataconverter.py", line 331, in _iterate_over_individual_files
    self._process_file, tqdm(args, unit="file(s)", colour="green")
  File "/lustre/hpc/hep/pcs557/github/graphnet/src/graphnet/data/dataconverter.py", line 83, in wrapper
    output_file = process_method(self, *args)
  File "/lustre/hpc/hep/pcs557/github/graphnet/src/graphnet/data/dataconverter.py", line 377, in _process_file
    data = self._extract_data(fileset)
  File "/lustre/hpc/hep/pcs557/github/graphnet/src/graphnet/data/dataconverter.py", line 434, in _extract_data
    results = self._extractors(frame)
[...]

Issue #380, opened 2023-01-11
sqlite3.OperationalError: table pisa_dependencies has no column named GENIEWeight


RasmusOrsoe:
**Describe the bug**
`i3pisaextractor` returns an empty dictionary if `I3MCWeightDict` is not in frame. This is a problem because, if you convert a mix of i3 files where some do have the `I3MCWeightDict` (Muons, neutrinos), and some do not (pure noise events, real data) the converter will fail on merging.

**To Reproduce**
try to convert and then merge any mix of i3-files to sqlite, where some i3 files have `I3MCWeightDict` and some don't.

**Expected behavior**
Should return a padded dictionary.

**Full traceback**
Please include the full error message to allow for debugging
```
Traceback (most recent call last):
  File "/groups/hep/pcs557/.local/lib/python3.7/site-packages/SQLAlchemy-1.4.27-py3.7-linux-x86_64.egg/sqlalchemy/engine/base.py", line 1783, in _execute_context
    cursor, statement, parameters, context
  File "/groups/hep/pcs557/.local/lib/python3.7/site-packages/SQLAlchemy-1.4.27-py3.7-linux-x86_64.egg/sqlalchemy/engine/default.py", line 716, in do_executemany
    cursor.executemany(statement, parameters)
sqlite3.OperationalError: table pisa_dependencies has no column named GENIEWeight
```

PR #383, opened 2023-01-12
Dataconverter fix

### LINKED ISSUE #382: torch imports in graphnet/data
**Describe the bug**
in `graphnet/data/__init__.py` we import torch and set the multiprocessing sharing strategy since [this](https://github.com/graphnet-team/graphnet/commit/8a41924c1cc1733a61a6923dcc8261a8172b4250) commit.  This means that installations for file conversions _without_ torch fails on the import statement.

**To Reproduce**
1. install graphnet without torch for file conversion
2. go to main branch
3. run i3-to-sqlite example

**Expected behavior**
We must make sure that data conversion functionality is independent of torch.

**Full traceback**
```
  File "make_database.py", line 7, in <module>
    from graphnet.data.extractors import (
  File "/lustre/hpc/hep/pcs557/github/graphnet/src/graphnet/data/__init__.py", line 8, in <module>
    import torch
ModuleNotFoundError: No module named 'torch'
```

RasmusOrsoe:
Closes #382 #381 #380 

Also modifies the `test_dataconverters` unit test, such that it now attempts to extract a pulsemap that is not in the i3 files if `backend == 'sqlite'`.


RasmusOrsoe:
Hey @asogaard - any pointers on how to satisfy the codeclimate issues?

`_extract_column_names` seems to fail the complexity test, even though this PR didn't change it?

MortenHolmRep:
> 

I believe you could extract the database call to reduce the complexity to the allowed limit.
Something like:

```
def _extract_column_names(self, db_paths: List[str], table_name: str) -> List[str]:
    for db_path in db_paths:
        tables_in_database = self._get_tables_in_database(db_path)
        if table_name in tables_in_database:
            con = _connect_to_db(db_path)
            columns = _query_column_names(con, table_name)
            if len(columns):
                return columns
    return []
```

Where the two new calls are
```
def _connect_to_db(db_path: str) -> sqlite3.Connection:
    con = sqlite3.connect(db_path)
    return con
```

```
def _query_column_names(con: sqlite3.Connection, table_name: str) -> List[str]:
    query = f"select * from {table_name} limit 1"
    columns = pd.read_sql(query, con).columns
    return columns
```

I haven't tested it though.

asogaard (approved):
Thanks for the error reports in #380, #381, and #382. It's good to have you back on `main`! :D  I agree with all of the diagnoses, and the PR looks excellent. I took the liberty of deleting a stray `.out` file and I have approved the two "cognitive complexity" issues in CodeClimate.

PR #389, opened 2023-01-25
Working base

Aske-Rosted:
Addition of several features used with neutrino-generator-21220 data, which seems to differ slightly compared to previously used data sets. 

added functionality includes

 - Skipping frame in I3 file in case of missing pulsemap (In the files I have only every other physics frame seems to be relevant) 
 - A warning message for when a frame is skipped. 
 - Added option for when the MCTree is named differently in my case "I3MCTree_preMuonProp" standard option still "I3MCTree"
 - Implementation of a function which renames detector features given a list of strings.  (This one is throwing a pre-commit error that I could not handle on my own)
 -  allows for passing name of features to perform coarsening on. Automatically finds the "time" feature assuming time is in the name of the feature.
 - adds a function which returns a random fraction of data without any specific selection criteria
 - adds functionality for loading data with with missing input values by forcing float (Warning message is printed. This might be a bit "hacky")
 - add an extractor for pulling zenith and azimuth comparisons from other algorithms.  

Sorry about this being a bit a bunched up collection of added functionality.

### DIFF
--- a/src/graphnet/data/dataconverter.py
+++ b/src/graphnet/data/dataconverter.py
@@ -430,7 +430,7 @@ class DataConverter(ABC, LoggerMixin):
             except:  # noqa: E722
                 continue
 
-            # Try to extract data from I3Frame else skip frame 
+            # Try to extract data from I3Frame else skip frame
             try:
                 results = self._extractors(frame)
             except KeyError as e:
--- a/src/graphnet/data/dataconverter.py
+++ b/src/graphnet/data/dataconverter.py
@@ -430,8 +430,16 @@ class DataConverter(ABC, LoggerMixin):
             except:  # noqa: E722
                 continue
 
-            # Extract data from I3Frame
-            results = self._extractors(frame)
+            # Try to extract data from I3Frame else skip frame 
+            try:
+                results = self._extractors(frame)
+            except KeyError as e:
+                if "Pulsemap" in str(e):
+                    self.warning(str(e) + ". Skipping frame")
+                    continue
+                else:
+                    raise e
+
             data_dict = OrderedDict(zip(self._table_names, results))
 
             # If an I3GenericExtractor is used, we want each automatically
--- a/src/graphnet/data/dataset.py
+++ b/src/graphnet/data/dataset.py
@@ -553,7 +553,15 @@ class Dataset(torch.utils.data.Dataset, Configurable, LoggerMixin, ABC):
             data = np.array([]).reshape((0, len(self._features) - 1))
 
         # Construct graph data object
-        x = torch.tensor(data, dtype=self._dtype)  # pylint: disable=C0103
+        try:
+            x = torch.tensor(data, dtype=self._dtype)  # pylint: disable=C0103
+        except TypeError as e:
+            x = torch.tensor(np.array(data, dtype=float), dtype=self._dtype)
+            self.warning(
+                r"The following type error was raised, forcing dtype float32. NaN's might occur in data! \n"
+                + str(e)
+            )
+
         n_pulses = torch.tensor(len(x), dtype=torch.int32)
         graph = Data(x=x, edge_index=None)
         graph.n_pulses = n_pulses
--- a/src/graphnet/data/extractors/__init__.py
+++ b/src/graphnet/data/extractors/__init__.py
@@ -10,7 +10,7 @@ from .i3featureextractor import (
 )
 from .i3truthextractor import I3TruthExtractor
 from .i3retroextractor import I3RetroExtractor
-from .i3splinempeextractor import I3SplineMPEICExtractor
+from .i3splinempeextractor import I3SplineMPEICExtractor, I3ComparisonExtractor
 from .i3tumextractor import I3TUMExtractor
 from .i3hybridrecoextractor import I3GalacticPlaneHybridRecoExtractor
 from .i3genericextractor import I3GenericExtractor
--- a/src/graphnet/data/extractors/i3splinempeextractor.py
+++ b/src/graphnet/data/extractors/i3splinempeextractor.py
@@ -1,6 +1,6 @@
 """I3Extractor class(es) for extracting SplineMPE reconstruction."""
 
-from typing import TYPE_CHECKING, Dict
+from typing import TYPE_CHECKING, Dict, List
 
 from graphnet.data.extractors.i3extractor import I3Extractor
 
@@ -24,6 +24,28 @@ class I3SplineMPEICExtractor(I3Extractor):
                 {
                     "zenith_spline_mpe_ic": frame["SplineMPEIC"].dir.zenith,
                     "azimuth_spline_mpe_ic": frame["SplineMPEIC"].dir.azimuth,
+                }
+            )
+
+        return output
+
+
+class I3ComparisonExtractor(I3Extractor):
+    """Class for extracting pointing predictions from various algorithms."""
+
+    def __init__(self, name: str):
+        """Construct I3ComparisonExtractor."""
+        # Base class constructor
+        super().__init__(name)
+
+    def __call__(self, frame: "icetray.I3Frame") -> Dict[str, float]:
+        """Extract pointing predictions."""
+        output = {}
+        if self._name in frame:
+            output.update(
+                {
+                    "zenith_" + self._name: frame[self._name].dir.zenith,
+                    "azimuth_" + self._name: frame[self._name].dir.azimuth,
                 }
             )
 
--- a/src/graphnet/data/sqlite/sqlite_selection.py
+++ b/src/graphnet/data/sqlite/sqlite_selection.py
@@ -5,10 +5,36 @@ from typing import List, Tuple, Optional
 
 import numpy as np
 import pandas as pd
-
+import random
 from graphnet.utilities.logging import get_logger
 
 logger = get_logger()
+
+
+def get_random_fraction(
+    database: str, fraction: float, seed: int = 0
+) -> List[int]:
+    """Get a random unique fraction without any special selection criteria.
+
+    Args:
+        database: Path to database from which to get event numbers.
+        fraction: desired fraction of the total number of events.
+        seed: Random number generator seed.
+
+    Returns:
+        List of event numbers.
+    """
+    rng = np.random.RandomState(seed=seed)
+    with sqlite3.connect(database) as con:
+        total_query = "SELECT event_no FROM truth"
+
+        tmp_dataframe = pd.read_sql(total_query, con)
+    event_no_list = (
+        tmp_dataframe.sample(frac=fraction, replace=False, random_state=rng)
+        .values.ravel()
+        .tolist()
+    )
+    return event_no_list
 
 
 def get_desired_event_numbers(
--- a/src/graphnet/models/coarsening.py
+++ b/src/graphnet/models/coarsening.py
@@ -219,20 +219,26 @@ class AttributeCoarsening(Coarsening):
 
 class DOMCoarsening(Coarsening):
     """Coarsen pulses to DOM-level."""
+
     def __init__(
         self,
         reduce: str = "avg",
         transfer_attributes: bool = True,
-        keys: Optional[List[str]] = ["dom_x", "dom_y", "dom_z", "rde", "pmt_area"]
+        keys: List[str] = [
+            "dom_x",
+            "dom_y",
+            "dom_z",
+            "rde",
+            "pmt_area",
+        ],
     ):
+        """Cluster pulses on the same DOM."""
         super().__init__(reduce, transfer_attributes)
         self._keys = keys
 
     def _perform_clustering(self, data: Union[Data, Batch]) -> LongTensor:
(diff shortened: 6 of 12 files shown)


src/graphnet/data/dataset.py
```diff
@@ -542,7 +542,15 @@ def _create_graph(
...
-        x = torch.tensor(data, dtype=self._dtype)  # pylint: disable=C0103
+        try:
+            x = torch.tensor(data, dtype=self._dtype)  # pylint: disable=C0103
+        except TypeError as e:
+            x = torch.tensor(np.array(data, dtype=float), dtype=self._dtype)
+            self.warning(
+                r"The following type error was raised, forcing dtype float32. NaN's might occur in data! \n"
+                + str(e)
+            )
+
```
asogaard, line 553:
```suggestion
        x = torch.tensor(data, dtype=self._dtype) # pylint: disable=C0103
```

src/graphnet/data/extractors/i3splinempeextractor.py
```diff
@@ -28,3 +28,25 @@ def __call__(self, frame: "icetray.I3Frame") -> Dict[str, float]:
             )
 
         return output
+
+
+class I3ComparisonExtractor(I3Extractor):
```
asogaard, line 33: Consider moving to its own file, e.g.,  `src/data/extractors/i3comparisonextractor.py`.

src/graphnet/data/extractors/i3splinempeextractor.py
```diff
@@ -1,6 +1,6 @@
 """I3Extractor class(es) for extracting SplineMPE reconstruction."""
 
-from typing import TYPE_CHECKING, Dict
+from typing import TYPE_CHECKING, Dict, List
```
asogaard, line 3: Not used (?).
```suggestion
from typing import TYPE_CHECKING, Dict
```

src/graphnet/data/extractors/__init__.py
```diff
@@ -10,7 +10,7 @@
 )
 from .i3truthextractor import I3TruthExtractor
 from .i3retroextractor import I3RetroExtractor
-from .i3splinempeextractor import I3SplineMPEICExtractor
+from .i3splinempeextractor import I3SplineMPEICExtractor, I3ComparisonExtractor
```
asogaard, line 13: Consider moving to its own file.

src/graphnet/data/sqlite/sqlite_selection.py
```diff
@@ -5,12 +5,38 @@
...
+
+        tmp_dataframe = pd.read_sql(total_query, con)
+    event_no_list = (
+        tmp_dataframe.sample(frac=fraction, replace=False, random_state=rng)
+        .values.ravel()
+        .tolist()
+    )
+    return event_no_list
+
+
```
asogaard, line 39: Consider whether this need can be covered by the DatasetConfig YAML-files (e.g., `1000 random events ~ (...)`).
```suggestion
```

src/graphnet/models/coarsening.py
```diff
@@ -220,11 +220,25 @@ def _perform_clustering(self, data: Union[Data, Batch]) -> LongTensor:
...
+            "dom_y",
+            "dom_z",
+            "rde",
+            "pmt_area",
+        ],
+    ):
+        """Cluster pulses on the same DOM."""
+        super().__init__(reduce, transfer_attributes)
+        self._keys = keys
+
```
asogaard, line 238:
```suggestion
        keys: Optional[List[str]] = None,
    ):
        """Cluster pulses on the same DOM."""
        super().__init__(reduce, transfer_attributes)
        if key is None:
            self._keys = [
                "dom_x",
                "dom_y",
                "dom_z",
                "rde",
                "pmt_area",
            ]
        else:
            self._keys = keys

```

asogaard (changes requested):
Thanks @Aske-Rosted!

src/graphnet/data/extractors/i3splinempeextractor.py
Aske-Rosted: done

src/graphnet/data/extractors/__init__.py
Aske-Rosted: done.

Aske-Rosted:
I have merged with newest version on main an implemented discussed alterations. IceTray check fails, but I am not sure if that is due to any changes that I have made.

asogaard:
> I have merged with newest version on main an implemented discussed alterations. IceTray check fails, but I am not sure if that is due to any changes that I have made.

I think this is due, at least partly, to the proposed change to how missing pulsemaps are handled, see [here](https://github.com/graphnet-team/graphnet/actions/runs/4180155413/jobs/7240864074#step:7:108) and [here](https://github.com/graphnet-team/graphnet/actions/runs/4180155413/jobs/7240864074#step:7:141), i.e., the proposed change around L433 in `src/graphnet/data/dataconverter.py`. Currently, a missing pulsemap means that an empty dictionary is returned and the processing continues unaffected. With the proposed change, _any_ missing pulsemap means that the entire frame is skipped. So I think the unit test is doing a great job of catching an unintended and potentially problematic behaviour here.

@Aske-Rosted, as I understood it from our chat, this proposed change is meant to filter out the "empty" (`NullSplit`) P-frames that accompany "genuine" (`InIceSplit` or whatever) P-frames. I think the best solution to this problem, that doesn't lead to unintended consequences as above, might be to do what @Andreas-MJ is currently working, which is to filter out these `NullSplit` P-frames directly. @RasmusOrsoe, do you think that would be a kosher way to go about things?

In any case, I think it might be worthwhile to open this as a separate issue, and then discuss there the best solution.

RasmusOrsoe:
@Aske-Rosted  @asogaard I agree that the unit test is doing it's job :-) 

Regarding the check - the most elegant solution I can think of is to add a line just after [this](https://github.com/graphnet-team/graphnet/blob/81fce2e7d9558ddabcea732fa5fb05837cd4c60c/src/graphnet/data/dataconverter.py#L429) where we could assert `assert is_null_split_frame(frame) == False`

So that block of code would become
```
while i3_file_io.more():
            try:
                assert is_null_split_frame(frame) == False
            except:  # noqa: E722
                continue 
```

Also, @MortenHolmRep I faintly recall that you wanted to filter away these nullsplitframes for the moon pointing project and that you added some code to main for this. Am I misremembering?

src/graphnet/data/extractors/i3comparisonextractor.py
```diff
@@ -0,0 +1,30 @@
...
+        super().__init__(name)
+
+    def __call__(self, frame: "icetray.I3Frame") -> Dict[str, float]:
+        """Extract pointing predictions."""
+        output = {}
+        if self._name in frame:
+            output.update(
+                {
+                    "zenith_" + self._name: frame[self._name].dir.zenith,
+                    "azimuth_" + self._name: frame[self._name].dir.azimuth,
```
asogaard, line 26: Consider adding other properties of I3Particles, e.g., energy, type, and position (might all be NaNs for most algorithms, but could still be useful to be able to extract these).

src/graphnet/data/extractors/i3splinempeextractor.py
```diff
@@ -28,3 +28,25 @@ def __call__(self, frame: "icetray.I3Frame") -> Dict[str, float]:
...
+        output = {}
+        if self._name in frame:
+            output.update(
+                {
+                    "zenith_" + self._name: frame[self._name].dir.zenith,
+                    "azimuth_" + self._name: frame[self._name].dir.azimuth,
+                }
+            )
+
+        return output
```
asogaard, line 52: Remove this, as the class has been moved to another module.
```suggestion
```

src/graphnet/data/sqlite/sqlite_selection.py
```diff
@@ -5,7 +5,7 @@
 
 import numpy as np
 import pandas as pd
-
+import random
```
asogaard, line 8: Not used?
```suggestion
```

src/graphnet/models/coarsening.py
```diff
@@ -271,23 +288,31 @@ def __init__(
         time_window: float,
         reduce: str = "avg",
         transfer_attributes: bool = True,
+        keys: List[str] = [
+            "dom_x",
+            "dom_y",
+            "dom_z",
+            "rde",
+            "pmt_area",
+        ],
```
asogaard, line 297:
```suggestion
        ],
        time_key: str = "dom_time",
```

src/graphnet/models/coarsening.py
```diff
@@ -271,23 +288,31 @@ def __init__(
...
     ):
         """Cluster pulses on the same DOM within `time_window`."""
         super().__init__(reduce, transfer_attributes)
         self._time_window = time_window
         self._cluster_method = DBSCAN(self._time_window, min_samples=1)
+        self._keys = keys
```
asogaard, line 303:
```suggestion
        self._keys = keys
        self._time_key = time_key
```

src/graphnet/models/coarsening.py
```diff
@@ -271,23 +288,31 @@ def __init__(
...
+        dom_index = group_by(data, self._keys)
         if data.batch is not None:
             features = data.features[0]
         else:
             features = data.features
 
-        ix_time = features.index("dom_time")
+        ix_time = [idx for idx, s in enumerate(features) if "time" in s][
+            0
+        ]  # features.index("dom_time")
```
asogaard, line 315:
```suggestion
        ix_time = features.index(self._time_key)
```

src/graphnet/data/dataconverter.py
```diff
@@ -430,8 +430,16 @@ def _extract_data(self, fileset: FileSet) -> List[OrderedDict]:
             except:  # noqa: E722
                 continue
 
-            # Extract data from I3Frame
-            results = self._extractors(frame)
+            # Try to extract data from I3Frame else skip frame
```
asogaard, line 433: I figure we need to remove or change this, see comment in main thread on this PR.

asogaard (changes requested):

MortenHolmRep:
> Also, @MortenHolmRep I faintly recall that you wanted to filter away these nullsplitframes for the moon pointing project and that you added some code to main for this. Am I misremembering?

It has been a few months, but as I recall it would have been the next step if we had continued on the project, but it was not implemented.

Aske-Rosted:
> 
Hmm, the "unintended behavior" that you are describing is precisely the behavior I am looking for. Is there any scenario where you need information from a P-frame where one (or more) of your pulse maps is missing?

Aske-Rosted:
I should now have addressed all the suggestions and corrections. I have implemented the handling of NullSplit frames in the way suggsted by @RasmusOrsoe as well as added some extra extraction features for the I3ParticleExtractor.

asogaard:
> > 
> 
> Hmm, the "unintended behavior" that you are describing is precisely the behavior I am looking for. Is there any scenario where you need information from a P-frame where one (or more) of your pulse maps is missing?

Yes, see "Expected behaviour" [here](https://github.com/graphnet-team/graphnet/issues/381#issue-1528917359). :)

src/graphnet/data/extractors/i3particleextractor.py
```diff
@@ -0,0 +1,43 @@
+"""I3Extractor class(es) for extracting SplineMPE reconstruction."""
```
asogaard, line 1:
```suggestion
"""I3Extractor class(es) for extracting I3Particle properties."""
```

src/graphnet/data/dataconverter.py
```diff
@@ -427,11 +427,13 @@ def _extract_data(self, fileset: FileSet) -> List[OrderedDict]:
         while i3_file_io.more():
             try:
                 frame = i3_file_io.pop_physics()
+                assert self._is_not_null_split_frame(frame)
```
asogaard, line 430: As I mention [here](https://github.com/graphnet-team/graphnet/pull/432#issuecomment-1433574619), I am not particularly fond of this solution, and would propose (1) amore general frame-skipping method and (2) separating this logic from whether the frame is a P-frame. But the proposed code works and is not unreasonable, so I'll leave it at your discretion, @Aske-Rosted.

asogaard (approved):
Hi @Aske-Rosted,

Thanks for this nice PR. I am approving it now, but before merging I would like you to have a look at:

* One suggestion to update a docstring.
* One comment regarding the way NullSplit frames are skipped.

I'll leave it to you whether and how to resolve these. :)

src/graphnet/data/dataconverter.py
RasmusOrsoe: @Aske-Rosted I think Andreas suggestion is better than what I mentioned earlier.

Aske-Rosted: Is the latest commit satisfactory? I am not 100% certain that all exceptions coming from the I3 modules will contain I3 in the exception name, but I would hope that they do...

asogaard:
All good, @Aske-Rosted! :)

PR #390, opened 2023-01-26
Main kaggle comp

RasmusOrsoe:
A bit of polishing required to make a graphnet example on the kaggle dataset.


RasmusOrsoe:
@asogaard Where did all the checks go?

src/graphnet/data/dataset.py
```diff
@@ -598,36 +598,51 @@ def _create_graph(
 
         # Additionally add original features as (static) attributes
         for index, feature in enumerate(graph.features):
-            graph[feature] = graph.x[:, index].detach()
```
asogaard, line 601: Ahh, this is because the Kaggle-dataset feature `x` conflicts with the torch-geometric built-in feature `x` referring to the features tensor?

src/graphnet/data/dataset.py
```diff
@@ -598,36 +598,51 @@ def _create_graph(
...
+            if feature not in ["x"]:
+                graph[feature] = graph.x[:, index].detach()
 
         # Add custom labels to the graph
         for key, fn in self._label_fns.items():
             graph[key] = fn(graph)
-
         return graph
 
     def _get_labels(self, truth_dict: Dict[str, Any]) -> Dict[str, Any]:
```
asogaard, line 609: We really ought to reconsider this function, and same for the double-bang one...

src/graphnet/models/detector/icecube.py
```diff
@@ -43,6 +43,34 @@ def _forward(self, data: Data) -> Data:
         return data
 
 
+class IceCube86Kaggle(Detector):
```
asogaard, line 46: Does the Kaggle data represent the main array only or also DeepCore?

asogaard (approved):
Looks really good! Thanks, @RasmusOrsoe.

src/graphnet/data/dataset.py
RasmusOrsoe: Yes.

src/graphnet/models/detector/icecube.py
RasmusOrsoe: Both!

src/graphnet/data/dataset.py
RasmusOrsoe: Yes, I agree. Maybe in the future, we could make a collection of `Labels` that the `Dataset` just defaults to.

src/graphnet/models/detector/icecube.py
asogaard: Then I figure it should be named `IceCubeDeepCoreKaggle` or just `IceCubeKaggle`.

RasmusOrsoe: It has been renamed to `IceCubeKaggle`

PR #398, opened 2023-01-27
added ability to overwrite distribution strategy

RasmusOrsoe:
One can now overwrite the default "ddp" strategy in trainer.fit()


MortenHolmRep (approved):
Looks good.

PR #409, opened 2023-01-31
Updating documentation webpage

### LINKED ISSUE #400: Polish Documentation landing Page
The landing page is very sparse. A bit of graphics and a hello would be nice.

Webpage: https://graphnet-team.github.io/graphnet/index.html

**Additional context**
We could seek inspiration from https://pytorch.org/docs/stable/index.html or other larger libraries.

asogaard:
Updates the auto-generated documentation webpage according to the feedback here https://github.com/openjournals/joss-reviews/issues/4971#issuecomment-1408985970

You can see the result here: https://asogaard.github.io/graphnet/

**To-do:**
- [x] Revert `docs.yml` GitHub Action to only act on main branch before merging.

Closes #400


asogaard:
Hi @RasmusOrsoe and @JostMigenda,

I think I have made considerable progress on the auto-generated website based on the feedback here: https://github.com/openjournals/joss-reviews/issues/4971#issuecomment-1408985970 and here: #400.

I might continue tinker a little bit under the hood with this, but I would appreciate if you could have a look at the current result here: https://asogaard.github.io/graphnet/ and let me know whether you think it does the job.

RasmusOrsoe (approved):
Really cool!!

JostMigenda:
Very nice work, thank you!

A few rough edges I’m still noticing:
* [ ] Docs for the class and its `__init__` function duplicate the arguments, see e.g. [`graphnet.data.sqlite.sqlite_dataset_perturbed`](https://asogaard.github.io/graphnet/api/graphnet.data.sqlite.html#module-graphnet.data.sqlite.sqlite_dataset_perturbed).
* [ ] A lot of content from the README and CONTRIBUTING files in the repo root is now duplicated e.g. in `docs/source/{install,contributing}.md`. That will cause additional work anytime you need to update something and they are likely to go out of sync over time. It would be better to have that in just one place and link there from the other.
* [ ] Submodule names are often cut off in the sidebars; which can be a bit awkward for deeply nested ones like `graphnet.models.detector.*` (see attached screenshot). I’m not sure if there’s an easy way to hide the common prefix there, as for subpackages.
* [ ] As noted in my earlier comment on the JOSS review, additional explanations would be useful in some module-level docstrings. (Though that is independent of the changes here and doesn’t need to hold up this PR.)

image

MortenHolmRep:
It looks really well made @asogaard ! I have two proposals.

1. It would be a good idea to divide the installation process into operating systems from the get-go instead of at the end step.
It replicates many of the sub-sections, but it also allows us to tailor the setup to the operating system and better give proposals for solutions to specific issues; it would look something like this:
-| Installing with Linux
--| Installing with IceTray
--| installing with stand
-| Installing with macOS
-| Installing with WSL

2. If it is at all possible, I would also make the suggestion of a PyTorch-like selection for the setup as seen [here](https://pytorch.org/get-started/locally/).
It removes unnecessary options for the user, and the current setup description, seen from a usability perspective, gives the user a set of instructions that some could follow blindly and copy-paste without looking at, e.g. CPU vs GPU setup etc.
With the PyTorch-like setup, we could have subjects like operating system, developer or user, CPU or GPU, etc.

asogaard:
Hi @JostMigenda,

I have done a bit more work on this to address the first three of your four points above. I completely agree with the fourth point, but will address this in a separate contribution. I might tinker a little bit e.g. with the formatting of a few docstrings, but in the meantime please let me know if you think the rough edges you've noticed have been adequately sanded down.

JostMigenda:
Thanks @asogaard!
2 and 3 look good now; for 1, it looks like the description of the args is missing now. E.g. for the `SQLiteDatasetPerturbed` class I used as an example above, I can see good explanations in the docstring of the `__init__` function—but the web page only contains type information.

asogaard:
Hi @JostMigenda,

Thanks for the keen eye! I have made updates that ought to do the job, see e.g. [here](https://asogaard.github.io/graphnet/api/graphnet.data.sqlite.sqlite_dataset_perturbed.html).

PR #416, opened 2023-02-01
I3 Module Refactoring

RasmusOrsoe:
This is a draft PR that refactors the i3 modules. The aim of the refactoring is to minimize the amount of IceTray-specific knowledge needed to write deployment code for current and future IceCube deployments. This is achieved by removing direct inheritance of IceTray `I3Module` python classes, which requires/suggests opaque structure on the code ( `def Physics()`, `self.AddParameter()`, etc.). We can get away with this because IceTray is compatible with native Python `Callables` (functions, classes with `__call__`) if used through `I3Tray.AddModule()`. 

**In deployment/graphnet_module.py**:

- `GraphNeTI3Module` is a general deployment class that contains general functionality; pulsemap extraction, graph creation and general structure.

- `I3InferenceModule` is a general inference module that will write predictions from a GNN to i3 frame. Single task `Model` is assumed (but this seems to be the case for it's predecessor too).

- `I3PulseCleanerModule` is a specialized deployment module that will write cleaned pulses to i3 frames. 

**In deployment/deployer.py**:

- `GraphNeTI3Deployer` is an deployment class that will apply modules in `graphnet_module.py` to i3 files via IceTray. Can run on multiple cores. 

If you agree with the refactoring, I'll add more doc strings and an example script.

Let me know what you think!


RasmusOrsoe:
@asogaard OK - so I think this is ready for your busy eyes.

mypy check fails because docker can't find a module that has been removed in this PR. Can you help me understand how I fix that?

Also; codeclimate is not a fan of my doc strings. How serious should I take this? I can remove them, but that would be a shame..

asogaard:
> @asogaard OK - so I think this is ready for your busy eyes.
> 
> mypy check fails because docker can't find a module that has been removed in this PR. Can you help me understand how I fix that?
> 
> Also; codeclimate is not a fan of my doc strings. How serious should I take this? I can remove them, but that would be a shame..

I'll try to have a look. :)  

* Re. mypy: If you're removing or renaming a class, I figure you should do that everywhere it's referenced. So that would mean either updating the benchmarking script to use the new `I3Module` class, or — if the the new class is completely different and it's not worth it to update the benchmarking script — I'd just remove the latter, as we're not really using it anyway.
* Re. codeclimate: I think it's good form to keep all lines within 79 characters, including docstrings. (That's the line lenght we're asking black to enforce, but the re-commit hooks we're using don't reformat docstrings, because they can risk breaking them. That's easier with code, which is what black takes care of.) In VS Code, you can add a vertical ruler at 79 characters as suggested [here](https://stackoverflow.com/a/29972073). Then it's pretty easy to see which lines are to long, and then just add newlines such that it fits neatly. That's what I do. :)

src/graphnet/deployment/i3modules/graphnet_module.py
```diff
@@ -24,126 +21,91 @@
...
+    def __call__(self, frame: I3Frame) -> bool:
+        """Define here how the module acts on the frame.
 
-        if isinstance(model, str):
-            self.model = StandardModel.load(model)
-        else:
-            self.model = model
+        Must return True if successful.
+        """
+        return True
```
asogaard, line 73: Don't think we should add code in an `abstractmethod`.
```suggestion
```

src/graphnet/deployment/i3modules/graphnet_module.py
```diff
@@ -153,42 +115,299 @@ def _extract_feature_array_from_frame(self, frame: I3Frame) -> np.array:
                 )
         return features
 
-    def _write_predictions_to_frame(
-        self, frame: I3Frame, prediction: np.array
+    def _submit_to_frame(
```
asogaard, line 120: I figure "add" might me more intuitive?
```suggestion
    def _add_to_frame(
```

src/graphnet/deployment/i3modules/graphnet_module.py
```diff
@@ -153,42 +115,299 @@ def _extract_feature_array_from_frame(self, frame: I3Frame) -> np.array:
...
+        self, frame: I3Frame, data: Dict[str, Any]
     ) -> I3Frame:
-        nb_preds = prediction.shape[0]
-        if isinstance(self.keys, str):
-            if nb_preds > 1:
-                keys = [f"{self.keys}_{ix}" for ix in range(nb_preds)]
-            else:
-                keys = [self.keys]
+        """Submit every field in data to I3Frame.
```
asogaard, line 123: See above.
```suggestion
        """Add every field in `data` to `frame`.
```

src/graphnet/deployment/i3modules/graphnet_module.py
```diff
@@ -153,42 +115,299 @@ def _extract_feature_array_from_frame(self, frame: I3Frame) -> np.array:
...
+                data[
+                    self.model_name + "_" + self.prediction_columns[i]
+                ] = I3Double(float(predictions[:, i][0]))
+            except IndexError:
+                data[
+                    self.model_name + "_" + self.prediction_columns[i]
+                ] = I3Double(predictions[0])
+
+        # Submission methods
+        frame = self._submit_to_frame(frame=frame, data=data)
```
asogaard, line 230: `_add_to_frame`?

examples/01_icetray/03_deployment_to_i3.py
```diff
@@ -0,0 +1,127 @@
...
+    deployment_modules = construct_modules(
+        model_dict=model_dict, gcd_file=gcd_file
+    )
+
+    # Construct I3 deployer
+    deployer = GraphNeTI3Deployer(
+        graphnet_modules=deployment_modules,
+        n_workers=1,
+        gcd_file=gcd_file,
+    )
```
asogaard, line 117: I think it could be beneficial to have two examples: One like this, that uses a `Deployer` class to parallelise I3 file processing; and one that just uses a simple I3Tray (the 20-or-so lines in the body of `GraphNeTI3Deployer._process_files`), i.e., what most people in IceCube are probably familier with.

src/graphnet/deployment/i3modules/deployer.py
```diff
@@ -0,0 +1,204 @@
...
+            List[GraphNeTI3Module],
+            List[I3InferenceModule],
+            List[I3PulseCleanerModule],
+            List[
+                Union[
+                    GraphNeTI3Module,
+                    I3InferenceModule,
+                    I3PulseCleanerModule,
+                ]
+            ],
```
asogaard, line 46: Since `I3InferenceModule` and `I3PulseCleanerModule` both inherit from `GraphNeTI3Module` they will both be instances of the latter, and it should be sufficient to specify the base class.
```suggestion
            GraphNeTI3Module,
            List[GraphNeTI3Module],            
```

src/graphnet/deployment/i3modules/deployer.py
```diff
@@ -0,0 +1,204 @@
...
+                    ]
+                )
+        else:
+            settings = [
+                input_files,  # type: ignore
+                self._gcd_file,  # type: ignore
+                output_folder,  # type: ignore
+                self._modules,  # type: ignore
+            ]  # type: ignore
+        return settings
```
asogaard: I have always had a hard time with settings-as-lists, because they require you to keep track of what property a certain list index corresponds to. Could I suggest instead adding a small class at the top of this file,
```python
@dataclass
class Settings:
    i3_files: List[str]
    gcd_file: str
    output_folder: str
    modules: List[GraphNeTI3Module]
```
Then you can do stuff like `settings.i3_files` rather than `settings[ix]` or unwrapping like `(i3_files, gcd_file, ...) = settings` which is probably less maintainable.

Then you can use this class in this call as suggested below:

```suggestion
    def _prepare_settings(
        self, input_files: List[str], output_folder: str
    ) -> List[Settings]:
        """Will prepare the settings for each worker."""
        try:
            os.makedirs(output_folder)
        except FileExistsError:
            assert (
                False
            ), f"{output_folder} already exists. To avoid overwriting existing files, the process has been stopped."

        if self._n_workers > len(input_files):
            self._n_workers = len(input_files)

        if self._n_workers > 1:
            file_batches = np.array_split(input_files, self._n_workers)
            settings: List[Settings] = []
            for i in range(self._n_workers):
                settings.append(
                    Settings(
                        file_batches[i],
                        self._gcd_file,
                        output_folder,
                        self._modules,
                    )
                )
        else:
            settings = [Settings(
                input_files,
                self._gcd_file,
                output_folder,
                self._modules,
            )]

        return settings
```

src/graphnet/deployment/i3modules/deployer.py
```diff
@@ -0,0 +1,204 @@
...
+                    icetray.I3Frame.DAQ,
+                    icetray.I3Frame.Physics,
+                    icetray.I3Frame.TrayInfo,
+                    icetray.I3Frame.Simulation,
+                ],
+                filename=output_folder + "/" + INFILE.split("/")[-1],
+            )
+            tray.Execute()
+            tray.Finish()
+        return
```
asogaard:
```suggestion
    def _process_files(
        self,
        settings: Settings,
    ) -> None:
        """Will start an IceTray read/write chain with graphnet modules.

        If n_workers > 1, this function is run in parallel n_worker times. Each worker will loop over an allocated set of i3 files. The i3 files with reconstructions will appear as copies of the original i3 files but with reconstructions added. Original i3 files are left untouched.
        """
        for i3_file in settings.i3_files:
            tray = I3Tray()
            tray.context["I3FileStager"] = dataio.get_stagers()
            tray.AddModule(
                "I3Reader",
                "reader",
                FilenameList=[settings.gcd_file, i3_file],
            )

            for i3_module in settings.modules:
                tray.AddModule(i3_module)

            tray.Add(
                "I3Writer",
                Streams=[
                    icetray.I3Frame.DAQ,
                    icetray.I3Frame.Physics,
                    icetray.I3Frame.TrayInfo,
                    icetray.I3Frame.Simulation,
                ],
                filename=settings.output_folder + "/" + i3_file.split("/")[-1],
            )
            tray.Execute()
            tray.Finish()
        return
```

src/graphnet/deployment/i3modules/deployer.py
```diff
@@ -0,0 +1,204 @@
...
+        Args:
+            input_files: Path(s) to i3 file(s) that you wish to apply the graphnet modules to.
+            output_folder: The output folder to which the i3 files are written.
+        """
+        start_time = time.time()
+        if isinstance(input_files, list):
+            random.shuffle(input_files)
+        else:
+            input_files = [input_files]
+        settings = self._setup_settings(
```
asogaard, line 207:
```suggestion
        settings = self._prepare_settings(
```

src/graphnet/deployment/i3modules/deployer.py
```diff
@@ -0,0 +1,204 @@
...
+
+            for process in processes:
+                process.join()
+        else:
+            self._process_files(
+                files=settings[0],
+                gcd_file=settings[1],
+                output_folder=settings[2],
+                modules=settings[3],
+            )  # type: ignore
```
asogaard, line 132:
```suggestion
    def _launch_jobs(self, settings: List[Settings]) -> None:
        """Will launch jobs in parallel if n_workers > 1, else run on main."""
        if self._n_workers > 1:
            assert len(settings) == self._n_workers, (
                f"Then number of settings ({len(settings)}) and workers "
                f"({self._n_workers}) don't agree."
            )
            processes = []
            for i in range(self._n_workers):
                processes.append(
                    multiprocessing.Process(
                        target=self._process_files, args=(settings[i],)
                    )
                )

            for process in processes:
                process.start()

            for process in processes:
                process.join()
        else:
            assert len(settings) == 1
            self._process_files(settings[0])
```

asogaard (commented):
Hi @RasmusOrsoe,

Thanks a lot for the, and massive kudos on the rewrite of the `I3Module` class(es). Looks really neat! I have mainly added some suggestions for the deployer-type class, to make it a little more legible.

RasmusOrsoe:
@asogaard I think that's it!

examples/01_icetray/04_i3_module_in_native_icetray_example.py
```diff
@@ -0,0 +1,167 @@
...
+    )
+    model = StandardModel(
+        detector=detector,
+        gnn=gnn,
+        tasks=[task],
+        optimizer_class=Adam,
+        optimizer_kwargs={"lr": 1e-03, "eps": 1e-03},
+    )
+    return model
+
```
asogaard, line 70: See comment above.

examples/01_icetray/04_i3_module_in_native_icetray_example.py
```diff
@@ -0,0 +1,167 @@
...
+                pulsemap_extractor=extractor,
+                model=model_path,
+                gcd_file=gcd_file,
+                prediction_columns=prediction_columns,
+                model_name=model_name,
+            )
+        )
+    return deployment_modules
+
+
```
asogaard, line 96: Same as above.
```suggestion
```

examples/01_icetray/04_i3_module_in_native_icetray_example.py
```diff
@@ -0,0 +1,167 @@
...
+    model_dict = {}
+    model_dict["graphnet_dynedge_energy_reconstruction"] = {
+        "model_path": construct_mock_model(),
+        "prediction_columns": ["energy_pred"],
+        "pulsemap": pulsemap,
+    }
+
+    deployment_modules = construct_modules(
+        model_dict=model_dict, gcd_file=gcd_file
+    )
```
asogaard, line 156:
```suggestion
```suggestion
    # Construct I3InferenceModule(s)
    extractor = I3FeatureExtractorIceCubeUpgrade(pulsemap=pulsemap)
    deployment_modules = [
        I3InferenceModule(
            pulsemap=pulsemap,
            features=features,
            pulsemap_extractor=extractor,
            model=construct_mock_model(),
            gcd_file=gcd_file,
            prediction_columns=["energy_pred"],
            model_name="graphnet_dynedge_energy_reconstruction",
        ),
        # You can add more module here.
    ]
```
```

File: examples/01_icetray/04_i3_module_in_native_icetray_example.py [hunk 14]
```diff
@@ -0,0 +1,167 @@
...
+                icetray.I3Frame.DAQ,
+                icetray.I3Frame.Physics,
+                icetray.I3Frame.TrayInfo,
+                icetray.I3Frame.Simulation,
+            ],
+            filename=output_folder + "/" + i3_file.split("/")[-1],
+        )
+        tray.Execute()
+        tray.Finish()
+    return
```
asogaard, line 132: Don't think it's necessary to wrap this in a function.
```suggestion
```

examples/01_icetray/04_i3_module_in_native_icetray_example.py
```diff
@@ -0,0 +1,167 @@
...
+    deployment_modules = construct_modules(
+        model_dict=model_dict, gcd_file=gcd_file
+    )
+
+    apply_to_files(
+        i3_files=input_files,
+        gcd_file=gcd_file,
+        output_folder=output_folder,
+        modules=deployment_modules,
+    )
```
asogaard, line 163:
```suggestion
    # Apply deployment modules to I3-files.
    for i3_file in input_files:
        tray = I3Tray()
        tray.context["I3FileStager"] = dataio.get_stagers()
        tray.AddModule(
            "I3Reader",
            "reader",
            FilenameList=[gcd_file, i3_file],
        )
        for i3_module in deployment_modules:
            tray.AddModule(i3_module)
        tray.Add(
            "I3Writer",
            Streams=[
                icetray.I3Frame.DAQ,
                icetray.I3Frame.Physics,
                icetray.I3Frame.TrayInfo,
                icetray.I3Frame.Simulation,
            ],
            filename=output_folder + "/" + i3_file.split("/")[-1],
        )
        tray.Execute()
        tray.Finish()
```

asogaard (commented):
Hi @RasmusOrsoe,

I have had a look at the example scripts, and made some suggestions to simplify them, as it seems there is a bit of duplicate code. But these are just suggestions, so let me know what you think. :)

examples/01_icetray/04_i3_module_in_native_icetray_example.py
RasmusOrsoe: I think it gives a nice break between graphnet-specfic things and icetray functionality. So unless you feel this strongly, I'd prefer to keep it this way.

RasmusOrsoe:
@asogaard  OK. I have added a mock model under `graphnet/data/tests/models` and simplified the example scripts. I've verified that they work.

examples/01_icetray/04_i3_module_in_native_icetray_example.py
asogaard: I'll leave that to you. :)  I'd at least update the docstring to remove the reference to `n_workers`. ;)

asogaard (approved):
I think this looks great. Thanks a lot, @RasmusOrsoe! 👏

PR #421, opened 2023-02-04
Add inactive sensors to graph

RasmusOrsoe:
This PR is a draft. It's a bit rough around the edges but outlines a method of adding inactive sensor modules to the graphs. 

The general idea is to add abstract (backend specific implementations needed) methods to the Dataset class that queries geometry_tables for sensor_id's. In SQLite I've created a method that constructs `geometry_table` (look-up table for geometry) from pulsemap tables. 

Right now, the `_get_inactive_sensors` will actually extract the active pulsemap on its own (and create the UID in the process). This means that every pulsemap will be extracted twice; a bit inefficient.

Let me know what you think of the general approach! I'll try to polish this up during the week.

### DIFF
--- a/src/graphnet/data/dataset.py
+++ b/src/graphnet/data/dataset.py
@@ -140,7 +140,7 @@ class Dataset(torch.utils.data.Dataset, Configurable, LoggerMixin, ABC):
         loss_weight_column: Optional[str] = None,
         loss_weight_default_value: Optional[float] = None,
         seed: Optional[int] = None,
-        add_inactive_sensors: bool = False,
+        include_inactive_sensors: bool = False,
         geometry_table: str = "geometry_table",
         sensor_x_position: str = "dom_x",
         sensor_y_position: str = "dom_y",
@@ -190,7 +190,7 @@ class Dataset(torch.utils.data.Dataset, Configurable, LoggerMixin, ABC):
                 subset of events when resolving a string-based selection (e.g.,
                 `"10000 random events ~ event_no % 5 > 0"` or `"20% random
                 events ~ event_no % 5 > 0"`).
-            add_inactive_sensors: If True, sensors that measured nothing during the event will be appended to the graph.
+            include_inactive_sensors: If True, sensors that measured nothing during the event will be appended to the graph.
             geometry_table: The sqlite table in which the detector geometry is stored.
             sensor_x_position: column name of x-coordinate of sensors. e.g. "dom_x",
             sensor_y_position: column name of y-coordinate of sensors. e.g. "dom_y",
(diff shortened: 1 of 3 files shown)


asogaard:
Hi @RasmusOrsoe,

Thanks for preparing this! I think this to a large extent was what was meant by #65.

I agree with the overall structure:
* Save a look-up table of unique sensors (PMTs; by position) to file.
* Query this look-up table every time you create a graph in a `Dataset` class, to "fill in" the missing sensors.

Some high-level considerations:
* Right now the geometry table is "baked in" to the SQLite database, which was used to produce it. Would it make sense to store geometry data in a generic (CSV?) format, so as to make it independent of the backend?
* Right now the inactive sensors are extracted in SQLiteDataset alone, using SQL-specific commands. Would it make sense to use the generic `Dataset` API to do this, e.g., something like
```python
class Dataset:
    (...)
    def get_detector_geometry(
        self, 
        detector_geometry_file: Optional[str] = None, 
        pulsemap: Optional[str] = None,
        coordinate_features: Optional[Sequence[str]] = None,
    ) -> pd.DataFrame:
        # Read from cache if it exists.
        if detector_geometry_file and os.path.exists(detector_geometry_file):
            return pd.read_csv(detector_geometry_file)

        # Construct detector geometry
        assert pulsemap is not None
        if coordinate_features is None:
            coordinate_features = self.features[:3]
            assert "_x" in coordinate_features[0].lower()
            assert "_y" in coordinate_features[1].lower()
            assert "_z" in coordinate_features[2].lower() 

        coordinates: List[Tuple] = self.query_table(pulsemap, coordinate_features)
        df_sensors = pd.DataFrame(
            data=np.unique(coordinates),
            columns=coordinate_features,
        )

        # Cache detector geometry
        if detector_geometry_file:
            df_sensors.to_csv(detector_geometry_file)

        return df_sensors
```
* Using floating-point numbers like the sensor coordinates to construct a unique identified (UID), by casting them to string, is inherently a little risky. We probably ought to do some rounding or fuzzy matching.
* Given `features` as a `List[Tuple]` I think the simplest, and most generic approach to combining the active and inactive sensors (as is being done on `SQLiteDataset._get_inactive_sensors`) would be to do something like
```python
class Dataset:
    (...)

    def _add_inactive_sensors(self, features: List[Tuple], padding_value: float = 0.) -> List[Tuple]:
        # Get DataFrames for active and inactive sensors
        df_features = pd.DataFrame(data=features, columns=self.features)
        df_inactive = self.detector_geometry  # constructed using self.get_detector_geometry(...)

        # Assuming a merge key (like UID) exist in both, make an outer join 
        # (i.e., a join that keeps an entry for every example of the "on"-key 
        # that exists in either. This should produce a DataFrame that contains 
        # at least one row for each UID. If the UID exists in `df_features`, 
        # those rows will be kept. If it doesn't it will be "taken" from 
        # `df_inactive`. The columns in the resulting DataFrame will be the 
        # union of columns in the two DataFrames being joined. Since we select 
        # only the merge key in `df_inactive`, the resulting DataFrame will have 
        # the same columns as `df_features. For rows corresponding to inactive 
        # sensors, all values except "UID" will then be `NaN`.
        df_features = df_features.merge(df_inactive[["UID"]], on="UID", how="outer")
        
        # Fill in missing values for inactive sensors.
        df_features = df_features.fillna(padding_value)

        return df_features.tolist()
```

RasmusOrsoe:
Superseded by #623

PR #426, opened 2023-02-07
Bash script for training on multiple models

MortenHolmRep:
Added a bash file that can train on multiple model configurations on the same dataset. In the example case using the Prometheus dataset on the three different models `example_direction_reconstruction_model.yml`, `example_energy_reconstruction_model.yml`, `example_vertex_position_reconstruction_model.yml`.

This enables a faster training pipeline on multiple models by decreasing the downtime between runs.
The bash file was tested on multiple datasets and models.

Let me know if anything needs changing.

### DIFF
--- a/examples/04_training/01_train_model.py
+++ b/examples/04_training/01_train_model.py
@@ -64,9 +64,7 @@ def main(
     )
 
     if suffix is not None:
-        archive = os.path.join(
-            EXAMPLE_OUTPUT_DIR, "train_model_{}".format("_".join(suffix))
-        )
+        archive = os.path.join(EXAMPLE_OUTPUT_DIR, f"train_model_{suffix}")
     else:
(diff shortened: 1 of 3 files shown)


examples/04_training/01_train_model.py
```diff
@@ -35,6 +35,7 @@ def main(
     batch_size: int,
     num_workers: int,
     prediction_names: Optional[List[str]],
+    suffix: Optional[str],
```
asogaard, line 38:
```suggestion
    suffix: Optional[str] = None,
```

asogaard (changes requested):
Hi @MortenHolmRep,

I think this looks great — thanks! 🎉  Especially for catching the typo in `configs/models/dynedge_position_example.yml` 🙈 

I have added a single suggestion; once that's done we can merge this.

asogaard (approved):

PR #427, opened 2023-02-10
docstring update

RasmusOrsoe:
Merged before I realized the docstring reference to n workers. This fixes that.


MortenHolmRep (approved):
The modification to the docstring makes sense, given the changes.

PR #430, opened 2023-02-14
EnsembleDataset

RasmusOrsoe:
This PR introduces a new `Dataset` called `EnsembleDataset`. It combines a list of smaller `Dataset`s into one large `Dataset`. 

You could use this to combine datasets from different sqlite databases, but is not back-end specific, so it works with parquet too.

@asogaard this means that the multiple-database-training functionality that still lives in `SQLiteDataset` can be removed!

It also comes with a little example.

### DIFF
--- a/examples/02_data/04_ensemble_dataset.py
+++ b/examples/02_data/04_ensemble_dataset.py
@@ -48,7 +48,7 @@ def main() -> None:
         truth_table=truth_table,
     )
 
-    ensemble_dataset = EnsembleDataset([dataset_1, dataset_2])
+    ensemble_dataset = EnsembleDataset(datasets=[dataset_1, dataset_2])
 
     logger.info(f"dataset_1 have length: {len(dataset_1)}")
     logger.info(f"dataset_2 have length: {len(dataset_2)}")
@@ -67,7 +67,8 @@ def main() -> None:
         for batch in tqdm(dataloader, unit=" batches", colour="green"):
             time.sleep(wait_time)
 
-    logger.info(batch["dataset_idx"])
+    for i in range(batch_size):
+        logger.info(f"Event {i} came from {batch['dataset_path'][i]}")
 
 
 if __name__ == "__main__":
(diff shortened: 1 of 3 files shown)


asogaard:
Hi @RasmusOrsoe,

Thanks for this! This is indeed very handy functionality. How does it compare to [pytorch's ConcatDataset](https://pytorch.org/docs/stable/data.html#torch.utils.data.ConcatDataset)? I have been using that [here](https://github.com/graphnet-team/graphnet/blob/main/src/graphnet/data/dataset.py#L63_L69) to concatenate multiple datasets, e.g., such that you can define a collective dataset made out of different component, like [here](https://github.com/graphnet-team/graphnet/blob/main/src/graphnet/utilities/config/dataset_config.py#L91-L109). Do you think this is a class we could leverage to our benefit, and then perhaps just "wrap" it in our own class, that we can then use wherever we're sticking things together?

But I _very_ much like the example, and the prospect of being able to remove some of the more specialised multi-database-bits from `SQLiteDataset`. ;)

RasmusOrsoe:
Hey @asogaard. Glad you like it! 

I did notice this [pytorch's ConcatDataset](https://pytorch.org/docs/stable/data.html#torch.utils.data.ConcatDataset) - the only reason I did not use this was because of the missing `dataset id's`. It's important that we know which dataset the event came from. It was not immidately clear to me how I could use that class & get the `dataset id`s . If you have an idea, I'd like to hear it!

asogaard:
Perhaps you could do something like
```python
graph["path"] = self.path
```
somewhere around [here](https://github.com/graphnet-team/graphnet/blob/main/src/graphnet/data/dataset.py#L596). Then each graph would contain a reference to the file (database or parquet) from which it originated. It wouldn't give you an index per se, but I guess you could then do something like
```python
paths = ["path/to/first.db", ...]
dataset = EnsembleDataset(paths)  #  or ConcatDataset(paths)
graph = dataset[0]
ix = paths.index(graph["path"])
```

RasmusOrsoe:
@asogaard  I've implemented your suggestion and updated the example. I like the result, but I have not quantified how much more memory is required to store the strings instead of idx. Perhaps for large batch sizes this is suboptimal? Anyway, it works.

asogaard (approved):
Looks good to me! :)

Issue #412, opened 2023-02-01
Snowstorm doesn't convert


Andreas-MJ:
**Describe the bug**
I am trying to convert I3 files from the Snowstorm sample to sql. The I3-files contain 2 P-frames in a row, where only one has the specified pulsemap I need. I get the following error message:

```
(graphnet) [andreasmj@hep04 ~]$ python /groups/icecube/andreasmj/work/graphnet/examples/01_icetray/01_convert_i3_files.py sqlite icecube-86
graphnet: INFO     2023-02-01 15:27:15 - get_logger - Writing log to logs/graphnet_20230201-152715.log
graphnet: INFO     2023-02-01 15:27:34 - SQLiteDataConverter.info - Saving results to /groups/icecube/petersen/GraphNetDatabaseRepository/Snowstorm/merged
graphnet: INFO     2023-02-01 15:27:34 - SQLiteDataConverter.info - Processing 1623 I3 file(s) in main thread (not multiprocessing)
  0%|                                                                                                                   | 0/1623 [00:00<?, ?file(s)/s]graphnet: WARNING  2023-02-01 15:27:35 - I3GenericExtractor.warning - Key SplitInIcePulses not in frame. Skipping
graphnet: WARNING  2023-02-01 15:27:35 - I3GenericExtractor.warning - Key LineFitParams not in frame. Skipping
graphnet: WARNING  2023-02-01 15:27:35 - warn_once - break_cyclic_recursion - Already evaluating object. Skipping recusion.
graphnet: WARNING  2023-02-01 15:27:35 - I3GenericExtractor.warning - Key LineFitParams not in frame. Skipping
graphnet: WARNING  2023-02-01 15:27:35 - I3GenericExtractor.warning - Key SplitInIcePulses not in frame. Skipping
graphnet: WARNING  2023-02-01 15:27:35 - I3GenericExtractor.warning - Key LineFitParams not in frame. Skipping
graphnet: WARNING  2023-02-01 15:27:36 - I3GenericExtractor.warning - Key LineFitParams not in frame. Skipping
graphnet: WARNING  2023-02-01 15:27:36 - I3GenericExtractor.warning - Key SplitInIcePulses not in frame. Skipping
graphnet: WARNING  2023-02-01 15:27:36 - I3GenericExtractor.warning - Key LineFitParams not in frame. Skipping
graphnet: WARNING  2023-02-01 15:27:37 - I3GenericExtractor.warning - Key LineFitParams not in frame. Skipping
[...]

PR #432, opened 2023-02-16
Skip NullSplit frames

Andreas-MJ:
Closes issue #412


asogaard:
@Andreas-MJ, thanks for this PR! A few comments:
* I forgot to say that the syntax in the description for automating issue-closing is "Closes #..." not "Closes issue #..." — sorry! (And one line of "Closes #..." per issue to be closed.) I have added the dependency in the "Development" section of the panel on the right.
* The [automated check for black](https://github.com/graphnet-team/graphnet/actions/runs/4194806139/jobs/7273457147) fails. I would suggest [adding pre-commit hooks](https://github.com/graphnet-team/graphnet/blob/81fce2e7d9558ddabcea732fa5fb05837cd4c60c/CONTRIBUTING.md#code-quality) to your local setup if you haven't already. That should take care of black automatically. Alternatively you can run `$ pre-commit run black src/graphnet/data/dataconverter.py` or `$ black --config black.toml src/graphnet/data/dataconverter.py` to manually enforce it. (NB: This is from memory! I might have gotten the commands wrong.)
* [CodeClimate](https://codeclimate.com/github/graphnet-team/graphnet/pull/432) has found four issues. The first three are related to line length, and you can see the discussion [here](https://github.com/graphnet-team/graphnet/pull/416#issuecomment-1425377612) about how to easily solve this going forward. The fourth one black will take care of once you run it.

Tagging @Aske-Rosted, @MortenHolmRep and @RasmusOrsoe: This could be a quite explicit solution tothe problem encountered by Aske, and Leon and Morten before him. Does anyone foresee this to be a problem? (I.e., is there any use case for `NullSplit` frames that we put at risk in this way?)

MortenHolmRep:
> Tagging @Aske-Rosted, @MortenHolmRep and @RasmusOrsoe: This could be a quite explicit solution tothe problem encountered by Aske, and Leon and Morten before him. Does anyone foresee this to be a problem? (I.e., is there any use case for `NullSplit` frames that we put at risk in this way?)

What a speedy fix to the issues being discussed!

NullSplit seems to be a thing that was introduced for IC79 and I have found the following related to NullSplit
[IC79 P-frames](https://wiki.icecube.wisc.edu/index.php/IC79_Level2a_file_structure#P_Frames)
[IC79 Split types](https://wiki.icecube.wisc.edu/index.php/IC79_offline_processing#NULLSPLIT_vs._SPLITSTREAM)
Something I do not have access to related to Q-frames and NullSplit implementation: [Introduce Null Splitting module that emits blank P-frame following a Q-frame](https://wiki.icecube.wisc.edu/index.php/Q-frames)

This [Link](https://wiki.icecube.wisc.edu/index.php/IT73-IC79_Composition_Analysis) talks about:

> IC79 L2a made standard nullsplit P frames with old IceTop information and all InIce pulse cleanings and reconstructions. For coincident IT-IC events a special hit cleaning and pulse selection was needed (see [IT73-IC79 Composition Analysis#Random_Coincidences](https://wiki.icecube.wisc.edu/index.php/IT73-IC79_Composition_Analysis#Random_Coincidences) ) . Also none of the standard L2 InIce reconstructions will be used as the IceTop direction and core reconstruction Laputop outperforms all InIce algorithms (for contained IceTop events) even when the InIce algorithms are seeded with the IceTop track reconstruction. A likelihood scan using gulliview.py shows the SPE and MPE likelihoods do NOT describe the likelihood space for muon bundles correctly. Hence the null split P frame will be removed.

All in all, I think it is fine to skip.

RasmusOrsoe:
The IceCube wizard has spoken; we can remove and should remove them. However @Andreas-MJ I would prefer that the fix is within the try/except block as I sketched in #389. 

image

asogaard:
> The IceCube wizard has spoken; we can remove and should remove them. However @Andreas-MJ I would prefer that the fix is within the try/except block as I sketched in #389.
> 

If I can offer my two cents here: I think it is good to have a function like `skip_frame(frame: I3Frame) -> bool` that can contain the logic to skip NullSplit frames, as well as other kinds, as @RasmusOrsoe suggests. But I would advise against repurposing the try-except block. As it is, this block is already too generic: it catches any and all exceptions, which is why we've had to add  the `# noqa: ...` caveat, which is already a sign something's fishy. And we'd be artificially raising an `AssertionError` where an if-statement would do the job, and would be more aligned with the syntax of a `skip_frame` function (returning a bool). To me, something like
```python
while i3_file_io.more():
    try:
    except I3SpecificException:
        continue 

    if skip_frame(frame):
        continue
```
would be more intuitive/readable, and no less elegant. But it's just a suggestion — I'll leave the verdict to y'all. :)

asogaard:
Hi @Andreas-MJ,
Since #389 contains a change that also addresses the same problem as this PR, I figure we can close this one once #389 is merged. I realise from #433 that #412 isn't resolved. Let the two of us discuss how to best proceed. :)

PR #435, opened 2023-02-17
HLC/SLC and FADW & ATWD pulseflags

RasmusOrsoe:
Adds HLC/SLC and FADW & ATWD pulseflags to pulsemaps.

(code is tested on level7 example i3 files)


MortenHolmRep (approved):
The bitwise shift looks fine, and I approve the code based on the valid tests.

With additional abbreviations (other than the 'rde' abbreviation we already had), I suggest we start adding descriptors of these to the readme for the various extractors, i.e. "The i3 feature extractor extracts the following list of items: rde - relative dom efficiency, hlc - hard local coincidence, etc.".
But I would say it falls under another issue about documentation of scripts in the examples folder, so I am fine with the PR as is.

Issue #431, opened 2023-02-15
Add option to generate coarsened pulsemap during db generation.


Aske-Rosted:
**Is your feature request related to a problem? Please describe.**
When working with high energy data the number of node activations becomes unreasonably large. Introducing a coarsening of the nodes can make the data workable. 

**Describe the solution you'd like**
I would like to have the option to generate a coarsened pulse map during the generation of the db files.

PR #438, opened 2023-02-22
Coarsening

Aske-Rosted:
An attempted implementation for solving #431. 

Creates a new class generators, which are called after the extractors in the  graphnet/src/graphnet/data/dataconverter.py script.

The idea is that this class in the future can be used more generally if one wishes to create a new feature variable pulsemap etc. when reading from the I3 files. 

Also introduces the CoarsePulseGenerator as the first example of a generator. Two of the functions called in CoarsePulseGenerator. 
have been left out of the class as they in principle could be moved to graphnet/src/graphnet/models/coarsening.py however I did not quite feel like that was appropriate either.


Aske-Rosted:
> 

I very much understand the reasoning behind doing the coarsening in the model itself and in that way not having to keep track of changes made to the data prior. However I think the answer to 

1. loading events with a large number of pulses using the Dataset/DataLoader classes is prohibitively slow, due to their vast size if un-preprocessed, and/or 
2. performing the clustering itself, if done within the Model, is prohibitively slow.

unfortunately is both. The clustering on the full dataset on the scale of one to a couple of days for the full dataset, and as far as I understand it would have to run once for each epoch. However, I say "I think" because I cannot really check the loading time of the full SRTInIcePulses due to the third issue

3. some of the events are so large that even a single event will not fit on the GPU Memory (or at least 2 events won't fit see  #441).  

I do not personally see a fix to this other than 

1. Coarsening in some step prior to the training process.
2. Doing the coarsening on CPU prior to throwing it to the GPU which I expect to be too slow from prior experience but admittedly have not actually checked. counting only the time spent in the coarsening script for 7.384 events 133 seconds was spent. This means that without any parallelization the naively expected increase in runtime can be roughly estimated by., 500.000/7.384 ≈ 68 
68 * 133 s ≈ 150 minutes,
Assuming that doubling number of CPU's halves the time spent with the 20 CPU's available to me this means that it might only increase the time of 1 epoch by 7.5 minutes. Since this is not unreasonable I will try to code up the "during runtime coarsening" and see what the actual increase in time is.  (However with the current setup I have increasing number of CPU's only increases runtime so I do not expect to be near the _optimal_ decrease in time by using multiple CPU's)
3. Upgrading the hardware...

MortenHolmRep:
@Aske-Rosted, I see that you are assigned to this PR. Have there been any recent developments that changed and/or solve your last post, or should someone else be assigned? :)

Aske-Rosted:
> @Aske-Rosted, I see that you are assigned to this PR. Have there been any recent developments that changed and/or solve your last post, or should someone else be assigned? :)

I have been looking into alternatives and have cleared to hurdle of running out of memory, however I cannot speak towards performance as I am unfortunately heavily IO-bound using this method. Some form of compression of the SQL files might relieve the stress on the IO... I think I currently need some advice or input as to what solution to work towards from now on.

PR #443, opened 2023-02-27
added model.inference() call

### LINKED ISSUE #442: _transform_prediction_inference not used by I3InferenceModule
**Describe the bug**
_transform_prediction_inference not used by I3InferenceModule. Therefore, predictions are not written to the expected units in cases where the module is loaded with a GNN that has _transform_inference specified.

**Expected behavior**
_transform_prediction_inference should be called.

RasmusOrsoe:
Closes #442


asogaard (approved):
👍

PR #444, opened 2023-02-28
added max_table_count to merger

RasmusOrsoe:
This PR introduces the optional argument `max_table_size` to `SQLiteDataConverter.merge_files`. The merging logic has been altered if this argument is not `None`. In a nutshell, this is what it does:

1. Will write to `"{database}_part_1.db"` until one table in that database exceeds `max_table_size`.
2. When the `max_table_size` has been exceeded, a new partition is created called `"{database}_part_2.db"`
3.  This will continue until there are no more temporary databases to merge.

This allows us to get a reasonable control with table sizes for datasets that have a high pulse count pr. event and should therefore improve query times.

Each of these partitions can be used with `EnsembleDataset` to create one, unified dataset for training. 

@Aske-Rosted fyi.

### DIFF
--- a/src/graphnet/data/sqlite/sqlite_dataconverter.py
+++ b/src/graphnet/data/sqlite/sqlite_dataconverter.py
@@ -71,7 +71,10 @@ class SQLiteDataConverter(DataConverter):
             self.warning(f"No data saved to {output_file}")
 
     def merge_files(
-        self, output_file: str, input_files: Optional[List[str]] = None
+        self,
+        output_file: str,
+        input_files: Optional[List[str]] = None,
+        max_table_size: Optional[int] = None,
     ) -> None:
         """SQLite-specific method for merging output files/databases.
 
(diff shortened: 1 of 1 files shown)


asogaard:
Out of curiosity: What is the benefit of doing this, vs. specifying a certain [batching scheme](https://github.com/graphnet-team/graphnet/blob/main/src/graphnet/data/dataconverter.py#L104-L139) when generating the individual databases?

RasmusOrsoe:
@asogaard I think the benefit here is that you get a more direct control over the number of rows in the pulsemap tables. You're right that you could mimic this behavior with the batch patterns, but I personally find this a bit more intuitive. But in all honesty, I do not really understand the batch pattern part of the code very well, so if I'm overlooking something obvious feel free to point it out!

MortenHolmRep (approved):
I think it looks good.
In my opinion the implementation does not supersede the batch scheme, but rather enables more control by giving more options.

PR #456, opened 2023-03-19
Simplify examples

### LINKED ISSUE #175: Make prediction column names self-contained
Hi

When we want to predict, the repo leans towards the approach:

`predictions = get_predictions(trainer, model, validation_dataloader, [target + '_pred'])`

This means that the user of the trained model is **_required_** to in advance know the column names of the output of the trained model.

I'd like to suggest that we instead move to a self-contained approach that would allow for a syntax on the form:

`predictions = get_predictions(trainer, model, validation_dataloader, model.output_column_names`)

In addition I think it would be good for portability to include syntax for trained models such as:

```
model_predicts_this_target = model.target
model_was_trained_on_this_pulsemap_and_therefore_requires_it_for_predictions = model.pulsemap

```

Rasmus

RasmusOrsoe:
This PR attempts to introduce a simplified training script (`01_train_model_simple.py`) where we showcase training syntax with minimal lines of code. What I did to simplify was:

1. Remove weights and biases from this example (it's been a showstopper for some people). 
2. Remove CLI for simplicity
3. Hard-coded configs for clearity
4. Provided a default value for callbacks (automatically includes early stopping if validation loader is provided)
5. Provided default values for `target_labels` and `output_labels` for all tasks, and making these available under `model.output_labels` and `model.target_labels` (closes #175 )

These changes has brought down the training example from 183 to 68 lines.

In addition, I've made the default values for `target_labels` and `output_labels` more visual in `Task`'s, and written them as `@propery` such that we have to supply these default labels for tasks in the future.  I think this makes it a bit easier to grasp what a given `Task` requires and returns.

I have left the other training scripts untouched but renamed them to make the distinction easier. 

README has been updated.


asogaard:
Hi @RasmusOrsoe,

Thanks for this! Some initial thoughts before I go into review-mode:

As I see it, this PR does two things,

* Adds a simplified training script (items 1-3 in your list above)
* Addresses #175 by adding default output names to each `Task` (items 4 and 5 in your list above)

The second point I totally get — it fixes a known and well-defined issue that we have discussed and agree is worth solving. 🚀 

The first point I am not sure I understand the need for. If the point is to disable W&B, I figure we could just add a `--wandb` argument to the CLI, and have it be off by default. As for removing the CLI altogether and hard-coding the config file (as opposed to providing a default one in the CLI), I am not sure I see how that could improve the example. But I am open to convincing. :)

RasmusOrsoe:
Hey @asogaard 

I definitely think there is a need for _some_ resource that provides a simple point of entry into the code base in as few lines as possible.

I'd like the example script to default to no wand (not because of personal feelings!) because a number of people have had trouble running the example scripts because of wand. It was actually caught on a live stream [here](https://youtu.be/OqXfsa8w5R4?t=3184).

I hard-coded the configs because the example scripts (our only point of entry currently) hides these away in default arguments and doesn't really provide a transparent view of what it really is. 

 Perhaps the conclusion is that an actual written tutorial is a better format that an example script. So we could proceed as:


1. Remove changes to example scripts
2. Add a written tutorial instead that explains each element in my suggested training script, such that new users are better equipped.

What do you think @asogaard?

asogaard:
> Hey @asogaard
> 
> I definitely think there is a need for _some_ resource that provides a simple point of entry into the code base in as few lines as possible.
> 
> I'd like the example script to default to no wand (not because of personal feelings!) because a number of people have had trouble running the example scripts because of wand. It was actually caught on a live stream [here](https://youtu.be/OqXfsa8w5R4?t=3184).
> 
> I hard-coded the configs because the example scripts (our only point of entry currently) hides these away in default arguments and doesn't really provide a transparent view of what it really is.
> 
> Perhaps the conclusion is that an actual written tutorial is a better format that an example script. So we could proceed as:
> 
> 1. Remove changes to example scripts
> 2. Add a written tutorial instead that explains each element in my suggested training script, such that new users are better equipped.
> 
> What do you think @asogaard?

I completely agree that our main training example should be as simple as possible, but no simpler! 😊  So any ideas for making them a more friendly point of entry are more than welcome!

I agree with your proposed solution, but please do feel free to make wandb opt-in (**or should I make a separate PR for that?**). Regarding the config files: all default values are described when you run
```bash
$ python examples/04_training/01_train_model.py --help
usage: 01_train_model.py [-h] [--gpus GPUS [GPUS ...]] [--max-epochs MAX_EPOCHS] [--early-stopping-patience EARLY_STOPPING_PATIENCE]
                         [--batch-size BATCH_SIZE] [--num-workers NUM_WORKERS] [--dataset-config DATASET_CONFIG] [--model-config MODEL_CONFIG]
                         [--prediction-names PREDICTION_NAMES [PREDICTION_NAMES ...]] [--suffix SUFFIX]
(...)
  --dataset-config DATASET_CONFIG
                        Path to dataset config file (default: {CONFIG_DIR}/datasets/training_example_data_sqlite.yml)
  --model-config MODEL_CONFIG
                        Path to model config file (default: {CONFIG_DIR}/models/example_energy_reconstruction_model.yml)
```
and similar for all example scripts. This is mentioned [here](https://github.com/graphnet-team/graphnet/tree/main/examples), but if people are not familiar the standard python CLI, I think this would be good to point out in a tutorial or similar!

RasmusOrsoe:
@asogaard  OK - I think that's it.

src/graphnet/models/model.py
```diff
@@ -164,6 +183,9 @@ def predict_as_dataframe(
             additional_attributes = []
         assert isinstance(additional_attributes, list)
 
+        if prediction_columns is None:
+            prediction_columns = self.prediction_columns
```
asogaard, line 187: These are not set for all instances of `Model` (currently, in this PR, this is only done for `StandardModel` which inherits from `Mode`), so either should we do that, otherwise we shouldn't presuppose the existence of this attribute.

src/graphnet/models/model.py
```diff
@@ -146,7 +165,7 @@ def predict(
     def predict_as_dataframe(
         self,
         dataloader: DataLoader,
-        prediction_columns: List[str],
+        prediction_columns: List[str] = None,
```
asogaard, line 168: If we want to make it optional whether to specify a list.
```suggestion
        prediction_columns: Optional[List[str]] = None,
```

src/graphnet/models/model.py
```diff
@@ -79,15 +81,32 @@ def fit(
         *,
         max_epochs: int = 10,
         gpus: Optional[Union[List[int], int]] = None,
-        callbacks: Optional[List[Callback]] = None,
+        callbacks: Optional[List[Callback]] = [ProgressBar()],
```
asogaard, line 84: It is not quite kosher to add lists as default arguments. (Something about [them being mutable](https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments).) So I would keep the None, and then add the `ProgressBar` to the list of callback if it isn't already there.

src/graphnet/models/model.py
```diff
@@ -79,15 +81,32 @@ def fit(
...
+                    has_es = True
+            if has_es is False:
+                callbacks.append(
+                    EarlyStopping(
+                        monitor="val_loss",
+                        patience=early_stopping_patience,
+                    )
+                )
+                self.info("EarlyStopping callback added automatically.")
+
```
asogaard, line 109: I get the point of adding the `ProgressBar` by default, as it has no functional impact. However, I'm a little bit unsure about this one: So far, we have chosen to set callback based on the `callbacks` argument, whereby the user needs to specify the list of callbacks they want, and is quite common practice. Now, we're considering mixing this pattern with adding callbacks based on specific, additional arguments (here: `early_stopping_patience`). I think we should do one or the other, i.e., setting callbacks through a user-provided list _or_ through dedicated arguments to `Model.fit`.

src/graphnet/models/model.py
```diff
@@ -178,6 +200,9 @@ def predict_as_dataframe(
                 "doesn't resample batches; or do not request "
                 "`additional_attributes`."
             )
+        self.info(
+            f"Column names for predictions are: \n {self.prediction_columns}"
+        )
```
asogaard, line 205: See comment above.

src/graphnet/models/standard_model.py
```diff
@@ -61,6 +61,15 @@ def __init__(
...
         self._scheduler_config = scheduler_config or dict()
+        self._configure_model_constants()
+
+    def _configure_model_constants(self) -> None:
+        self.target = [
+            target for task in self._tasks for target in task._target_labels
+        ]
+        self.prediction_columns = [
+            column for task in self._tasks for column in task._output_labels
+        ]
```
asogaard, line 72: Just a suggestion: Standardise the names of `target` -> `target_labels` and `preduction_columns` -> `prediction_labels`, and expose these as properties (e.g., `model.target_labels`).
```suggestion

    @property
    def target_labels(self) -> List[str]:
        return [
            label for task in self._tasks for label in task._target_labels
        ]
    
    @property
    def prediction_labels(self) -> List[str]:
        return [
            label for task in self._tasks for label in task._output_labels
        ]
```

src/graphnet/models/task/reconstruction.py
```diff
@@ -12,6 +12,8 @@ class AzimuthReconstructionWithKappa(Task):
     """Reconstructs azimuthal angle and associated kappa (1/var)."""
 
     # Requires two features: untransformed points in (x,y)-space.
+    default_target_labels = ["azimuth"]
+    default_output_labels = ["azimuth_pred", "azimuth_kappa"]
```
asogaard, line 16: To be consistent with suggestion in `standard_model.py`. Similar below.
```suggestion
    default_prediction_labels = ["azimuth_pred", "azimuth_kappa"]
```

src/graphnet/models/task/reconstruction.py
```diff
@@ -28,6 +30,8 @@ class AzimuthReconstruction(AzimuthReconstructionWithKappa):
     """Reconstructs azimuthal angle."""
 
     # Requires two features: untransformed points in (x,y)-space.
+    default_target_labels = ["azimuth"]
+    default_output_labels = ["azimuth_pred"]
```
asogaard, line 34:
```suggestion
    default_prediction_labels = ["azimuth_pred"]
```

src/graphnet/models/task/reconstruction.py
```diff
@@ -46,6 +50,15 @@ class DirectionReconstructionWithKappa(Task):
     """Reconstructs direction with kappa from the 3D-vMF distribution."""
 
     # Requires three features: untransformed points in (x,y,z)-space.
+    default_target_labels = [
+        "direction"
+    ]  # contains dir_x, dir_y, dir_z see https://github.com/graphnet-team/graphnet/blob/95309556cfd46a4046bc4bd7609888aab649e295/src/graphnet/training/labels.py#L29
+    default_output_labels = [
```
asogaard, line 56:
```suggestion
    default_prediction_labels = [
```

src/graphnet/models/task/reconstruction.py
```diff
@@ -61,6 +74,8 @@ class ZenithReconstruction(Task):
     """Reconstructs zenith angle."""
 
     # Requires two features: zenith angle itself.
+    default_target_labels = ["zenith"]
+    default_output_labels = ["zenith_pred"]
```
asogaard, line 78:
```suggestion
    default_prediction_labels = ["zenith_pred"]
```

src/graphnet/models/task/reconstruction.py
```diff
@@ -72,6 +87,8 @@ class ZenithReconstructionWithKappa(ZenithReconstruction):
     """Reconstructs zenith angle and associated kappa (1/var)."""
 
     # Requires one feature in addition to `ZenithReconstruction`: kappa (unceratinty; 1/variance).
+    default_target_labels = ["zenith"]
+    default_output_labels = ["zenith_pred", "zenith_kappa"]
```
asogaard, line 91:
```suggestion
    default_prediction_labels = ["zenith_pred", "zenith_kappa"]
```

src/graphnet/models/task/reconstruction.py
```diff
@@ -85,6 +102,8 @@ class EnergyReconstruction(Task):
     """Reconstructs energy using stable method."""
 
     # Requires one feature: untransformed energy
+    default_target_labels = ["energy"]
+    default_output_labels = ["energy_pred"]
```
asogaard, line 106:
```suggestion
    default_prediction_labels = ["energy_pred"]
```

src/graphnet/models/task/reconstruction.py
```diff
@@ -97,6 +116,8 @@ class EnergyReconstructionWithPower(Task):
     """Reconstructs energy."""
 
     # Requires one feature: untransformed energy
+    default_target_labels = ["energy"]
+    default_output_labels = ["energy_pred"]
```
asogaard, line 120:
```suggestion
    default_prediction_labels = ["energy_pred"]
```

src/graphnet/models/task/reconstruction.py
```diff
@@ -108,6 +129,8 @@ class EnergyReconstructionWithUncertainty(EnergyReconstruction):
     """Reconstructs energy and associated uncertainty (log(var))."""
 
     # Requires one feature in addition to `EnergyReconstruction`: log-variance (uncertainty).
+    default_target_labels = ["energy"]
+    default_output_labels = ["energy_pred", "energy_sigma"]
```
asogaard, line 133:
```suggestion
    default_prediction_labels = ["energy_pred", "energy_sigma"]
```

src/graphnet/models/task/reconstruction.py
```diff
@@ -122,6 +145,13 @@ class VertexReconstruction(Task):
     """Reconstructs vertex position and time."""
 
     # Requires four features, x, y, z, and t.
+    default_target_labels = ["vertex"]
+    default_output_labels = [
```
asogaard, line 149:
```suggestion
    default_prediction_labels = [
```

src/graphnet/models/task/reconstruction.py
```diff
@@ -138,6 +168,12 @@ class PositionReconstruction(Task):
     """Reconstructs vertex position."""
 
     # Requires three features, x, y, and z.
+    default_target_labels = ["position"]
+    default_output_labels = [
```
asogaard, line 172:
```suggestion
    default_prediction_labels = [
```

src/graphnet/models/task/reconstruction.py
```diff
@@ -169,6 +207,8 @@ class InelasticityReconstruction(Task):
     """
 
     # Requires one features: inelasticity itself
+    default_target_labels = ["elasticity"]
+    default_output_labels = ["elasticity_pred"]
```
asogaard, line 211:
```suggestion
    default_prediction_labels = ["elasticity_pred"]
```

src/graphnet/models/task/task.py
```diff
@@ -27,13 +27,26 @@ class Task(Model):
...
+    def default_target_labels(self) -> List[str]:
+        """Return default target labels."""
+        return self._default_target_labels
+
+    @property
+    @abstractmethod
+    def default_output_labels(self) -> List[str]:
+        """Return default output labels."""
+        return self._default_output_labels
+
```
asogaard, line 41:
```suggestion
    @property
    @abstractmethod
    def default_prediction_labels(self) -> List[str]:
        """Return default prediction labels."""
        return self._default_prediction_labels

```

src/graphnet/models/task/task.py
```diff
@@ -27,13 +27,26 @@ class Task(Model):
...
     @save_model_config
     def __init__(
         self,
         *,
         hidden_size: int,
-        target_labels: Union[str, List[str]],
         loss_function: "LossFunction",
+        target_labels: Union[str, List[str]] = None,
+        output_labels: Optional[Union[str, List[str]]] = None,
```
asogaard, line 49:
```suggestion
        target_labels: Optional[Union[str, List[str]]] = None,
        prediction_labels: Optional[Union[str, List[str]]] = None,
```

src/graphnet/models/task/task.py
```diff
@@ -46,10 +59,13 @@ def __init__(
             hidden_size: The number of nodes in the layer feeding into this
                 tasks, used to construct the affine transformation to the
                 predicted quantity.
+            loss_function: Loss function appropriate to the task.
             target_labels: Name(s) of the quantity/-ies being predicted, used
                 to extract the  target tensor(s) from the `Data` object in
                 `.compute_loss(...)`.
-            loss_function: Loss function appropriate to the task.
+            output_labels: The name(s) of each column that is outputted by
```
asogaard, line 66:
```suggestion
            prediction_labels: The name(s) of each column that is predicted by
```

src/graphnet/models/task/task.py
```diff
@@ -76,14 +92,22 @@ def __init__(
...
         if isinstance(target_labels, str):
             target_labels = [target_labels]
-
+        if output_labels is not None:
+            if isinstance(output_labels, str):
+                output_labels = [output_labels]
+        else:
+            output_labels = [target + "_pred" for target in target_labels]
+        assert isinstance(target_labels, List)  # mypy
+        assert isinstance(output_labels, List)  # mypy
```
asogaard, line 106: **Shouldn't we use the `default_prediction_labels` here as well?**
```suggestion
        if prediction_labels is not None:
            if isinstance(prediction_labels, str):
                prediction_labels = [prediction_labels]
        else:
            prediction_labels = [target + "_pred" for target in target_labels]
        assert isinstance(target_labels, List)  # mypy
        assert isinstance(prediction_labels, List)  # mypy
```

src/graphnet/models/task/task.py
```diff
@@ -76,14 +92,22 @@ def __init__(
...
         # Member variables
         self._regularisation_loss: Optional[float] = None
         self._target_labels = target_labels
+        self._output_labels = output_labels
```
asogaard, line 110:
```suggestion
        self._prediction_labels = prediction_labels
```

asogaard (commented):
Some suggestions. :)

RasmusOrsoe:
OK. I think this is it @asogaard

src/graphnet/models/model.py
```diff
@@ -164,6 +194,14 @@ def predict_as_dataframe(
...
         assert isinstance(additional_attributes, list)
 
+        if prediction_columns is None:
+            try:
+                prediction_columns = self.prediction_columns
+            except AttributeError:
+                assert (
+                    1 == 2
+                ), "Could not infer prediction_columns from model. Please specify prediction_columns."
+        assert isinstance(prediction_columns, list)
```
asogaard, line 204: I don't think this is a great solution. 😊  If we're assuming the existence of the property `self.prediction_columns` in `Model.predict_as_dataframe` then we should ensure that it is set for all instances of `Model`.

You could do something like:
```python
class Model:
    def __init__(self, <...>, prediction_columns: Optional[List[str]] = None, ...):
        <...>
        self._prediction_columns = prediction_columns

    @property
    def prediction_columns(self) -> Optional[List[str]]:
        return self._prediction_columns

    @property
    def target_columns(self) -> Optional[List[str]]:
        raise NotImplementedError
```
This means that all instances of `Model` have the property `self.prediction_columns`, and that instances of `StandardModel` override the two properties `self.prediction_columns` and `self.target_columns` in a way that is meaningful for their implementation.

src/graphnet/models/standard_model.py
```diff
@@ -62,6 +62,16 @@ def __init__(
...
 
+    @property
+    def target_labels(self) -> List[str]:
+        """Return target label."""
+        return [label for task in self._tasks for label in task._target_labels]
+
+    @property
+    def prediction_labels(self) -> List[str]:
+        """Return prediction labels."""
+        return [label for task in self._tasks for label in task._output_labels]
```
asogaard, line 73: The name of the member attribute should be updated.
```suggestion
        return [label for task in self._tasks for label in task._prediction_labels]
```

src/graphnet/models/task/task.py
```diff
@@ -76,14 +92,22 @@ def __init__(
...
         if isinstance(target_labels, str):
             target_labels = [target_labels]
-
+        if prediction_labels is not None:
+            if isinstance(prediction_labels, str):
+                prediction_labels = [prediction_labels]
+        else:
+            prediction_labels = [target + "_pred" for target in target_labels]
+        assert isinstance(target_labels, List)  # mypy
+        assert isinstance(prediction_labels, List)  # mypy
```
asogaard, line 106: Shouldn't you also use `self.default_prediction_labels` here, @RasmusOrsoe?

asogaard (commented):
I have added a few comments that I think might have been missed in the last batch, @RasmusOrsoe. :)

RasmusOrsoe:
@asogaard Could we maybe go through these comments on zoom sometime?

asogaard:
It's probably due to the pandas version problem I mention in the other PR. We'll rerun the tests once that is merged.

asogaard:
> @asogaard I think this it it. However, now I see that the data conversion unit tests fail even though no changes have been made to that part of the repo. I've scrolled through the error messages and I'm a bit puzzled. Do you have any ideas what could be causing this?

Try merging `main` into this feature branch, following #476. 🤞

src/graphnet/models/task/task.py
```diff
@@ -76,14 +92,22 @@ def __init__(
...
         if isinstance(target_labels, str):
             target_labels = [target_labels]
-
+        if prediction_labels is not None:
+            if isinstance(prediction_labels, str):
+                prediction_labels = [prediction_labels]
+        else:
+            prediction_labels = self.default_prediction_labels
+        assert isinstance(target_labels, List)  # mypy
+        assert isinstance(prediction_labels, List)  # mypy
```
asogaard, line 106: To keep the structure consistent with `target_labels` above.
```suggestion

        if prediction_labels is None:
            prediction_labels = self.default_prediction_labels
        if isinstance(prediction_labels, str):
            prediction_labels = [prediction_labels]
        
        assert isinstance(target_labels, List)  # mypy
        assert isinstance(prediction_labels, List)  # mypy
```

src/graphnet/models/model.py
```diff
@@ -110,6 +121,34 @@ def fit(
...
+                callbacks.append(
+                    EarlyStopping(
+                        monitor="val_loss",
+                        patience=5,
+                    )
+                )
+                self.warn_once(
+                    "Got validation dataloader but no EarlyStopping callback. An EarlyStopping callback has been added automatically with patience=5 and monitor = 'val_loss'."
+                )
+        return callbacks
```
asogaard, line 150: Avoiding one level of nesting, wrapping a line to 79 characters, and just a wee bit of cosmetics.
```suggestion
        if val_dataloader is None:
            return callbacks
            
        has_early_stopping = False
        assert isinstance(callbacks, list)
        for callback in callbacks:
            if isinstance(callback, EarlyStopping):
                has_early_stopping = True
                
        if not has_early_stopping:
            callbacks.append(
                EarlyStopping(
                    monitor="val_loss",
                    patience=5,
                )
            )
            self.warn_once(
                "Got validation dataloader but no EarlyStopping callback. An "
                "EarlyStopping callback has been added automatically with "
                "patience=5 and monitor = 'val_loss'."
            )
        return callbacks
```

asogaard (approved):
Thanks so much for your hard work here, @RasmusOrsoe! 🙏   I think this looks great, and is ready merge. I have added a few cosmetic suggestions that you can consider before hitting the big, green button.

PR #465, opened 2023-03-21
Remove len(data) assertion for SQLiteDataconverter save method

### LINKED ISSUE #464: SQLiteDataconverter fails when extraction from i3 file is empty
**Describe the bug**
SQLiteDataconverter fails when extraction from i3 file is empty. This could be files with no physics frames.

**Expected behavior**
Should just skip these cases and not assert len(data).

**Full traceback**
```
  File "/cvmfs/icecube.opensciencegrid.org/py3-v4.1.1/RHEL_7_x86_64/lib/python3.7/multiprocessing/pool.py", line 121, in worker
    result = (True, func(*args, **kwds))
  File "/lustre/hpc/hep/pcs557/github/graphnet/src/graphnet/data/dataconverter.py", line 83, in wrapper
    output_file = process_method(self, *args)
  File "/lustre/hpc/hep/pcs557/github/graphnet/src/graphnet/data/dataconverter.py", line 384, in _process_file
    self.save_data(data, output_file)
  File "/lustre/hpc/hep/pcs557/github/graphnet/src/graphnet/data/sqlite/sqlite_dataconverter.py", line 35, in save_data
    assert len(data)
AssertionError
```

RasmusOrsoe:
closes #464

### DIFF
--- a/src/graphnet/data/sqlite/sqlite_dataconverter.py
+++ b/src/graphnet/data/sqlite/sqlite_dataconverter.py
@@ -38,46 +38,43 @@ class SQLiteDataConverter(DataConverter):
                 f"No data saved to {output_file}"
             )
             return
-        
-            saved_any = False
-            dataframe = OrderedDict([(key, pd.DataFrame()) for key in data[0]])
-            for data_dict in data:
-                for key, data_values in data_dict.items():
-                    df = construct_dataframe(data_values)
 
-                    if (
-                        self.any_pulsemap_is_non_empty(data_dict)
-                        and len(df) > 0
-                    ):
-                        # only include data_dict in temp. databases if at least one pulsemap is non-empty,
-                        # and the current extractor (df) is also non-empty (also since truth is always non-empty)
-                        if len(dataframe[key]):
-                            assert isinstance(dataframe[key], pd.DataFrame)
-                            dataframe[key] = dataframe[key].append(
-                                df, ignore_index=True, sort=True
-                            )
-                        else:
-                            dataframe[key] = df
+        saved_any = False
+        dataframe = OrderedDict([(key, pd.DataFrame()) for key in data[0]])
+        for data_dict in data:
+            for key, data_values in data_dict.items():
+                df = construct_dataframe(data_values)
 
-            # Save each dataframe to SQLite database
-            self.debug(f"Saving to {output_file}")
-            for table, df in dataframe.items():
-                if len(df) > 0:
-                    create_table_and_save_to_sql(
-                        df,
-                        table,
-                        output_file,
-                        default_type="FLOAT",
-                        integer_primary_key=not (
-                            is_pulse_map(table) or is_mc_tree(table)
-                        ),
-                    )
-                    saved_any = True
+                if self.any_pulsemap_is_non_empty(data_dict) and len(df) > 0:
+                    # only include data_dict in temp. databases if at least one pulsemap is non-empty,
+                    # and the current extractor (df) is also non-empty (also since truth is always non-empty)
+                    if len(dataframe[key]):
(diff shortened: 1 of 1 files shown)


asogaard (commented):
Hi @RasmusOrsoe,

Thanks for this PR! The intention behind the `assert`-statement, I believe, was to let the user know that no data was extracted, as this shouldn't happen. If there are legitimate cases where we expect this to happen, and don't want the code to stop, I agree `assert` might not be the way to go. However, instead of wrapping all of these lines in an `if`-statement, thereby increasing the cognitive complexity, would it be possible to do something like:

```python
if len(data) == 0:
    self.warning(
        "No data was extracted from the processed I3 file(s). "
        f"No data saved to {output_file}".
    )
    return

dataframe = OrderedDict([(key, pd.DataFrame()) for key in data[0]])
for (...)
```

RasmusOrsoe:
@asogaard I've added your recommendation.

src/graphnet/data/sqlite/sqlite_dataconverter.py
```diff
@@ -32,43 +32,52 @@ def save_data(self, data: List[OrderedDict], output_file: str) -> None:
...
-
-        if saved_any:
-            self.debug("- Done saving")
+        if len(data) == 0:
+            self.warning(
+                "No data was extracted from the processed I3 file(s). "
+                f"No data saved to {output_file}"
+            )
+            return
         else:
```
asogaard, line 41: If you add a `return`-statement, there is no need for the `else`-block. 
```suggestion
            return
        
```

src/graphnet/data/sqlite/sqlite_dataconverter.py
```diff
@@ -32,43 +32,52 @@ def save_data(self, data: List[OrderedDict], output_file: str) -> None:
...
+                        integer_primary_key=not (
+                            is_pulse_map(table) or is_mc_tree(table)
+                        ),
+                    )
+                    saved_any = True
+
+            if saved_any:
+                self.debug("- Done saving")
+            else:
+                self.warning(f"No data saved to {output_file}")
```
asogaard, line 80: Update indentation to be consistent with removal of `else`-statement above.
```suggestion
        saved_any = False
        dataframe = OrderedDict([(key, pd.DataFrame()) for key in data[0]])
        for data_dict in data:
            for key, data_values in data_dict.items():
                df = construct_dataframe(data_values)

                if (
                    self.any_pulsemap_is_non_empty(data_dict)
                    and len(df) > 0
                ):
                    # only include data_dict in temp. databases if at least one pulsemap is non-empty,
                    # and the current extractor (df) is also non-empty (also since truth is always non-empty)
                    if len(dataframe[key]):
                        assert isinstance(dataframe[key], pd.DataFrame)
                        dataframe[key] = dataframe[key].append(
                            df, ignore_index=True, sort=True
                        )
                    else:
                        dataframe[key] = df

        # Save each dataframe to SQLite database
        self.debug(f"Saving to {output_file}")
        for table, df in dataframe.items():
            if len(df) > 0:
                create_table_and_save_to_sql(
                    df,
                    table,
                    output_file,
                    default_type="FLOAT",
                    integer_primary_key=not (
                        is_pulse_map(table) or is_mc_tree(table)
                    ),
                )
                saved_any = True

        if saved_any:
            self.debug("- Done saving")
        else:
            self.warning(f"No data saved to {output_file}")
```

asogaard (commented):
Hi @RasmusOrsoe,

Again in order to reduce the complexity, I have included two concrete suggestions to remove a superfluous `else`-block. Let me know what you think. :)

RasmusOrsoe:
@asogaard  Alrighty - should be good now

asogaard (approved):
**chef's kiss**

PR #469, opened 2023-03-28
added I3QUESOExtractor

RasmusOrsoe:
Adds I3Extractor for the graphnet-powered event selection for IceCube Upgrade named QUESO.

### DIFF
--- a/src/graphnet/data/extractors/i3quesoextractor.py
+++ b/src/graphnet/data/extractors/i3quesoextractor.py
@@ -1,0 +1,46 @@
+"""Extract Event Selection Labels from the QUESO event selection."""
+
+from typing import TYPE_CHECKING, Dict
+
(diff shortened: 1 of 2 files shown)


src/graphnet/data/extractors/i3quesoextractor.py
```diff
@@ -0,0 +1,46 @@
...
+            "QuesoL3_Vars_cleaned_num_hits_fid_vol",
+            "QuesoL3_Vars_cleaned_vertexZ",
+            "QuesoL3_Vars_uncleaned_length",
+        ]
+        output = {}
+        for key in keys:
+            try:
+                value = frame[key].value
+            except KeyError:
+                value = self._padding_value
```
asogaard, line 39: I think something like this is more legible, but they're functionally the same.
```suggestion
            if key in frame:
                value = frame[key].value
            else:
                value = self._padding_value
```

PR #485, opened 2023-04-14
Add key property to Label add simplify Dataset.add_label

### LINKED ISSUE #483: Add `key` property to `Label`
The `Label` class has a single property, which is only available as an "internal" variable, `self._key`.

This is not a purely internal variable — it is relevant to be able to access it from outside the class — so it might be useful to make it "public" as a `self.key` property.

**Describe alternatives you've considered**
We could just use `label._key` to access the name of the key, but that's not really good form.

**Additional context**

### LINKED ISSUE #484: Make `Dataset.add_label` more compatible with `Label` class
Currently the [`Dataset.add_label`](https://github.com/graphnet-team/graphnet/blob/main/src/graphnet/data/dataset.py#L352) method takes arguments `(key: str, label: Callable)`. However, the `Label` class allows for specifying the key in the constructor, and thus it becomes a little clumsy to then have to specify the key again when adding the label, i.e., `dataset.add_label(label.key, label)` (see also #483).

It could be simpler to change the method syntax for `Dataset.add_label(fn: Callable, key: Optional[str])`; if `fn` is an instance if `Label`, we could then directly take the key from there; if not, we would then `assert isinstance(key, str)`.

asogaard:
Closes #483 
Closes #484

### DIFF
--- a/src/graphnet/data/dataset.py
+++ b/src/graphnet/data/dataset.py
@@ -21,15 +21,16 @@ from torch.utils.data import ConcatDataset
 from torch_geometric.data import Data
 
 from graphnet.constants import GRAPHNET_ROOT_DIR
+from graphnet.data.utilities.string_selection_resolver import (
+    StringSelectionResolver,
+)
+from graphnet.training.labels import Label
 from graphnet.utilities.config import (
     Configurable,
     DatasetConfig,
     save_dataset_config,
 )
 from graphnet.utilities.logging import Logger
-from graphnet.data.utilities.string_selection_resolver import (
-    StringSelectionResolver,
-)
 
 
 class ColumnMissingException(Exception):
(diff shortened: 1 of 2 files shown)


RasmusOrsoe (changes requested):
Hey @asogaard. Very nice!

Could you add the `self._key = direction` to our one-and-only Direction label?

asogaard:
> Hey @asogaard. Very nice!
> 
> Could you add the `self._key = "direction"` to our one-and-only Direction label?

The `_key` property of the `Direction` label gets set [here](https://github.com/graphnet-team/graphnet/pull/485/files#diff-ed1c5bef98dc6e0a316eb4f07300070489309bb313d8d1090b06d28879198083R45), i.e., in the call to the base class (`Label` constructor). It gets set as `"azimuth_zenith"` based on the two keys used to define the direction. Is that what you have in mind, @RasmusOrsoe?

RasmusOrsoe (approved):
@asogaard sorry! I missed that completely. 

Looking at it now; It's a bit of a strange key-name. Should we stick with that? I would feel better about having it default to "direction". I leave that to you!

asogaard:
I have updated the default label to `"direction"`, @RasmusOrsoe. 👍

PR #486, opened 2023-04-14
Remove sqlite_selection.py

### LINKED ISSUE #466: Retire sqlite_selection.py
The module [`sqlilte_selection.py`](https://github.com/graphnet-team/graphnet/blob/main/src/graphnet/data/sqlite/sqlite_selection.py) contains a number of quite specialised functions for selecting subsets of events, with hardcoded variables and such. 

The `DatasetConfig` class allows for doing quite granular selections (see, e.g., [here](https://github.com/graphnet-team/graphnet/blob/main/src/graphnet/utilities/config/dataset_config.py#L89-L94) and [here](https://github.com/graphnet-team/graphnet/blob/main/configs/datasets/PID_classification_last_one_lvl3MC.yml#L26-L43)) that are more portable. 

Do we want to continue supporting the functions in [`sqlilte_selection.py`](https://github.com/graphnet-team/graphnet/blob/main/src/graphnet/data/sqlite/sqlite_selection.py), both these functions specifically and also this way of defining selections as utility functions in the main `graphnet` repo? Or are there analysis-specific and should be kept in dedicated analysis repos? _If_ we want to keep supporting them: Are the current functions the right ones? How do we make sure that they are not too specific to certain use cases/datasets?

**Update:** It is agreed to remove this module.

asogaard:
Closes #466


MortenHolmRep (approved):
Approved in accordance with the group discussion.

Issue #478, opened 2023-04-13
Make GraphNeT compatible with pandas >= 2.0


asogaard:
Currently there is at least on breaking change in pandas 2.0 (`DataFrame.append` being removed) which means that GraphNeT doesn't work with it. It would be nice to be compatible with this newest version.

PR #490, opened 2023-04-23
Changed df.append to pd.concat to be compatible with pandas 2.0

AMHermansen:
I changed all the deprecated DataFrame.append to pandas.concat to be compatible with pandas2.0. There may be other issues which need to be addressed, before the entire repository is compatible with pandas 2.0.
This addresses (at least to some extent Issue #478

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -16,7 +16,6 @@ INSTALL_REQUIRES = [
     "dill>=0.3",
     "matplotlib>=3.5",
     "numpy>=1.21",
-    "pandas>=1.3, <2.0",
     "pyarrow",
     "pydantic",
     "ruamel.yaml",
--- a/setup.py
+++ b/setup.py
@@ -16,6 +16,7 @@ INSTALL_REQUIRES = [
     "dill>=0.3",
     "matplotlib>=3.5",
     "numpy>=1.21",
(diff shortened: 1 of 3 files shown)


asogaard (commented):
Hi @AMHermansen,

Thanks for this PR! 👏   A few suggestions:

* As this PR seeks to make graphnet compatible with pandas >= 2.0 I think it should also include the removal of [this line in setup.py](https://github.com/graphnet-team/graphnet/blob/main/setup.py#L19).
* Simply changing all calls to `DataFrame.append` -> `pd.concat` technical works, but also ignores the spirit of the deprecation, [see here](https://github.com/pandas-dev/pandas/issues/35407#issue-665344871):
> (...) users are always better off building up a list of values and passing them to the constructor, or building up a list of NDFrames followed by a single concat
* Instead of defining an empty `pd.DataFrame` outside each for-loop and then concat-appending new DataFrames to it, I think it would be more true to the spirit of the `pd.concat` method to simple add DataFrames to an initially empty list inside the for-loop, and then calling `pd.concat` only once, on this list of elements, immediately after the for-loop. 
 
Let me know what you think. :)

asogaard (approved):
Looks good! 🎉

PR #494, opened 2023-05-02
Multi-class classification example script

MortenHolmRep:
Added example script for multi-class classification that use the dataset and model configs.


RasmusOrsoe (approved):
Very nice @MortenHolmRep!!

PR #500, opened 2023-05-03
Feature/define tito model

ArturoLlorente:
Added GNN used in Kaggle tito solution (1st position). Layers "EdgeConv0" and "DynTrans" and gnn "DynEdgeTITO" created.


RasmusOrsoe:
Hey @ArturoLlorente thanks for this contribution! Just to follow up; you are intending to add an example script to run the model under `graphnet/examples`, right? If so, I'll wait with reviewing until it's added.

src/graphnet/models/gnn/dynedge_kaggle_tito.py
```diff
@@ -0,0 +1,253 @@
...
+            nb_inputs: Number of input features on each node.
+            features_subset: The subset of latent features on each node that
+                are used as metric dimensions when performing the k-nearest
+                neighbours clustering. Defaults to [0,1,2].
+            layer_size_scale: Integer that scales the size of hidden layers.
+            global_pooling_schemes: The list global pooling schemes to use.
+                Options are: "min", "max", "mean", and "sum".
+        """
+        # DynEdge layer sizes
+        dynedge_layer_sizes = [(256, 256) for layer in range(layer_size_scale)]
```
RasmusOrsoe, line 56: This seems to define both the latent dimensions and number of layers in your DynTrans convolution, so I think the variable should be named `dyntrans_layer_sizes` instead. Also, if you look at the implementation of DynEdge [here](https://github.com/graphnet-team/graphnet/blob/eaa122a75f0b86e5c9e096e681125178cae76875/src/graphnet/models/gnn/dynedge.py#L32), then we allow for the layer sizes to be given as an input, where the number of layers is inferred based on this input. I think it would be ideal to do something similar here, so one can adjust these hyperparameters. So instead of specifying `layer_size_scale: int = 4`, the user could specify `dyntrans_layer_sizes: Optional[List[Tuple[int, ...]]] = None`, and if the parameter is `None`, it could default to the choice you have written here.

src/graphnet/models/components/layers.py
```diff
@@ -59,3 +65,140 @@ def forward(
...
+    """Implementation of dynTrans1 layer used in TITO solution for.
+
+    'IceCube - Neutrinos in Deep' kaggle competition.
+    """
+
+    def __init__(
+        self,
+        layer_sizes: Optional[List[int]] = None,
+        aggr: str = "max",
+        nb_neighbors: int = 8,
```
RasmusOrsoe, line 123: It appears that `nb_neighbors` is unused. Am I missing something?

src/graphnet/models/gnn/dynedge_kaggle_tito.py
```diff
@@ -0,0 +1,253 @@
...
+        self._readout = torch.nn.Sequential(*readout_layers)
+
+        # Transformer layer(s)
+        encoder_layer = TransformerEncoderLayer(
+            d_model=last_posting_layer_output_dim,
+            nhead=8,
+            batch_first=True,
+            norm_first=False,
+        )
+        self._transformer_encoder = TransformerEncoder(
```
RasmusOrsoe, line 168: Because `num_layers = 0`, `self._transformer_encoder` is actually set to 
```
<bound method Module.state_dict of TransformerEncoder(
  (layers): ModuleList()
)>
```
which looks like an empty `ModuleList`, so no learnable weights. Is this intentional? It appears to me that this is redundant code. Also, the diagram shown in the Kaggle solution doesn't appear to have a transformer readout-layer, which this would be, so I think this was his way of turning it off. :-)

src/graphnet/models/gnn/dynedge_kaggle_tito.py
```diff
@@ -0,0 +1,253 @@
...
+            torch.log10(data.n_pulses),
+        )
+
+        # DynEdge-convolutions
+        for conv_layer in self._conv_layers:
+            x, _edge_index = conv_layer(x, data.edge_index, batch)
+
+        # Transformer convolutions
+        x, mask = to_dense_batch(x, batch)
+        x = self._transformer_encoder(x, src_key_padding_mask=mask)
```
RasmusOrsoe, line 234: See comment above; `self._transformer_encoder` is set to an empty module list. Are you sure this is needed?

src/graphnet/models/components/layers.py
```diff
@@ -59,3 +65,140 @@ def forward(
...
+
+        # Recompute adjacency
+        edge_index = None
+
+        # Transformer layer
+        x, mask = to_dense_batch(x, batch)
+        x = self._transformer_encoder(x, src_key_padding_mask=~mask)
+        x = x[mask]
+
+        return x, edge_index
```
RasmusOrsoe, line 204: The new `edge_index` is set to `None`; do we even need to return it? If not, please consider removing this

RasmusOrsoe (changes requested):
Hey @ArturoLlorente thank you very much for this PR! I have left a few comments for you.

src/graphnet/models/components/layers.py
```diff
@@ -59,3 +65,140 @@ def forward(
         ).to(self.device)
 
         return x, edge_index
+
+
+class EdgeConv0(MessagePassing, LightningModule):
```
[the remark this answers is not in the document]
ArturoLlorente: That's a good suggestion, I'll change the name

src/graphnet/models/components/layers.py
ArturoLlorente: You are right, I'll remove it.

src/graphnet/models/components/layers.py
```diff
@@ -59,3 +65,140 @@ def forward(
...
+        self.features_subset = features_subset
+
+        self.norm_first = False
+
+        self.norm1 = LayerNorm(d_model, eps=1e-5)  # lNorm
+
+        # Transformer layer(s)
+        encoder_layer = TransformerEncoderLayer(
+            d_model=d_model,
+            nhead=8,
```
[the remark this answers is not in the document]
ArturoLlorente: I will implement it, but before was problematic with CodeClimate (it allows a maximum number of input variables).

src/graphnet/models/components/layers.py
ArturoLlorente: Removed, also from the GNN constructor.

src/graphnet/models/gnn/dynedge_kaggle_tito.py
ArturoLlorente: - Nice suggestion, that's a better name for the variable.
- That was my first approach, however I got CodeClimate error when performing the PR. I will leave as it is, as I think otherwise everytime anyone pull the code and then try to do a PR, it will give an error. I expect that this may change if the PR gets approved, so maybe it is better to use the original approach (_dyntrans_layer_sizes: Optional[List[Tuple[int, ...]]] = None_). Pending on your answer to proceed.

src/graphnet/models/gnn/dynedge_kaggle_tito.py
ArturoLlorente: That's a nice observation, you're right. I will delete this empty layer

src/graphnet/models/gnn/dynedge_kaggle_tito.py
RasmusOrsoe: Cool! CodeClimate is a guideline; Most of the time the suggestions are good to follow, but in some instances it's impractical and in those cases we approve the PR anyway. 

Take a look at [DynEdge](https://codeclimate.com/github/graphnet-team/graphnet/src/graphnet/models/gnn/dynedge.py/source#issue-2436ce949c626cc70f6b3b34e042d1a8) in CodeClimate. As it exists in the repo today, there are some flags raised by CodeClimate. Those are OK. So if your implementation has the same flags raised, it's OK.

If you still have the code available that lets the layer sizes be defined by the user, as discussed above, I'd suggest that you add it. If you prefer to proceed as-is, then I think we should create a feature request for this functionality and look at it later. Let me know what you prefer!

RasmusOrsoe (approved):
@ArturoLlorente Well done!

PR #520, opened 2023-05-17
move id extraction into I3FeatureExtractor

RasmusOrsoe:
This PR moves the extraction of "`string`", "`pmt_number`", "`dom_number`" and "`dom_type`" out of `I3FeatureExtractorIceCubeUpgrade` and into `I3FeatureExtractorIceCube86`.  Note that `I3FeatureExtractorIceCubeUpgrade` will still extract these variables due to inheritance.

These variables are available on regular IceCube simulation and not inherent to Upgrade. By moving them, we make sure other datasets will have it available too.


MortenHolmRep (approved):
Sounds like a reasonable change and LGTM :)

Issue #474, opened 2023-04-12
Northeren Tracks: Removing cut on pulse counts


RasmusOrsoe:
**What is benchmarked**
The `<= 900` cut on pulse count is a quick way to address the high memory requirements of applying DynEdge to high energy datasets (removes just ~2%). However, for point source searches it's likely the events with the large pulse counts that provide the most of an analysis sensitivity. Therefore, it would be ideal to find another way to lower memory usage than removing events. Therefore:

Can we get rid of the cut by adjusting the layer sizes in DynEdge?

**Target variables used for evaluation**
`'classification_emuon_entry'`

**Step-by-step**
When we construct DynEdge
```
gnn = DynEdge(
            nb_inputs=detector.nb_outputs,
            global_pooling_schemes=["min", "max", "mean"],
            )
```
the optional parameter `dynedge_layer_sizes` defaults to the rather generous` [(128, 256), (336, 256), (336, 256), (336, 256)]`. In the past we've observed that one can decrease the layer sizes quite a bit without seeing a difference in energy resolution.
1. Remove the pulse count cut on the selections.
2. Construct DynEdge but where you pass `dynedge_layer_sizes` with a significantly smaller layer sizes (try decreasing by a factor 2, 4 or 6)
3. Run the script for a few epochs using `HVInIcePulses` pulsemap; do you get memory errors?
4. If decreasing the layer sizes by a factor 6 does not help; try to pass only one global pooling scheme, e.g. `global_pooling_schemes=["max"]`. 
5. If 4 also fails, try to also decrease the batch size by a factor (2, 4, 6).
6. If you at any point succeed in getting the training to run stable for a few epochs, then let the training complete and compare the results with a converged model that was trained using the pulse count cut, and send the plot to me on slack :-)

Issue #462, opened 2023-03-21
Should we change the way Detector, Coarsening, and geometry tables work?


RasmusOrsoe:
One important part of the philosophy for Models in GraphNeT is that they are self-contained; functionality that a specific model requires (standardization, transformation and other auxiliary calculations) are available within the model such that it's deployable as a single package that only depends on data. i.e:

`event -> Model -> prediction`

Where the components that make up `Model` exists in the sequence `[Coarsening, Detector, GNN, Task]`.

The `Detector` class is supposed to represent all the detector-specific code - and the exchange of one detector class with another is supposed to be the only difference between training a Model on IceCube vs. any other experiment. Over the past year, the Detector class is the component that has seen the least of our attention, and I've noticed a few issues with the current implementation. Here is what I've noticed:

- Way back when, we used `sklearn.preprocessing` for standardization of data, but had to let this go because of efficiency and elegance. This need as occasionally resurfaces. The class still contains old lines of code that dates back to this. It's a mess (that I made).
- The Detector contains hard-coded standardization that is perhaps not ideal for all data that the class represents, sparking the need to create multiple versions of the same detector.
- Each Detector class expects a **specific** set of input features; if they differ in name or number the code fails. This again sparks the need to create multiple copies of each detector, because if one wants to train only on a subset of features that's associated with the detector, the assertion fails. 
- Because Coarsening can change both the meaning of input variables and the number of them, the current placement of Coarsening will provoke errors mentioned above. 

None of the problems above are unsolvable; the checks and balances on input dimensions and feature names exists for a good reason, and we could just redesign the implementation to be more flexible, change the order of `Coarsening` and `Detector` and those of us requiring external libraries for standardization could just eat the efficiency loss. **However**,  the path ahead becomes more opaque when taking into account new functionality; **Coarsening** & **Geometry Tables** (#421). The observations I have are:

**Coarsening**
- Our current implementation of `Coarsening` is not easily understood - it has to take in batched data and aggregated correctly in vectorized form - all of which makes it naturally hard to follow. I fear that it might become difficult to extend it much further if future needs arise. But it works and gives self-contained models. 
[...]

PR #521, opened 2023-05-19
The implementation of the RNN_DynEdge model

Aske-Rosted:
An implementation of an model that first puts the DOM data through an RNN network before feeding it to the DynEdge network, along with some functions working on DOM level data in order to facilitate the RNN. 

This is a way of tackling the issue which is also addressed in #474.

Due to limited hardware on my end I have not had the ability to properly performance check this new model.

### DIFF
--- a/src/graphnet/data/dataset.py
+++ b/src/graphnet/data/dataset.py
@@ -17,6 +17,8 @@ from typing import (
 import numpy as np
 import torch
 from torch_geometric.data import Data
+from torch.utils.data import TensorDataset
+
 
 from graphnet.constants import GRAPHNET_ROOT_DIR
 from graphnet.data.utilities.string_selection_resolver import (
@@ -150,6 +152,7 @@ class Dataset(Logger, Configurable, torch.utils.data.Dataset, ABC):
         loss_weight_column: Optional[str] = None,
         loss_weight_default_value: Optional[float] = None,
         seed: Optional[int] = None,
+        timeseries: Optional[bool] = None,
     ):
         """Construct Dataset.
 
@@ -195,6 +198,7 @@ class Dataset(Logger, Configurable, torch.utils.data.Dataset, ABC):
                 subset of events when resolving a string-based selection (e.g.,
                 `"10000 random events ~ event_no % 5 > 0"` or `"20% random
                 events ~ event_no % 5 > 0"`).
+            timeseries: Whether the dataset is a timeseries dataset. Defaults to None (Not a timeseries dataset).
         """
         # Base class constructor
         super().__init__(name=__name__, class_name=self.__class__.__name__)
@@ -218,6 +222,7 @@ class Dataset(Logger, Configurable, torch.utils.data.Dataset, ABC):
         self._index_column = index_column
         self._truth_table = truth_table
         self._loss_weight_default_value = loss_weight_default_value
+        self._timeseries = timeseries
 
         if node_truth is not None:
             assert isinstance(node_truth_table, str)
@@ -375,7 +380,14 @@ class Dataset(Logger, Configurable, torch.utils.data.Dataset, ABC):
         features, truth, node_truth, loss_weight = self._query(
             sequential_index
         )
-        graph = self._create_graph(features, truth, node_truth, loss_weight)
+        if self._timeseries is not False:
+            graph = self._create_timeseries(
+                features, truth, node_truth, loss_weight
+            )
+        else:
+            graph = self._create_graph(
+                features, truth, node_truth, loss_weight
+            )
         return graph
 
     # Internal method(s)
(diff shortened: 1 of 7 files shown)


src/graphnet/data/dataset.py
```diff
@@ -150,6 +152,7 @@ def __init__(
         loss_weight_column: Optional[str] = None,
         loss_weight_default_value: Optional[float] = None,
         seed: Optional[int] = None,
+        timeseries: Optional[bool] = None,
```
RasmusOrsoe, line 155: It seems like` timeseries` is a boolean that toggles between our "default" graph definition and your problem/model specific graph definition. I think this is a good clue that we should introduce a more modular way of choosing graph representations that doesn't make the `Dataset` inflated with arguments. I think the `GraphDefinition` mentioned in #462 is what we need now. I think we should introduce one argument to `Dataset` called `graph_definition: GraphDefinition`. This module should be in charge of 

1. Standardizing/scaling of node features (Modular)
2. Creation of `Data` object (Problem specific - new implementation for new problem)
3. All changes/attachments/adjustments made to graph object in `Dataset` (functionality in base class)
4. Assignment of edges from` EdgeDefinition `  (Modular)

 Here is some very specific pseudo-code of how `GraphDefinition` could look like:

```
from typing import Tuple, Any, List, Optional, Union, Dict, Callable
from abc import ABC, abstractmethod
import torch
from torch_geometric.data import Data
import numpy as np

class GraphDefinition(Model):
    """
    An Abstract class to create graph definitions from.
    """
    @save_model_config
    def __init__(self,
                 preprocessing_module: Optional[PreprocesingModule],
                 edge_definition: Optional[EdgeDefinition]):
        
        # Member Variables
        self._preprocessing = preprocessing_module
        self._edge_definiton = edge_definition

        # Base class constructor

    @abstractmethod
    def _create_graph(self, x: np.array) -> Data:
        """ Problem/model specific graph definition. 
            Should not standardize/scale data.
            May assign edges.

        Args:
            x: node features for a single event

        Returns:
            Data object (a single graph)
        """

    def __call__(self, 
                 node_features: List[Tuple[float, ...]],
                 node_feature_names: List[str],
                 truth_dicts: List[Dict[str, Any]],
                 custom_label_functions: List[Dict[str, Callable]],
                 loss_weight_column: Optional[str] = None,
                 data_path: Optional[str] = None,
                 ) -> Data:
        
        # Standardize / Scale  node features
        node_features = self._preprocessing(node_features, node_feature_names)

        # Create graph 
        graph = self._create_graph(node_features)

        # Attach number of pulses as static attribute.
        graph.n_pulses = torch.tensor(len(node_features), dtype=torch.int32)

        # Assign edges
        graph = self._edge_definiton(graph)

        # Attach data path - useful for Ensemble datasets.
        if data_path is not None:
            graph['dataset_path'] = data_path

        # Attach loss weights if they exist
        graph = self._add_loss_weights(graph = graph,
                                       loss_weight = loss_weight,
                                       loss_weight_column=loss_weight_column,
                                       loss_weight_default_value = loss_weight_default_value)
        
        # Attach default truth labels and node truths
        graph = self._add_truth(graph = graph,
                                truth_dicts = truth_dicts)
        
        # Attach custom truth labels
        graph = self._add_custom_labels(graph = graph,
                                        custom_label_functions=  custom_label_functions)
        
        # Attach node features as seperate fields. MAY NOT CONTAIN 'x'
        graph = self._add_features(graph = graph,
                                   node_feature_names = node_feature_names)
    
        return graph
    
    def _add_loss_weights(self,
                          loss_weight: np.array,
                          graph: Data,
                          loss_weight_column: str,
                          loss_weight_default_value: Union[int,float],
                          ) -> Data:
        """Attempt to store a weight in the graph that can be used to weight 
            the loss during training. I.e. 
            `graph[loss_weight_column] = loss_weight`

        Args:
            loss_weight: The non-negative weight to be stored.
            graph: Data object representing the event.
            loss_weight_column: The name under which the weight is stored in
                                 the graph.
            loss_weight_default_value: The default value used if 
                                        none was retrieved.

        Returns:
            A graph with loss weight added, if available.
        """
        # Add loss weight to graph.
        if loss_weight is not None and loss_weight_column is not None:
            # No loss weight was retrieved, i.e., it is missing for the current
            # event.
            if loss_weight < 0:
                if loss_weight_default_value is None:
                    raise ValueError(
                        "At least one event is missing an entry in "
                        f"{loss_weight_column} "
                        "but loss_weight_default_value is None."
                    )
                graph[loss_weight_column] = torch.tensor(
                    self._loss_weight_default_value, dtype=self._dtype
                ).reshape(-1, 1)
            else:
                graph[loss_weight_column] = torch.tensor(
                    loss_weight, dtype=self._dtype
                ).reshape(-1, 1)
        return graph
    
    def _add_truth(self, graph: Data, 
                          truth_dicts: List[Dict[str, Any]]
                          ) -> Data:
        """Add truth labels from ´truth_dicts´ to ´graph´. I.e.:
            ´graph[key] = truth_dict[key]´


        Args:
            graph: graph where the label will be stored
            truth_dicts: dictionary containing the labels

        Returns:
            graph with labels
        """
        # Write attributes, either target labels, truth info or original
        # features.
        for truth_dict in truth_dicts:
            for key, value in truth_dict.items():
                try:
                    graph[key] = torch.tensor(value)
                except TypeError:
                    # Cannot convert `value` to Tensor due to its data type,
                    # e.g. `str`.
                    self.debug(
                        (
                            f"Could not assign `{key}` with type "
                            f"'{type(value).__name__}' as attribute to graph."
                        )
                    )
        return graph

    def _add_features_individually(self,
                      graph: Data,
                      node_feature_names: List[str],
                      ) -> Data:
        # Additionally add original features as (static) attributes
        graph.features = node_feature_names
        for index, feature in enumerate(node_feature_names):
            if feature not in ["x"]: # reserved for node features.
                graph[feature] = graph.x[:, index].detach()
            else:
                self.warnonce(f"""Cannot assign graph['x']. This field is 
                              reserved for node features. Please rename your 
                              input feature.""")
        return graph
    
    def _add_custom_labels(self, graph: Data,
                           custom_label_functions: Dict[str, Callable]
                           ) -> Data:
        # Add custom labels to the graph
        for key, fn in custom_label_functions.items():
            graph[key] = fn(graph)
        return graph
```

Whats missing here is `PreprocesingModule` (Essentially a refactor of `Detector` that only does scaling/standardization). Ideally, this class should be able to deal with graphs that are created on a subset of the features that the module preprocesses. E.g. for IceCube86, which has features` ["dom_x","dom_y", "dom_z", "dom_time", "charge","rde","pmt_area"]` it should be possible to pass graphs with `features = ["dom_x", "dom_y", "dom_z", "dom_time"]` as this is a legit use case (right now it just throws a tantrum).

`EdgeDefinition` is our `GraphBuilder` class, which I think we should just rename and keep as-is.

Our "default" graph would then be defined as:

```
class DefaultGraph(GraphDefinition):
    def _create_graph(self, node_features: np.array, dtype: torch.dtype) -> Data:
        return Data(x= torch.tensor(node_features, dtype=dtype), edge_index=None)
```
Where it can then be instantiated like

```
graph_definition = DefaultGraph(edge_definition = KNNGraph,
                                                     preprocessing_module = IceCube86)
```

This way, all the code that alters our `graph` in `Dataset` has been moved into one self-consistent class. In `Dataset.create_graph ` the code is now simpler:

```
def _create_graph(
        self,
        features: List[Tuple[float, ...]],
        truth: Tuple[Any, ...],
        node_truth: Optional[List[Tuple[Any, ...]]] = None,
    ) -> Data:
        """Create Pytorch Data (i.e. graph) object.

        Args:
            features: List of tuples, containing event features.
            truth: List of tuples, containing truth information.
            node_truth: List of tuples, containing node-level truth.
            loss_weight: A weight associated with the event for weighing the
                loss.

        Returns:
            Graph object.
        """
        # Convert nested list to simple dict
        truth_dict = {
            key: truth[index] for index, key in enumerate(self._truth)
        }

        # Define custom labels
        labels_dict = self._get_labels(truth_dict)

        # Convert nested list to simple dict
        if node_truth is not None:
            node_truth_array = np.asarray(node_truth)
            assert self._node_truth is not None
            node_truth_dict = {
                key: node_truth_array[:, index]
                for index, key in enumerate(self._node_truth)
            }
        
        # Create list of truth dicts with labels
        truth_dicts = [labels_dict, truth_dict]
        if node_truth is not None:
            truth_dicts.append(node_truth_dict)

        # Catch cases with no reconstructed pulses
        if len(features):
            node_features= np.asarray(features)[:, 1:]
        else:
            node_features = np.array([]).reshape((0, len(self._features) - 1))

        # Construct graph data object
        graph = self._graph_definition(node_features = node_features,
                                        node_feature_names =  self._features[1:], # first entry is index column
                                        truth_dicts = truth_dicts,
                                        custom_label_functions = self._label_fns,
                                        loss_weight_column = self._loss_weight_column,
                                        loss_weight = loss_weight,
                                        loss_weight_default_value = self._loss_weight_default_value,
                                        data_path = self._path)
        return graph
```

Similar changes will have to be made to [make_graph](https://github.com/graphnet-team/graphnet/blob/1abb441ef3c3dd84e8624e0deb5e1d50845d1379/src/graphnet/deployment/i3modules/graphnet_module.py#L76) in `GraphNeTI3Module`. 

When we then construct `Models`, we would then pass `GraphDefinition` instead of `Detector`, and we could add a `self.warnonce` if the model is being evaluated on graphs that it was not trained on. Ideally, the graph definition that a model was trained on should be available via the model (e.g. `my_model.graph_definition( .. )` ) once its saved and loaded into a new session, like **asogaard** drew in the diagrams in #462, such that it's available when the model is deployed.

PR #556, opened 2023-07-07
Fixed typos in docstrings

AMHermansen:
Fixed some minor typos in the docstring for a couple of functions in the dataset.py file

### DIFF
--- a/src/graphnet/data/dataset.py
+++ b/src/graphnet/data/dataset.py
@@ -382,7 +382,7 @@ class Dataset(Logger, Configurable, torch.utils.data.Dataset, ABC):
     def _resolve_string_selection_to_indices(
         self, selection: str
     ) -> List[int]:
-        """Resolve selection as string to list of indicies.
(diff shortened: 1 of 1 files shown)


AMHermansen:
It seems that there were some formatting errors in the file prior to this PR, should I also update the formatting, or do we leave it as is @RasmusOrsoe

RasmusOrsoe:
Hey @AMHermansen. Thanks for correcting the typos! 

The code quality checks must pass. We did change between formats of doc strings some time ago, perhaps that is what caused this. In either case, it would be much appreciated if you would polish this such that the checks pass.

AMHermansen:
Hello @RasmusOrsoe 

It seems that when I was correcting a typo I exceed the length limit in one of the descriptions, and I believe that descriptions are limited to a single line. Hence I've reworded the troubling line. While doing so I also noticed some more typos, which I've also corrected.

RasmusOrsoe (approved):
Very nice! Thanks @AMHermansen

PR #558, opened 2023-07-18
Graph Definition

RasmusOrsoe:
This PR addresses the ongoing discussion in #462 and #521 (which has been a roadblock for a while) by changing `Model` such that it now consists of the modules

`Model = [GraphDefinition, GNN, Task]`

Where `GraphDefinition` is a single, problem/model dependent class that contains all the code responsible for data representations. 

- [x] Add GraphDefinition
- [x] Implement our default k-nn graph as importable graph definition
- [x] Redefine Model to depend on GraphDefinition
- [x] Refactor Detector
- [x] Update example scripts
- [x] Update config files
- [x] Consider default values
- [x] Delete redundant modules; GraphBuilder etc.
- [x] Update getting_started.md
- [x] Refactor Deployment modules

**TLDR: Model, Dataset and GraphNeTI3Module now depends on GraphDefinition, which allows us to easily represent data as sequences, images, or whatever your heart desires. This change is breaking; older config files and pickled models are not compatible with these changes, but state_dicts are.** 

Conceptually, `GraphDefinition` contains all the code that alters the raw data from `Dataset` before it's passed to `GNN`. It's a single, swapable module that can be passed to `Dataset` and deployment modules. `GraphDefinition` consists of multiple submodules, and the data flow is `GraphDefinition = [Detector, NodeDefinition, EdgeDefinition]` and can be seen [here](https://github.com/graphnet-team/graphnet/blob/5dc129829f2377ebf00ebe12084b4f23fa1ab5fd/src/graphnet/models/graphs/graph_definition.py#L73). **The definition exists at a point in the dataflow where events are unbatched, meaning that the construction of data representations can be done on CPU and in parallel, before it's batched and sent to the GPU**. That means that the sequence creation included in #521 becomes much simpler and likely faster @Aske-Rosted, and should also be useful for the transformer exploration by @MoustHolmes. 

The modules are defined as

**NodeDefinition** :
A generic class that defines what a node represents. Problem-specific versions can be implemented by overwriting the abstract method 
```
def _construct_nodes(self, x: torch.tensor) -> Data:
        """Construct nodes from raw node features ´x´.

        Args:
            x: standardized node features with shape ´[num_pulses, d]´,
            where ´d´ is the number of node features.

        Returns:
            graph: graph without edges.
        """
```
`_construct_nodes` is the playground we've been missing for a while; it gives us the freedom to fully define exactly how we want the data to be structured for our Models. Here, one can use nodes to represent DOMs (by using `Coarsening` or some other method), create images for CNNs, define sequences or other forms of data representations. Our standard of representing pulses as nodes is just
```
class NodesAsPulses(NodeDefinition):
    """Represent each measured pulse of Cherenkov Radiation as a node."""

    def _construct_nodes(self, x: torch.tensor) -> Data:
        return Data(x=x)
```

**EdgeDefinition**: 
 A generic class that defines how edges are drawn between nodes in the graph. This is essentially a refactor of our `GraphBuilder`. One can create problem-specific implementations by overwriting the abstract method

```
def _construct_edges(self, graph: Data) -> Data:
        """Construct edges and assign them to graph. I.e. ´graph.edge_index = edge_index´.

        Args:
            graph: graph without edges

        Returns:
            graph: graph with edges assigned.
        """
```

**Detector**:
 Virtually unchanged from it's known form. In charge of standardizing data and is now able to work on a subset of the feature space that it is defined on. I cleaned the class up a little bit. In the future, it will hold detector-specific geometry tables as mentioned in #462. 

Our usual k-nn graph with nodes representing pulses can then be created like so:
```
from graphnet.models.graphs import GraphDefinition
from graphnet.models.graphs.nodes import NodesAsPulses
from graphnet.models.graphs.edges import KNNEdges
from graphnet.models.detector.prometheus import Prometheus

graph_definition = GraphDefinition(node_definiton = NodesAsPulses(nb_nearest_neighbours=8),
                                   edge_definiton = KNNEdges(),
                                   detector = Prometheus(),
                                     )
                               
```
Alternatively, you can also just import this graph definition directly, as it's included in the PR:

```

graph_definition = KNNGraph(
        detector=Prometheus(),
        node_definition=NodesAsPulses(),
        nb_nearest_neighbours=8,
        node_feature_names=features,
    )
```

It is the problem-specific implementation of a graph definition that defines the number of input parameters to our GNNs, available through `graph_definition.nb_outputs`. When we instantiate a Model, the syntax is now:

```
 gnn = DynEdge(
        nb_inputs=graph_definition.nb_outputs,
        global_pooling_schemes=["min", "max", "mean", "sum"],
    )
    task = ..
model = StandardModel(
        graph_definition=graph_definition,
        gnn=gnn,
        tasks=[task],
        ...)
```

**Other things to note:**

1. `Dataset` is now simpler, as graph-altering code has been moved to `GraphDefinition`
2. Changes are compatible with the Config system.
3. Some folder structure has been changed in `graphnet.data` to avoid circular imports.


RasmusOrsoe:
I'm sorry for the size of this PR; it wasn't practical to divide this into multiple, smaller PRs. The real change is the introduction of `GraphDefinition`; this code lives in `graphnet.models.graphs`. Other changes include `graphnet.models.detector` and `graphnet.data.dataset`.  

I think the state of this PR is mature enough for you to get an idea of where this draft is heading. A few superficial steps are missing, but I'd prefer to complete this after you've had a chance to comment on the general idea. 

@Aske-Rosted @MoustHolmes @AMHermansen please take some time to review this!

AMHermansen (commented):
Thank you for taking the time, to make these changes. I've left some comments, mostly minor stuff.

I believe that moving forward the code would become clearer if we removed magic numbers from the pre-processing methods, and used class-variables instead. This would also allow us to more easily change detector specific transformations in the future, by implementing setter functions for the class variables, instead of being forced to make changes to the actual code-base.

I had some difficulties with GitHub, so specific comments are below.

src/graphnet/models/detector/icecube.py
AMHermansen, line 1: Instead of keeping magic-numbers in the transformation, the code would become cleaner by introducing class variables for the detector, and it would make it easier in the future to change how the preprocessing transformations of a detector, by implementing a setter function, or even take it as a default argument in the constructor.

The code would look something like this then:
`class IceCube86(Detector):
     _time_offset = 1.0e04
     _time_scale = 3.0e4

     def _dom_time(self, x: torch.tensor) -> torch.tensor:
           return (x - self._time_offset) / self._time_scale`

And similarly for the other transformations.

src/graphnet/models/detector/icecube.py
```diff
@@ -15,32 +16,33 @@
...
+            "rde": self._rde,
+            "pmt_area": self._pmt_area,
+        }
+        return feature_map
+
+    def _dom_xyz(self, x: torch.tensor) -> torch.tensor:
+        return x / 500.0
+
+    def _dom_time(self, x: torch.tensor) -> torch.tensor:
+        return x - 1.0e04 / 3.0e4
```
AMHermansen, line 36: It seems that you're missing a parenthesis.

src/graphnet/models/detector/prometheus.py
AMHermansen, line 1: Similar suggesting as for the other detector.

src/graphnet/models/graphs/graph_definition.py
```diff
@@ -0,0 +1,270 @@
...
+        super().__init__(name=__name__, class_name=self.__class__.__name__)
+
+        # Member Variables
+        self._detector = detector
+        self._edge_definiton = edge_definition
+        self._node_definition = node_definition
+        if node_feature_names is None:
+            # Assume all features in Detector is used.
+            node_feature_names = list(self._detector.feature_map().keys())  # type: ignore
+        print(node_feature_names)
```
AMHermansen, line 60: I'm not sure if this print statement is intentional or a debugging print statement. If it is intentional then perhaps we should include some more information like:

`print(f"The node features are {node_feature_names}")`

AMHermansen (commented):
Hello @RasmusOrsoe I like the changes proposed in this PR. I have some minor suggestions, that you can think about :)

src/graphnet/data/dataset/dataset.py
AMHermansen, line 1: I believe that the typos corrected in #556 have been reverted here.

AMHermansen (commented):
I forgot to add this comment to the previous one :)

configs/datasets/test_data_sqlite.yml
```diff
@@ -1,26 +1,34 @@
...
-truth_table: truth
-seed: 21
-selection: null
\ No newline at end of file
+loss_weight_column: null
+loss_weight_default_value: null
+loss_weight_table: null
+node_truth: null
+node_truth_table: null
+path: /home/iwsatlas1/oersoe/github/graphnet/data/examples/sqlite/prometheus/prometheus-events.db
```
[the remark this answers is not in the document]
RasmusOrsoe: Thanks for spotting that. It's been changed!

configs/datasets/dev_lvl7_robustness_muon_neutrino_0000.yml
```diff
@@ -1,4 +1,17 @@
 path: /groups/icecube/asogaard/data/example/dev_lvl7_robustness_muon_neutrino_0000.db
```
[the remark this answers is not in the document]
RasmusOrsoe: Good point. I've looked through the code base and I actually cannot find anything that depends on this config file, so I wonder why we still have it. Perhaps it's worth deleting.

examples/04_training/01_train_model.py
```diff
@@ -72,6 +72,7 @@ def main(
 
     # Construct dataloaders
     dataset_config = DatasetConfig.load(dataset_config_path)
+    print(dataset_config_path)
```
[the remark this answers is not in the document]
RasmusOrsoe: Thanks!

src/graphnet/models/graphs/graph_definition.py
RasmusOrsoe: Removed!

src/graphnet/models/detector/icecube.py
RasmusOrsoe: Thank you!

src/graphnet/models/detector/icecube.py
RasmusOrsoe: The static nature of the `Detector` is a design choice; it forces the user to implement a class if one needs a different standardization. It has its up and down sides; it can lead to multiple versions of essentially the same Detector, but it also ensures that `Models` are fully summarized by their respective components. I fear adding class variables will lead to more lines of code pr. Detector, and imply to users that its OK to overwrite member variables, which would result in configuration files not fully summarizing the `Models`. 

However, I'm very open to discussing this design choice; perhaps one could in the future refactor `Detector` such that it depends on a standardization module. This way, each `Detector` will only exist once in the code base, but one could have many different standardization modules for each `Detector`. But I think this is beyond the scope of this PR. What do you think @AMHermansen?

RasmusOrsoe:
I have addressed the initial comments from @AMHermansen, refactored the last few detectors and introduced GraphDefinition in the deployment code. In relation to this, I've created #560 which contains a list of to-do items for the deployment modules which fall outside the scope of this PR.

What's left is to update `getting_started.md` and considerations on default values. I'll wait with this last step until @Aske-Rosted has had a chance to give this a look after the ICRC conference.

configs/datasets/test_data_sqlite.yml
```diff
@@ -1,26 +1,34 @@
...
-  - azimuth
-  - zenith
-  - pid
-  - elasticity
-  - sim_type
-  - interaction_type
+features: [sensor_pos_x, sensor_pos_y, sensor_pos_z, t]
+graph_definition:
+  arguments:
+    columns: [0, 1, 2]
```
Aske-Rosted, line 4: I think the naming of this "columns" arguments (which I understand as the columns included in the "distance" calculation of your edge definition) becomes very vague. At first glance it hard to tell what this feature does. Something like "edge_defining_columns" would be more descriptive I think.

src/graphnet/data/dataset/parquet/__init__.py
```diff
@@ -0,0 +1,11 @@
...
+# Configuration
+from graphnet.utilities.imports import has_torch_package
+
+if has_torch_package():
+    import torch.multiprocessing
+    from .parquet_dataset import ParquetDataset
+
+    torch.multiprocessing.set_sharing_strategy("file_system")
+
+del has_torch_package
```
Aske-Rosted, line 11: Is there a reason for deleting the `has_torch_package` here but keeping it in `src/graphnet/data/dataset/sqlite/__init__.py`?

src/graphnet/utilities/config/model_config.py
```diff
@@ -133,7 +133,7 @@ def _construct_model(
             fn_kwargs={"trust": trust},
         )
 
-        # Construct model based on arguments
+        # Construct model based on
```
Aske-Rosted, line 136: was this change intentional?

Aske-Rosted:
I have just finished looking through everything, I think it looks great! I also believe that this will add the necessary flexibility for a lot of the things I've been trying to implement, where I previously had to go make changes in the dataset class e.g.   #521 . I am looking forward to trying it out.

src/graphnet/models/detector/detector.py
```diff
@@ -1,115 +1,53 @@
...
-                and len(set(data.features[0])) == 1
-            ):
-                data_features = [features[0] for features in data.features]
+    @final
+    def _standardize(
+        self, node_features: torch.tensor, node_feature_names: List[str]
+    ) -> Data:
+        for idx, feature in enumerate(node_feature_names):
+            try:
+                node_features[:, idx] = self.feature_map()[feature](  # type: ignore
```
AMHermansen, line 41: I'm not sure if you're supposed to call self.feature_map, since it is classified as a property?

src/graphnet/models/graphs/graph_definition.py
```diff
@@ -0,0 +1,269 @@
...
+
+        # Member Variables
+        self._detector = detector
+        self._edge_definiton = edge_definition
+        self._node_definition = node_definition
+        if node_feature_names is None:
+            # Assume all features in Detector is used.
+            node_feature_names = list(self._detector.feature_map().keys())  # type: ignore
+        self._node_feature_names = node_feature_names
+        if dtype is None:
```
AMHermansen, line 61: If I understand this code correctly you're just getting the same behavior as if you changed the default value of dtype in the constructor to torch.float instead of None. But in a slightly more more complicated way :)

AMHermansen (commented):
Hello @RasmusOrsoe I've noticed some more minor things.

src/graphnet/models/detector/icecube.py
AMHermansen: I've thought about what you've written and made a feature request #562 the proposed solution in there is still in the thought phase, and I have very limited experience with config-files, so I'm not sure if it is even feasible to do it in that way.

src/graphnet/data/dataset/parquet/__init__.py
RasmusOrsoe: I actually don't know. I did not add this line; the difference you point out here is present in the main branch currently also. I have added the `del` statement to the sqlite part now too.

src/graphnet/models/graphs/graph_definition.py
RasmusOrsoe: You're absolutely right. Fixed :-)

src/graphnet/utilities/config/model_config.py
RasmusOrsoe: Nopes. Fixed.

configs/datasets/test_data_sqlite.yml
RasmusOrsoe: I agree that the name of that argument is not unambiguous, but I also fail to find better alternatives that are not very long.  The name of this argument is the same as what we used to call it in the KNNGraphBuilder (see h[ere](https://github.com/graphnet-team/graphnet/blob/becb3e62425aa77676797dc6dfb91354d68568b1/src/graphnet/models/graph_builders.py#L25)) and I do think that the doc string for this argument (see [here](https://github.com/graphnet-team/graphnet/blob/af943658f7d6c5ec5f79617b41b751e3ebfaee6c/src/graphnet/models/graphs/graphs.py#L34)) is pretty clear. So even though it's a bit challenging to understand what that argument does when one reads the config file, I think we should keep it as-is and refer users to the docs instead (which is the intended usage anyway). Is that OK with you? @Aske-Rosted

Aske-Rosted: Completely fine.

RasmusOrsoe:
Thank you very much for your comments. I believe I have now addressed all of them, and I have updated the `GETTING_STARTED.MD ` accordingly. @Aske-Rosted @AMHermansen.

Issue #560, opened 2023-07-28
Update Deployment Modules


RasmusOrsoe:
As part of a push to version 2.0, we should update deployment functionality:

- [ ] Refactor deployment code to provide a general structure #539 
- [ ] Create a method for sharing pre-trained models for each experiment #546 
- [ ] Create unit tests of deployment code that tests core functionality and ensure that pre-trained models in deployment in analyses are supported #559

Issue #546, opened 2023-05-26
Add respository of pre-trained models, and easy means to access/download them


asogaard:
Consider using [PyTorch Hub](https://pytorch.org/hub/).

Related to #545

RasmusOrsoe:
@closed by #567

PR #567, opened 2023-08-21
Deployment update

### LINKED ISSUE #559: Create Unit Tests for Deployment modules
We should add unit tests of deployment functionality in `graphnet.deployment`.

RasmusOrsoe:
This PR addresses two of the issues mentioned in #560.

1. #546  - A series of pre-trained models used for the QUESO event selection has been added under `src/models/pretrained/icecube/QUESO`.  I have added a constant path `PRETRAINED_MODEL_DIR` which points to `src/models/pretrained` under `graphnet.constants`. 
2. #559  - A unit test have been added under `tests/deployment/queso_test.py`. Here, basic functionality of deploying `I3InferenceModule`'s and` I3PulseCleaningModule` is tested using pre-trained models for the QUESO event selection. A second unit test verifies that the predictions written to i3 files are consistent with predictions in the past (checks frame by frame predictions in `graphnet/data/tests/deployment/QUESO`. This way, we make sure that going forward, PRs will not pass if they fail basic deployment functionality or if they lead to discrepancies in predictions for these models used by our IceCube Collaborators. 

I also noticed a bug in the docker file that I think has lead to the fails we've seen lately @AMHermansen. This is hopefully fixed now. 

I was a bit torn on whether to place pretrained models under `graphnet.deployment` or `graphnet.models` . Any thoughts on this is especially welcome. :-) 

Closes #559 and #546


docker/gnn-benchmarking/apply.py
```diff
@@ -11,13 +11,19 @@
...
 
-# Constants (from Dockerfile)
-MODEL_PATH = "model.pth"
+# Constants
+MODEL_NAME = "total_neutrino_energy"
+BASE_PATH = (
+    f"{GRAPHNET_ROOT_DIR}/src/graphnet/models/pretrained/icecube/upgrade/QUESO"
+)
+MODEL_CONFIG = f"{BASE_PATH}/{MODEL_NAME}/{MODEL_NAME}_config.yml"
+STATE_DICT = f"{BASE_PATH}/{MODEL_NAME}/{MODEL_NAME}_state_dict.yml"
```
AMHermansen, line 26: I believe you used the wrong file extension for the state_dict.

examples/01_icetray/04_i3_module_in_native_icetray_example.py
```diff
@@ -71,7 +72,10 @@ def main() -> None:
     # Constants
     features = FEATURES.UPGRADE
     input_folders = [f"{TEST_DATA_DIR}/i3/upgrade_genie_step4_140028_000998"]
-    mock_model_path = f"{TEST_DATA_DIR}/models/mock_energy_model.pth"
+    base_path = f"{PRETRAINED_MODEL_DIR}/icecube/upgrade/QUESO"
+    model_name = "total_neutrino_energy"
+    model_config = f"{base_path}/{model_name}/{model_name}_config.yml"
+    state_dict = f"{base_path}/{model_name}/{model_name}_state_dict.yml"
```
AMHermansen, line 78: Same as the previous comment about state_dicts.

Aske-Rosted (approved):
Everything looks good to me.

src/graphnet/models/graphs/graph_definition.py
```diff
@@ -166,7 +166,7 @@ def _validate_input(
         for idx in range(len(node_feature_names)):
             assert (
                 node_feature_names[idx] == self._node_feature_names[idx]
-            ), """ Order of node features are not the same."""
+            ), f""" Order of node features are not the same. Got {node_feature_names} vs. {self._node_feature_names}"""
```
AMHermansen, line 169: I like including more information in this assert statement. However the additional context doesn't immediately clarify to me which name is from the config file and which is the native name of the detector.

I also think your suggested wording is misleading, since `node_feature_names` are usually the names from the detector, whereas `self._node_feature_names` are usually defined from a config file.

AMHermansen (commented):
Thanks for taking the time to improve the deployment functionality. I have some input that you can think about :smile: 

I like keeping the models in `graphnet.models`. Since it is easier to navigate to find them from a quick overview, and they might be relevant outside of deployment.

I believe state_dicts are supposed to be .pth files, but I found some places in the code, where you had them as .yml files.

I've also added a comment about the assert - error message, in `graph_definition`, which I think could be made a bit clearer.

docker/gnn-benchmarking/apply.py
RasmusOrsoe: Thanks. The dangers of copy paste...

examples/01_icetray/04_i3_module_in_native_icetray_example.py
RasmusOrsoe: Thanks. The dangers of copy paste...

src/graphnet/models/graphs/graph_definition.py
RasmusOrsoe: This assert statement tries to catch cases where the `features` argument to the `Dataset` is of a different ordering than the `features` argument given to the `GraphDefinition`. Maybe we should consider removing the `features` argument for the `Dataset` class and simply let it extract it from the `GraphDefinition`. Anyway - I've updated error message to be a bit clearer.

AMHermansen (approved):

PR #578, opened 2023-09-05
Add check for torch threads in deployer

RasmusOrsoe:
PyTorch Commands  `torch.set_num_interop_threads()` and `torch.set_num_threads()` can only be set once in the same session, otherwise an error is thrown. That means that `GraphNeTI3Deployer` can only be initialized once in the same session. This PR introduces a check to avoid this error, such that `GraphNeTI3Deployer` multiple times in the same session.


AMHermansen (approved):
This looks like a pretty straight forward change. I don't see any issues with it.

PR #584, opened 2023-09-08
automatic attribute length adjustment

RasmusOrsoe:
This small PR removes the need for arguments `index_column` and `node_level` to `model.predict_as_dataframe` by replacing the if-else statements with an automated check that is able to adjust attribute sequence length to match the prediction lengths. 

With this change, `model.predict_as_dataframe` is able to automatically adjust the length of all additional attributes for both node-level and graph-level predictions.


AMHermansen (approved):
Looks good

Issue #573, opened 2023-09-01
Use numpy random generators instead of legacy numpy random


AMHermansen:
We are currently using the legacy random number generation from numpy. [See here](https://numpy.org/doc/stable/reference/random/legacy.html#legacy), while it doesn't appear that numpy has intentions of removing this code, it would be better if we moved to the supported way of generating random numbers [See here](https://numpy.org/doc/stable/reference/random/generator.html#numpy.random.Generator).

I can't quite tell when `numpy.random.RandomState` was deprecated, but the new way is implemented in all of our required numpy versions.

One example where deprecated random number generation is used:
training/utils.py line 118.

AMHermansen:
Closed by #585

PR #585, opened 2023-09-09
(Issue #573) Update Random Number Generation to Use NumPy Random Generators

samadpls:
This pull request addresses issue #573 by updating the random number generation in the codebase. It replaces the deprecated use of `numpy.random.RandomState` with the recommended approach of using NumPy random generators (`numpy.random.Generator`).

### DIFF
--- a/src/graphnet/data/dataset/sqlite/sqlite_dataset_perturbed.py
+++ b/src/graphnet/data/dataset/sqlite/sqlite_dataset_perturbed.py
@@ -3,6 +3,7 @@
 from typing import Dict, List, Optional, Tuple, Union
 
 import numpy as np
+from numpy.random import default_rng, Generator
 import torch
 from torch_geometric.data import Data
 
@@ -36,6 +37,7 @@ class SQLiteDatasetPerturbed(SQLiteDataset):
         loss_weight_table: Optional[str] = None,
         loss_weight_column: Optional[str] = None,
         loss_weight_default_value: Optional[float] = None,
+        seed: Optional[Union[int, Generator]] = None, 
     ):
         """Construct SQLiteDatasetPerturbed.
 
@@ -78,6 +80,10 @@ class SQLiteDatasetPerturbed(SQLiteDataset):
                 in this case to events with no value in the corresponding
                 table/column. That is, if no per-event loss weight table/column
                 is provided, this value is ignored. Defaults to None.
+            seed: Optional seed for random number generation (int or numpy Generator). 
+                If provided, it will be used to initialize the random number generator 
+                for data perturbation. Defaults to None.
+ 
         """
         # Base class constructor
         super().__init__(
@@ -108,6 +114,16 @@ class SQLiteDatasetPerturbed(SQLiteDataset):
             self._features.index(key) for key in self._perturbation_dict.keys()
         ]
 
+        if seed is not None:
+            if isinstance(seed, int):
+                self.rng = default_rng(seed)
+            elif isinstance(seed, Generator):
+                self.rng = seed
+            else:
+                raise ValueError("Invalid seed. Must be an int or a numpy Generator.")
+        else:
+            self.rng = default_rng()
+
     def __getitem__(self, sequential_index: int) -> Data:
         """Return graph `Data` object at `index`."""
         if not (0 <= sequential_index < len(self)):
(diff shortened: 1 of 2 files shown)


AMHermansen (commented):
Hello @samadpls :wave: thank you for this PR.

By searching for np.random in the GitHub repository I found at least one other instance where deprecated random number generation is used: `data/dataset/sqlite/sqlite_dataset_perturbed.py`. Line 130

Since this PR aims to remove the legacy random number generation from numpy to the new supported version I think we should also update how it's done in `data/dataset/sqlite/sqlite_dataset_perturbed.py`, possibly by adding a new optional seed or a `np.Generator` argument in the constructor. :smile: 
Let me know what you think

samadpls:
> Hello @samadpls 👋 thank you for this PR.
> 
> By searching for np.random in the GitHub repository I found at least one other instance where deprecated random number generation is used: `data/dataset/sqlite/sqlite_dataset_perturbed.py`. Line 130
> 
> Since this PR aims to remove the legacy random number generation from numpy to the new supported version I think we should also update how it's done in `data/dataset/sqlite/sqlite_dataset_perturbed.py`, possibly by adding a new optional seed or a `np.Generator` argument in the constructor. 😄 Let me know what you think

Hi @AMHermansen  👋, thank you for the comment, I've added an optional seed argument in the constructor. Let me know if you have any further suggestions

AMHermansen (approved):
Looks good :smiley:

src/graphnet/data/dataset/sqlite/sqlite_dataset_perturbed.py
```diff
@@ -78,6 +80,10 @@ def __init__(
                 in this case to events with no value in the corresponding
                 table/column. That is, if no per-event loss weight table/column
                 is provided, this value is ignored. Defaults to None.
+            seed: Optional seed for random number generation (int or numpy Generator).
```
AMHermansen, line 83: It seems the lines in the docstring are too long. Since we're using type hints it is not necessary to include the allowed values in the docstring.

src/graphnet/training/utils.py
```diff
@@ -115,7 +115,7 @@ def make_train_validation_dataloader(
 ) -> Tuple[DataLoader, DataLoader]:
     """Construct train and test `DataLoader` instances."""
     # Reproducibility
-    rng = np.random.RandomState(seed=seed)
+    rng = np.random.default_rng(seed=seed)
```
AMHermansen, line 118: It looks like the "rng" is passed to `sklearn.model_selection.train_test_split` and `pandas.DataFrame.sample`. Maybe we should just pass the seed value directly to sklearn.model_selection.train_test_split?

AMHermansen (changes requested):
@samadpls I stand corrected by the unittests / code styling.
I've left some comments which should hopefully make the tests pass, you could try and install pre-commit hooks as well to make sure the code quality passes :smiley:

src/graphnet/training/utils.py
samadpls: Do you mean changing `random_state=rng` to `random_state=seed`? If yes, then we can also remove `rng` as it will not be used anywhere.

AMHermansen: I thought about passing the `np.Generator ` to `pd.DataFrame.sample` and the seed or state to `sklearn.model_selection.train_test_split`, but I don't see any issue passing the seed to both.

samadpls: I apologize for misunderstanding of your comment.

AMHermansen:
@RasmusOrsoe Codeclimate is currently failing for this PR due to too high "Cognitive Complexity" should we ignore it here, or does it need refactoring?

samadpls:
> @RasmusOrsoe Codeclimate is currently failing for this PR due to too high "Cognitive Complexity" should we ignore it here, or does it need refactoring?

not sure if adding `seed` in the parameter to the `super().__init__` fix it reduce the cognitive complexity of the code

RasmusOrsoe:
Hello @samadpls - thank you for this contribution!

Automated checks are running; let's see what they say.

The codeclimate issue is not a problem @AMHermansen. We can merge if the rest is OK.

AMHermansen (approved):
Looks good

PR #586, opened 2023-09-09
Fix type hint

AMHermansen:
np.array and torch.tensor are factory functions and not the actual classes. This PR changes the type hint to correctly refer to the classes.


RasmusOrsoe (approved):
Thank you for this!

PR #591, opened 2023-09-13
Add meta class to handle config saving

AMHermansen:
Currently we're using a decorator around every `__init__` of classes that inherits from `Model/Dataset`. This creates redundant boilerplate code. This PR aims to introduce a new way of handling the backend config saving. This is done by creating metaclasses responsible for intercepting the objects created and store a corresponding config dict to the created objects.

The following is done:
1. Create relevant metaclasses (`ModelConfigSaverMeta`, `DatasetConfigSaverMeta`) along with common interfaces between these metaclasses and `ABCMeta`.
2. Remove usage of `save_model_config` and `save_dataset_config` and add deprecation warnings when they're used in code.

This means the config saving is purely done in the backend, away from any frontend users, making the code easier to read and avoids potential slip-ups where one might forget to use `@save_model_config`.

This PR currently introduces a bug. I suspect the bug is a conflict with #584.

Could you maybe try to take a look at this @RasmusOrsoe

### DIFF
--- a/src/graphnet/data/dataset/dataset.py
+++ b/src/graphnet/data/dataset/dataset.py
@@ -27,7 +27,7 @@ from graphnet.training.labels import Label
 from graphnet.utilities.config import (
     Configurable,
     DatasetConfig,
-    save_dataset_config,
+    DatasetConfigSaverABCMeta,
 )
 from graphnet.utilities.config.parsing import traverse_and_apply
 from graphnet.utilities.logging import Logger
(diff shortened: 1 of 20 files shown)


AMHermansen:
Bug resolved by #593
I've also decided to keep `save_model_config` and `save_dataset_config` but add deprecation warnings when their used. This should keep backwards compatibility.

AMHermansen:
mypy is only failing for code, which wasn't modified by this PR.

RasmusOrsoe:
@AMHermansen Let's take a look at this together after the dev meeting, if you have time.

AMHermansen:
@RasmusOrsoe I've now changed the typehints from the source code, that was conflicting with mypy (except for the SQLitePertubedDataset). This means that currently it is only the examples which are failing for mypy, feel free to have a look at this PR when you get the time :smile: .

PR #593, opened 2023-09-14
Fix #592

### LINKED ISSUE #592: Example script terminates with ValueError due to incorrect extension of additional attributes.
**Describe the bug**
Example script terminates with ValueError

which is due to to automatically extending of additional attributes, where it seems to not have been properly checked for event vs. node level prediction. 

**To Reproduce**
fresh graphnet install 
run examples/04_training/01_train_model.py

**Expected behavior**

**Full traceback**
```
ValueError: all the input array dimensions except for the concatenation axis must match exactly, but along dimension 0, the array at index 0 has size 191 and the array at index 1 has size 22089
```

**Additional context**
I will commit a fix ASAP.

Aske-Rosted:
Tries to fix the issue described in #592 @RasmusOrsoe can you check that this fix does not interfere with a node level prediction task?

### DIFF
--- a/src/graphnet/models/model.py
+++ b/src/graphnet/models/model.py
@@ -227,6 +227,7 @@ class Model(Logger, Configurable, LightningModule, ABC):
         attributes: Dict[str, List[np.ndarray]] = OrderedDict(
             [(attr, []) for attr in additional_attributes]
         )
+
         for batch in dataloader:
             for attr in attributes:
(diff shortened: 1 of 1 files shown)


RasmusOrsoe (approved):
Shoot.. Thanks for spotting this. We really do need to cover this part of the code base better with unit tests.

PR #595, opened 2023-09-16
docs: Updated `_get_all_indices` docstring

### LINKED ISSUE #497: Clarify dataset function name
In graphnet/src/graphnet/data/dataset.py is written:
`def _get_all_indices(self) -> List[int]:
        """Return a list of all available values in 'self._index_column'."""`

In some datasets the events may have multiple rows describing the same event and therefore containing the same event_id. 
To avoid confusion , it may be good to include the keyword 'unique'.

samadpls:
Closes #497
Updated the docstring for the `_get_all_indices` method in `dataset.py` to clarify that it returns unique values from `self._index_column`.


AMHermansen (approved):
Looks good :raised_hands:

PR #597, opened 2023-09-19
Filter mask

Aske-Rosted:
adds the option of using I3 files FilterMask to implement a filter when converting files


RasmusOrsoe:
Hey @Aske-Rosted thanks for this.

Could we add a check before attempting to access the filter quantities that checks it's actually there? Because if its not, the conversion will break. Here is some pseudo-code:

```
def _filter_mask(
        self, frame: "icetray.I3Frame", I3filters: List[str]
    ) -> bool:
        """Check if specified condition(s) are met.

        Args:
            frame: I3Frame to check.
            I3filters: List of I3Filters to check for pass.
        """
        if "FilterMask" in frame:
              for filter in I3filters:
                  if filter in frame["FilterMask"]:
                          if frame["FilterMask"][filter].condition_passed is False:
                              return True
                  else:
                         self.warning_once(f"Filter '{filter}' not in frame. Filter condition cannot be checked.")
              return False
       else:
             self.warning_once("FilterMask not available in file. Filter conditions cannot be checked.")
```

Aske-Rosted:
Does this look good? @RasmusOrsoe

src/graphnet/data/dataconverter.py
```diff
@@ -435,6 +437,8 @@ def _extract_data(self, fileset: FileSet) -> List[OrderedDict]:
                     continue
             if self._skip_frame(frame):
                 continue
+            if self._filter_mask(frame, self._I3filters):
```
RasmusOrsoe, line 440: `self._filter_mask` will return `False` by default `if "FilterMask" not in frame`. That means that everyone will be unable to extract i3frames without this field. That's not good!

You can resolve this by either

A) Make `_filter_mask` return `True` by default in these cases
B) change `if self._filter_mask(frame, self._I3filters)` to `if self._filter_mask(frame, self._I3filters) & len(self._I3filters) > 0`.

src/graphnet/data/dataconverter.py
```diff
@@ -566,3 +570,27 @@ def _skip_frame(self, frame: "icetray.I3Frame") -> bool:
         if frame["I3EventHeader"].sub_event_stream == "NullSplit":
             return True
         return False
+
+    def _filter_mask(
+        self, frame: "icetray.I3Frame", I3filters: List[str]
```
RasmusOrsoe, line 575: You can get rid of the codeclimate error if you let _filter_mask access the i3filters inside the function, instead of passing them as argument.

RasmusOrsoe (commented):
Hey! Almost there. Left two comments

src/graphnet/data/dataconverter.py
RasmusOrsoe: @Aske-Rosted

Aske-Rosted: It should very much work such that a `True` return from the _filter_mask call will skip the frame (through a continue call) while a `False` return will move on to include the frame. I agree this is a little counter intuitive and should probably be specified in the description.

RasmusOrsoe: Sorry. I was fooled. You are right!

Aske-Rosted:
The IceTray unit test is failing due to missing due to  `[Errno 28] No space left on device` which I don't see how is related to these changes. @RasmusOrsoe would you mind giving it a look over?

RasmusOrsoe:
@Aske-Rosted I re-ran the workflow and now the disk space issue is gone. However, now a new error about missing files is there. Could you double check that you did not by accident cause this?

Aske-Rosted:
> @Aske-Rosted I re-ran the workflow and now the disk space issue is gone. However, now a new error about missing files is there. Could you double check that you did not by accident cause this?

Fixed the unit test issue.

src/graphnet/data/dataconverter.py
```diff
@@ -167,6 +169,14 @@ def __init__(
         self._input_file_batch_pattern = input_file_batch_pattern
         self._workers = workers
 
+        # I3Filters (NullSplitI3Filter is always included)
+        self._I3Filters = [NullSplitI3Filter()] + I3_Filters
+
+        for filter in self._I3Filters:
```
RasmusOrsoe, line 175: `self._i3filters`

RasmusOrsoe (commented):
Looks great now. Just a comment on snake-case argument naming. Well done @Aske-Rosted !

RasmusOrsoe:
@Aske-Rosted The IceTray disk space error is now fixed in main. Please try to update the branch and see if this doesn't fix the error.

Aske-Rosted:
> @Aske-Rosted The IceTray disk space error is now fixed in main. Please try to update the branch and see if this doesn't fix the error.

hmm new unit errors and I do not really see how they are related to this change... any thoughts @RasmusOrsoe?

RasmusOrsoe (approved):
Looks great!

Issue #588, opened 2023-09-11
Move functionality from PerturbedDataset into GraphDefinition or it's submodules


RasmusOrsoe:
`SQLiteDatasetPerturbed` was created for the DynEdge paper to estimate the induced change in resolutions by perturbing input parameters. At the time, the most natural way to incorporate the perturbation was via the `Dataset` class. This is no longer the case.

1. We should move the perturbation functionality into `GraphDefinition` (or submodules) and allow the user to turn perturbation on/off. E.g: `my_graph_definition = GraphDefinition(perturb = True)` # Draws perturbations from default distributions (normal with pre-defined mean and variance)

2. We should delete` PerturbedDataset`'s .

AMHermansen:
I like the refactoring of `SQLiteDatasetPerturbed`. I think it would be better to make a subclass of `GraphDefinition` or `Detector` called `PerturbedGraphDefinition` or `PerturbedDetector` this would be better in line with our clean-code style guide Design rules 2 ( [see here](https://gist.github.com/wojteklu/73c6914cc446146b8b533c0988cf8d29) ). And it would allow people to more easily make perturbations from a more customizable distribution.

PR #603, opened 2023-09-22
Config files refactor, Examples polish, Perturbation in GraphDefinition

### LINKED ISSUE #583: Make sure examples reflect intended usage
Some of our current examples can give a misleading intended usage of the code base. .e.g #582. We should iterate through the examples and make sure we're not providing misunderstandings.

Here's a few things I've noticed:

1. Make sure that it's clear that config files are for reproducibility only. 
2. GenericExtractor is a special module thats slow and not a great place to start for data conversion.

RasmusOrsoe:
A continuation of #591.

Closes #583 and #588


AMHermansen:
Since I've also contributed to this PR maybe it would be better if @Aske-Rosted had a look at it?

examples/04_training/README.md
```diff
@@ -2,44 +2,44 @@
...
+(graphnet) $ python examples/04_training/01_train_model_dynedge.py --gpus 0 \
     --model-config configs/models/example_direction_reconstruction_model.yml  \
     --prediction-names zenith_pred zenith_kappa_pred azimuth_pred azimuth_kappa_pred
 ```
 
-**`02_train_model_without_configs.py`** Shows how to train a GNN on neutrino telescope data **without configuration files,** i.e., by programatically constructing the dataset and model used. This is good for debugging and experimenting with different dataset settings and model configurations, as it is easier to build the model using the API than by writing configuration files from scratch. For instance, try running:
+**`02_train_model_dynedge_from_config.py** Shows how to train a GNN on neutrino telescope data **using configuration files** to construct the dataset that loads the data and the model that is trained. This is the recommended way to configure standard dataset and models, as it is easier to ready and share than doing so in pure code. This example can be run using a few different models targeting different physics use cases. For instance, you can try running:
```
AMHermansen, line 37: I think you mean 03_...

AMHermansen (approved):
I found a small typo in the README.md, but besides that it looks good.
I would like if someone who didn't work on this also had a look at the changes :) before merging this.

examples/04_training/README.md
RasmusOrsoe: Thanks!

RasmusOrsoe:
@Aske-Rosted Could you give this a look at some point? As @AMHermansen both him and I contributed to this, so a 3rd pair of eyes would be brilliant.

examples/04_training/02_train_tito_model.py
Aske-Rosted, line 1: if  the Not version proof indicates that
```
model.save_state_dict()
model.save_config()
```
is more robust than 
`model.save()`
and that this might fail due to a version conflict then I would consider changing the order of the calls

Aske-Rosted (approved):
I left one minor comment otherwise I, did not see anything that looked suspicious to me.

examples/04_training/02_train_tito_model.py
RasmusOrsoe: Thanks for this. What the comment means is not that executing these lines might lead to an error, but that archiving your model as .pth file under one version of graphnet can give you problems if you try to load it in a different version of graphnet. I'll update the comment to reflect this better.

PR #608, opened 2023-09-29
Fix `OSError: [Errno 28] No space left on device` in IceTray workflow

RasmusOrsoe:
Solves the `OSError: [Errno 28] No space left on device` issue in IceTray build for unit tests by switching the docker image from `combo:stable` to a newer and light-weight image `icetray-prod-v1.8.1-ubuntu20.04-X64`.

Downside is that this image does not come with git installed. So I had to comment out lines in `.github/workflows/build.yml` that solves some permission issues required for the code coverage report to be published. Opened an issue with this in  #609 . 

I think we should pass this fix so PRs are now longer held up, and solve #609 soon.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -36,7 +36,7 @@ jobs:
     name: Unit tests - IceTray
     needs: [ check-codeclimate-credentials ]
     runs-on: ubuntu-latest
-    container: icecube/icetray:stable-slim
+    container: icecube/icetray:combo-stable-slim
     steps:
       - name: Set environment variables
         run: |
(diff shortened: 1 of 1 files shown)


RasmusOrsoe:
@Aske-Rosted or @AMHermansen could take a look at this? The disk space error is holding up quite a few PRs now, so getting this through would be very nice.

Aske-Rosted:
> @Aske-Rosted or @AMHermansen could take a look at this? The disk space error is holding up quite a few PRs now, so getting this through would be very nice.

Seems good to me, I like the extra verbosity. Is there no way for us to simply increase the amount of available space?

Aske-Rosted (approved):

RasmusOrsoe:
> > @Aske-Rosted or @AMHermansen could take a look at this? The disk space error is holding up quite a few PRs now, so getting this through would be very nice.
> 
> Seems good to me, I like the extra verbosity. Is there no way for us to simply increase the amount of available space?

I'm much in favor of such a solution, but as of now that is not possible. We would need to find a research grant to pay for the github costs, or invest time in having the github actions run on a dedicated machine hosted at either NBI or TUM.

PR #610, opened 2023-10-03
Feature/sequence bucketing

ArturoLlorente:
Implemented sequence bucketing as a collate function ( collator_sequence_bucketing) used by DataLoader module. This function sorts all graphs and splits the data into mini-batches, where the forward pass is performed separately.

Splits of the different mini-batches can be customized (by default 0-80%, 80-100%).

This method is effective when working with Transformers, since the computational complexity grows with the square of the sequence length. Improvement noted when then maximum number of pulses is greater.


RasmusOrsoe:
@ArturoLlorente The IceTray disk space error is now fixed in main. Please try to update the branch and see if this doesn't fix the error.

RasmusOrsoe (approved):
Looks great @ArturoLlorente

PR #611, opened 2023-10-03
Redefinition of Tito model

ArturoLlorente:
changes to have a more general GNN definition. All models used in the Tito solution can be replicated with this class. Previous version was only covering one especific case.

Small changes in the training example to fit new input variables.


RasmusOrsoe:
@ArturoLlorente Could you try to merge main into this branch again? Hopefully the workflows should work fine now.

RasmusOrsoe (approved):
Looks good!

Issue #420, opened 2023-02-04
Streamline installation process


MortenHolmRep:
**Is your feature request related to a problem? Please describe.**
Improve installation description for a more streamlined installation process

**Describe the solution you'd like**
Two proposals
1. It would be a good idea to divide the installation process into operating systems from the get-go instead of at the end step.
It replicates many of the sub-sections, but it also allows us to tailor the setup to the operating system and better give proposals for solutions to specific issues; it would look something like this:
-| Installing with Linux
--| Installing with IceTray
--| installing with stand
-| Installing with macOS
-| Installing with WSL

2. If it is at all possible, I would also make the suggestion of a PyTorch-like selection for the setup as seen [here](https://pytorch.org/get-started/locally/).
It removes unnecessary options for the user, and the current setup description, seen from a usability perspective, gives the user a set of instructions that some could follow blindly and copy-paste without looking at, e.g. CPU vs GPU setup etc.
With the PyTorch-like setup, we could have subjects like operating system, developer or user, CPU or GPU, etc.

PCJimmmy:
Have spent the last 12 hours trying to install graphnet on local Ubuntu system and run a forked copy of the baseline kernel from the kaggle competition.  The existing documentation for installation is not doing the trick for me.

asogaard:
Following up on this: With the update to PyG 2.0, the installation of pytorch-geometric/PyG should be [much easier](https://pytorch-geometric.readthedocs.io/en/latest/install/installation.html), which in turn might allow us to make the GraphNeT installation process that much simpler.

MortenHolmRep:
> Following up on this: With the update to PyG 2.0, the installation of pytorch-geometric/PyG should be [much easier](https://pytorch-geometric.readthedocs.io/en/latest/install/installation.html), which in turn might allow us to make the GraphNeT installation process that much simpler.

It looks like the PyG 2.0 quickstart method is identical to the Pytorch version outlined in the second bullet point.
However, I have not tested it in our docs.
The code for the element can be seen in the source from line ~164


``` 
<section id="quick-start">
<h2>Quick Start<a class="headerlink" href="#quick-start" title="Permalink to this heading"></a></h2>
<style>
  .quick-start {
    display: flex;
    flex-direction: row;
    flex-wrap: nowrap;
    margin-bottom: 20px;
  }

  .title-column {
    flex-grow: 0;
  }

  .content-column {
    flex-grow: 1;
  }

  .row {
    display: flex;
[...]

PR #612, opened 2023-10-04
Workflow Fix

### LINKED ISSUE #609: Introduce a step in `build-icetray` workflow that installs git
Code coverage report is not published from the unit tests carried out in the icetray workflow. 

We should figure out an elegant way of installing git in this new icetray docker image, and reintroduce the publication step.

RasmusOrsoe:
This PR addresses a series of issues associated with the github workflows

1. As mentioned in #608, Current icetray build does not have git. This PR introduces a icetray build that has git.
2.  This PR modifies the docker-image production workflow that is run after each merge, which has been broken for a while due to missing disk space. I've added a few commands here that should hopefully free up enough space for the workflow to complete.
3. PyG is not compatible with the latest version of pytorch (2.1) and this has been the course of errors on the macOS and ubuntu builds. This PR fixes this by modifying the `setup.py` to require `torch==2.0.0 `.

closes #609

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -106,4 +106,4 @@ jobs:
           coverage run --source=graphnet -m pytest tests/utilities
           coverage report -m
       - name: Print available disk space after unit tests
-        run: df -h  
+        run: df -h
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -108,22 +108,22 @@ jobs:
       - name: Print available disk space after unit tests
         run: df -h  
 
-  build-macos:
-    name: Unit tests - macOS
-    runs-on: macos-latest
-    steps:
-      - uses: actions/checkout@v3
-      - name: Set up Python 3.8
-        uses: actions/setup-python@v4
-        with:
-          python-version: 3.8
-      - name: Install package
-        uses: ./.github/actions/install
-        with:
-          editable: true
-          hardware: "macos"
-      - name: Run unit tests and generate coverage report
-        run: |
-          set -o pipefail  # To propagate exit code from pytest
-          coverage run --source=graphnet -m pytest tests/  --ignore=tests/data/ --ignore=tests/deployment/ --ignore=tests/examples/
-          coverage report -m
+#  build-macos:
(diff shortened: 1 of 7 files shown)


AMHermansen (approved):
I think we should completely remove the yaml-lines related to the macOS unittests instead of just commenting them out, but besides that it looks good.

RasmusOrsoe:
@AMHermansen @ArturoLlorente @Aske-Rosted After I adressed @AMHermansen's comments about removing the references to the macos build in the workflow file, I noticed that similar software errors began to appear for the ubuntu and icetray build too.

I have now figured out what's causing this: the newest version of pytorch (2.1) is not compatible with PyG, and in our `setup.py` we specified `torch >= 2.0` which seems to default to 2.1 most of the time. By fixing the torch version to 2.0, all builds, including macos works. I have therefore re-introduced macos.


@Aske-Rosted @AMHermansen Given these new changes, could you please give this PR a second look before we merge?

AMHermansen (commented):
Hello @RasmusOrsoe 
Could it be that the apparent "incompatibility" between pytorch and pytorch geometric, is a result of the installation process tries to install the pytorch geometric version which is designed for pytorch 2.0?  [torch geometric installation guide ](https://pytorch-geometric.readthedocs.io/en/latest/install/installation.html) If so I think it would be better to change the installation process somehow.

RasmusOrsoe:
@AMHermansen We're actually staying quite true to the intended way of installing these dependencies. The problem here is that we specify `torch>= x`, `torch-geometric >= y` and the support packages for pyg that we depend on too is grabbed through a torch-version-specific wheel link in `requirements.txt`. Because we don't keep an upper bound on the versions, these dependencies can drift out of sync. In this case it was the wheel-link that became incompatible with the torch version. 
PyG  has been working on moving these support packages into either PyG or torch, so in the future things will be simpler. But for now I don't see a more elegant solution. 

In either case, everything now works at `torch 2.1`. In the future, when we introduce an install matrix as suggested in #420 , the user experience of installation will be on-par with what other major libraries offer. But these technical details for work flows might not change much.

PR #613, opened 2023-10-09
Update gpu requirements.txt

RasmusOrsoe:
#612 introduced a recommended torch version for gpu installation that doesn't exist. This PR changes the torch version to one that does exist.


Aske-Rosted (approved):
looks fine.

PR #615, opened 2023-10-10
Update coverage reporting

RasmusOrsoe:
Our coverage reporting has not been accurate for a while, making our coverage appear lower than it actually is. This is fixed in this PR.

The coverage score is based off the coverage report from the icetray workflow. Following this PR, the icetray workflow will run all unit tests (making it a little slow) and provide an accurate coverage report based on these.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -65,7 +65,8 @@ jobs:
           editable: true
       - name: Run unit tests and generate coverage report
         run: |
-          coverage run --source=graphnet -m pytest tests/
+          coverage run --source=graphnet -m pytest tests/ --ignore=tests/examples/04_training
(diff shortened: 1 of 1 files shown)


Aske-Rosted (approved):
While I do not quite understand the need for separating out the `04_training` from the other tests, this otherwise seems reasonable.

RasmusOrsoe:
@Aske-Rosted Very legitimate point. The story is that occasionally these workflows might exit with error code 1, which means they were terminated by github. There's no details to why but online people mention excessive RAM or CPU usage, or that they run for too long. When I was experimenting with the workflows I found that splitting the coverage command into two made it more consistent. I could just be lucky - I don't know. It's been quite difficult to debug this because it's an occasional thing...

PR #617, opened 2023-10-15
'ckpt_path' ambiguous in Model class

ArturoLlorente:
`src/graphnet/models/model.py` class takes `ckpt_path` when calling `fit` method. This variable is used for both:

 - `Trainer(dafault_root_dir)` is defined as: _Default path for logs and weights when no logger/ckpt_callback passed. Default: ``os.getcwd()``._
 - `Trainer.fit(ckpt_path)` is defined as: _Path/URL of the checkpoint from which training is resumed. Could also be one of two special keywords ``"last"`` and ``"hpc"``. If there is no checkpoint file at the path, an exception is raised._

Previous deifinition of `Model` was taking same input fot both variables, which was raising an error.
New implementation takes two different inputs:

-  `ckpt_save_path` as the path used to save the ckeckpoints of the training (`Trainer(dafault_root_dir)`).
-  `ckpt_path` as the checkpoint from which training is resumed (`Trainer.fit(ckpt_path)`).


RasmusOrsoe (approved):
@ArturoLlorente Can you confirm that this change removed your original pathing issues? Also, does pathing issues arise when you set a checkpoint path with and without the wandb callback in this current version of the code?

For transparency; 
@ArturoLlorente and I talked about this PR in person.  @ArturoLlorente found a bug where `default_root_dir=ckpt_path,` lead pathing issues when specifying checkpoint paths.

ArturoLlorente:
Resuming of the training using `ckpt_path` was tested and working properly.
No conflict with `ModelCheckpoint()` callback or `Trainer(default_root_dir=)` were raised

PR #623, opened 2023-10-25
Geometry Tables

RasmusOrsoe:
This PR introduces geometry tables as an inherent property of `Detector` classes, and includes new functionality in `GraphDefinition` that allows users to add inactive sensors to their graphs. This PR supersedes #421.

_Why is this functionality relevant?_
By having the option to add inactive sensors, i.e. the "negative information", we get the ability to compile full data representations of events (Both the active "positive" information and the inactive "negative" information) using `GraphDefinition`. For certain tasks and certain machine learning paradigms, this is a necessity, which we will now support! 

**Details**:
- `Detector` gets four new properties: `sensor_index_name`, `sensor_position_names`, `string_index_name` and `geometry_table`. These properties are set using class variables for each detector.
- Instead of the generic `Prometheus` detector class, I've introduced `ORCA150` which is the correct detector name, and for convenience the ´Prometheus´ class just inherits this. Standardization functions have seen slight modifications.
- `GraphDefinition` gets one new attribute `output_feature_names` defined by the output feature names of the `NodeDefinition`.
-  `GraphDefinition` gets four new arguments:        
 ```
        add_inactive_sensors: bool = False,
        sensor_mask: Optional[List[int]] = None,
        string_mask: Optional[List[int]] = None,
        sort_by: str = None, 
```

> If `add_inactive_sensors=True`, sensors not in the raw input is added with padded pulse attributes.
> if ` sensor_mask` is given, sensors with id's in this list is dropped from the input.
> if ` string_mask` is given, sensors on those strings are dropped form the input.
> `sort_by` specifies a column in `GraphDefinition.output_feature_names` to sort `graph.x` from.

The masking mentioned above is done **after** adding inactive sensors. This new functionality is the main contribution in this PR.
- Renamed arguments in `Detector` from `node_features` to `input_features` see #616 
- Updated `prometheus-events.db` and `prometheus-events.parquet`. They now hold 50 events with pulses between 2 and 100. 
- Added a static sqlite database for upgrade (made from our test i3 file). This is used in the new unit test.
- Expanded the unit tests of `GraphDefinition` to test the new functionality.

Below is a plot that shows execution time of adding the inactive sensors to the input features as a function of event length.
image
As seen in the plot, the computational cost associated with this operation is minimal and is far superior to the method shown in #421.

### CODE REVIEW

File: src/graphnet/models/detector/prometheus.py [hunk 1]
```diff
@@ -26,4 +35,8 @@ def _sensor_pos_z(self, x: torch.tensor) -> torch.tensor:
         return (x + 350) / 100
 
     def _t(self, x: torch.tensor) -> torch.tensor:
-        return ((x / 1.05e04) - 1.0) * 20.0
```
AMHermansen, line 29: Any particular reason why the time normalization function was changed?

src/graphnet/models/graphs/graph_definition.py
```diff
@@ -55,6 +59,12 @@ def __init__(
                                to None.
             seed: seed or Generator used to randomly sample perturbations.
                   Defaults to None.
+            add_inactive_sensors: If True, inactive sensors will be appended
+                to the graph with padded pulse information. Defaults to False.
+            sensor_mask: A list of sensor id's to be masked from the graph. Any
+                sensor listed here will be removed from the graph. Defaults to None.
+            string_mask: A list of string id's to be masked from the graph. Defaults to None.
+            sort_by: Name of node feature to sort by. Defaults to None.
```
AMHermansen, line 67: I'm not quite sure I understand what it means to have None as default for sort_by.

AMHermansen (approved):
Looks good.

src/graphnet/models/detector/prometheus.py
RasmusOrsoe: The previous scaling is sub-optimal for the full dataset. I'll share the full dataset as a CuratedDataset soon.

src/graphnet/models/graphs/graph_definition.py
RasmusOrsoe: It means that there is no sorting done.

PR #624, opened 2023-11-02
Refactored model and standardmodel

### LINKED ISSUE #599: Should we refactor Model and StandardModel?
Currently we have a lot of methods defined in the `Model` class, however since `Model` is a superclass of almost every commonly used class in GraphNeT, we get weird functionality, where a user is free to call methods like `predict, predict_as_dataframe, fit`etc. from classes like `Detector, GraphDefinition`. I'm not quite sure what the behavior would be, either you would get some nonsensical behavior or a runtime-error, all of these are marked as "public" methods and should be callable from any subclass of Model. Most of these methods are to my understanding only supposed to be called from a `StandardModel` instance. Hence I propose that we figure out what methods are essential for all our classes and which should be moved to the `StandardModel`.

From a quick glance of the `Model` class I believe the only truly universal methods are:
`save, load, save_state_dict, load_state_dict, from_config`.
I think moving the remaining functions to StandardModel, would clean up the code, and remove weird behavior where a Detector class tries to call a fit function.

**Additional context**
Perhaps there are other methods, which are meaning full for all classes that I missed.

AMHermansen:
Closes #599


src/graphnet/models/standard_model.py
```diff
@@ -94,22 +180,14 @@ def configure_optimizers(self) -> Dict[str, Any]:
             )
         return config
 
-    def forward(
-        self, data: Union[Data, List[Data]]
-    ) -> List[Union[Tensor, Data]]:
+    def forward(self, data: Data) -> List[Union[Tensor, Data]]:
```
RasmusOrsoe, line 183: Looks like you've removed the functionality introduced in #610. Was this intentional? If so, could you elaborate on why?

RasmusOrsoe (commented):
Hey @AMHermansen!

Thank you for this! I think the refactor looks good, but it also appears that you have removed the functionality introduced in #610 . Was this intentional?

src/graphnet/models/standard_model.py
AMHermansen: @RasmusOrsoe Good catch, it would seem I was doing the refactor from an outdated branch prior to #610 . I've now added the functionality from #610 to the refactor.

RasmusOrsoe (approved):
@AMHermansen Thanks!

PR #631, opened 2023-11-17
Update awtd and fadc parsing method

RasmusOrsoe:
Currently we extract the pulse variables `fadc` and `awtd` in `I3FeatureExtractorIceCube86` like so:

```
output["awtd"].append((pulse.flags >> 1) & 0x1)  # bit 1
output["fadc"].append((pulse.flags >> 2) & 0x1)  # bit 2
```
However, recent studies by Tom Stuttard has shown this is inconsistent. This PR updates the parsing method of `awtd` to the new suggested method  in the low energy oscillation working group. This method relies on the pulse width to identify pulses read out using the `awtd` digitizer. This means that the `fadc `bool is now defined as `~awtd`. For this reason, the separate entry for `fadc` has been removed in this PR.

### DIFF
--- a/src/graphnet/data/extractors/i3featureextractor.py
+++ b/src/graphnet/data/extractors/i3featureextractor.py
@@ -202,7 +202,9 @@ class I3FeatureExtractorIceCube86(I3FeatureExtractor):
         https://github.com/icecube/icetray/issues/2721 Note that the issue
         states to use 8ns, but I have found that actually 6ns is correct.
         """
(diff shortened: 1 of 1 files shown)


Aske-Rosted:
Seems reasonable.

AMHermansen (commented):
I don't really understand the IceTray related code, and thus don't feel confident approving this.

Aske-Rosted (approved):
Sorry forgot to approve...

PR #633, opened 2023-11-23
cleaning code

AMHermansen:
Default parameters are evaluated at definition, meaning that a `NodesAsPulses` object will be created, the moment this file is imported. This starts a logging procedure, and is generally undesireable.


RasmusOrsoe (approved):
Thanks for tracking this down. It's been annoying my logs for a while!

PR #636, opened 2023-11-29
enforce gpu arg in .predict_as_dataframe

RasmusOrsoe:
Currently, the `gpus` argument from the CLI in our training examples doesn't get passed to `model.predict_as_dataframe`, essentially forcing the prediction to happen on CPU.

This has caused some confusion for users that have tried to extend the training examples into training scripts for their own use cases. 

This PR passes the argument to the `model.predict_as_dataframe`.


AMHermansen (approved):
Looks good

PR #639, opened 2023-12-02
Task refactor

RasmusOrsoe:
<work in progress >

### DIFF
--- a/src/graphnet/models/task/__init__.py
+++ b/src/graphnet/models/task/__init__.py
@@ -1,3 +1,8 @@
 """Physics task-specific modules to be used as model "read-outs"."""
 
-from .task import Task, IdentityTask
+from .task import (
+    Task,
+    IdentityTask,
+    StandardLearnedTask,
+    StandardFlowTask,
+)
--- a/src/graphnet/models/task/classification.py
+++ b/src/graphnet/models/task/classification.py
@@ -5,7 +5,7 @@ from typing import Any
 import torch
 from torch import Tensor
 
-from graphnet.models.task import Task, IdentityTask
+from graphnet.models.task import IdentityTask, StandardLearnedTask
 
 
 class MulticlassClassificationTask(IdentityTask):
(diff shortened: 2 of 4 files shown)


AMHermansen (approved):
This looks good.

I'm not sure if we will be able to guarantee that `StandardModel` will be compatible with all tasks in the future. In particular, I could imagine that the `predict/predict_as_dataframe`, might not work with all imagineable tasks in the future. Perhaps we should restrict the tasks provided to `StandardModel` to be `StandardLearnedTask`?

Now that we are going about refactoring tasks I think that the aggregation parts of the GNN's might be better placed in the task section. The reason would be that the GNN's would now only be responsible for feature learning, and it would be easier to do transfer-learning between various tasks, where you might have one tasks which predicts on an entire pulsemap and another task, which does regression/classification.

In particular I think we could introduce an `Aggregation` module, which will take in a variable length pulsemap and return a "fixed length" tensor. This Module would then act as the "link" between the feature-pulsemap in a GNN and the MLP in the StandardLearnedTask, but would be omitted when doing per-node reconstructions.

RasmusOrsoe:
> This looks good.
> 
> I'm not sure if we will be able to guarantee that `StandardModel` will be compatible with all tasks in the future. In particular, I could imagine that the `predict/predict_as_dataframe`, might not work with all imagineable tasks in the future. Perhaps we should restrict the tasks provided to `StandardModel` to be `StandardLearnedTask`?
> 

You are absolutely right. The next few PRs in the normalizing flow series introduce a few extra lines of code that makes .fit and .predict compatible with this new `FlowTask`. I have also just now introduced a check in this PR to make sure all tasks are `StandardLearnedTask`.

> Now that we are going about refactoring tasks I think that the aggregation parts of the GNN's might be better placed in the task section. The reason would be that the GNN's would now only be responsible for feature learning, and it would be easier to do transfer-learning between various tasks, where you might have one tasks which predicts on an entire pulsemap and another task, which does regression/classification.
> 
> In particular I think we could introduce an `Aggregation` module, which will take in a variable length pulsemap and return a "fixed length" tensor. This Module would then act as the "link" between the feature-pulsemap in a GNN and the MLP in the StandardLearnedTask, but would be omitted when doing per-node reconstructions.

Perhaps - but I do think it makes a lot of sense to allow models to have aggregation in them. To me it sounds like you're describing ensemble functionality, but perhaps we can discuss this on Tuesday's meeting.

PR #640, opened 2023-12-02
`StandardModel.fit`: Automatically load in best-fit weights from early stopping when training is finished.

RasmusOrsoe:
Implements my proposed changes to default callbacks in `StandardModel.fit`. See #632


AMHermansen (approved):
Looks good

Issue #579, opened 2023-09-05
Adopt pytorch lightning DataModule instead of DataLoaders and Dataset


AMHermansen:
It would be beneficial to go away from using native pytorch `Dataset/Dataloader` and use pytorch lightning `DataModule` instead([see here](https://lightning.ai/docs/pytorch/stable/data/datamodule.html)) , since this would allow GraphNeT to better interact with the rest of the lightning framework. In particular we would get access to the lightning profiling tool ([see here](https://lightning.ai/docs/pytorch/stable/tuning/profiler_basic.html)), which could shed some light on where it might make sense to make optimizations.

**Describe the solution you'd like**
Refactor Dataset and DataLoader related classes to be DataModules instead.

**Additional context**
Doing this would seem to solve #275

What are your thoughts @RasmusOrsoe

RasmusOrsoe:
I'm OK with such a refactor if

1. We get new functionality from it. E.g. profiling (But is it worth the trouble?)
2. We do not loose the ability to loop over batches in dataloaders in native pytorch i.e.
```
 for batch in DataLoader:
       pred = model(batch)
```

AMHermansen:
It should be straight forward to get whichever Dataloader you wish (train / validation / test / prediction) using the DataModule's methods `train_dataloader` etc.
You would simply do:

`train_dataloader = GraphNeTDataModule.train_dataloader()`

samadpls:
Hey @AMHermansen can i work on this issue?

AMHermansen:
Hello @samadpls

You are more than welcome to work on this issue. Right now however it doesn't seem clear to me what actually needs to be changed before this issue is resolved, therefore I think a good first step is to make an outline of exactly what needs to be changed. Some initial ideas would be:

1. Implement a DataModule class. Does this class require it's own config, or is the config managed by the DataModule's Datasets?
2. We need to figure out how the DataModule class should interact with the StandardModel class. In particular, currently we're only implementing `train_step/validation_step` out of the four steps `train/validation/test/predict`, should we also implement `test_step/predict_step`? Most of our logic involving prediction seems to be targeted at testing.

Maybe for a start we should be satisfied only implementing the first step, and then we can consider how much of StandardModel (if anything at all) needs to be refactored, before we have harmonious interaction between the DataModule and the StandardModel.

@RasmusOrsoe @samadpls What are your thoughts?

RasmusOrsoe:
[...]

PR #642, opened 2023-12-02
Adopted pytorch lightning DataModule instead of DataLoaders and Dataset

samadpls:
Adopted PyTorch Lightning DataModule instead of managing individual DataLoaders and Dataset.
This pull request addresses and resolves issue #579


src/graphnet/data/datamodule.py
```diff
@@ -0,0 +1,338 @@
+from typing import Dict, Any, Optional, List, Tuple, Union
+import lightning as L
```
AMHermansen, line 3: Very cool that you're trying to get this to work. We are currently using the "older" syntax with pytorch_lightning and not lightning. I suspect that the run-time error that you're getting might a result of this.

Maybe you could try to do:
`import pytorch_lightning as pl`
and then also change where you've used `L.` Hope this works :)

src/graphnet/data/datamodule.py
```diff
@@ -0,0 +1,429 @@
...
+        train_val_split: Optional[List[float]] = [0.9, 0.10],
+        split_seed: int = 42,
+    ) -> None:
+        """Create dataloaders from dataset.
+
+        Args:
+            dataset_reference: A non-instantiated reference to the dataset class.
+            dataset_args: Arguments to instantiate graphnet.data.dataset.Dataset with.
+            selection: (Optional) a list of event id's used for training and validation, Default None.
+            test_selection: (Optional) a list of event id's used for testing, Default None.
```
RasmusOrsoe, line 39: A few of these lines are longer than 79 characters. Could you break them up onto multiple lines, please? VSCode has the ability to displace a vertical ruler in the editor  that indicates character length. See [here](https://dev.to/brad_beggs/vs-code-vertical-rulers-for-prettier-code-3gp3). 

E.g: 

```python
"""
dataset_reference: A non-instantiated reference to the dataset class.
dataset_args: Arguments to instantiate 
            graphnet.data.dataset.Dataset with.
selection: (Optional) a list of event id's used for training
           and validation, Default None.
"""
```

src/graphnet/data/datamodule.py
```diff
@@ -0,0 +1,429 @@
...
+
+        Checks whether the dataset is an instance of SQLiteDataset,
+        ParquetDataset, or Dataset. Raises a TypeError if an invalid dataset
+        type is detected, or if an EnsembleDataset is used.
+        """
+        print(self._dataset, "Dataset\n")
+        print(
+            f"Type of self._dataset before validation check: {type(self._dataset)}"
+        )
+        # if type(self._dataset) not in [SQLiteDataset, ParquetDataset, Dataset]:
```
RasmusOrsoe, line 193: Looks like a few lines here are commented out

src/graphnet/data/datamodule.py
```diff
@@ -0,0 +1,429 @@
...
+        if isinstance(self._dataset_args["path"], list):
+            if self._selection is not None:
+                try:
+                    # Check that the number of dataset paths is equal to the
+                    # number of selections given as arg.
+                    assert len(self._dataset_args["path"]) == len(
+                        self._selection
+                    )
+                except AssertionError:
+                    raise ValueError(
```
RasmusOrsoe, line 213: You can shorten the `ValueError` lines by breaking it into multiple smaller strings. Like so:

```python
raise ValueError("The number of dataset paths" 
                 f" ({len(self._dataset_args['path'])})" 
                 " does not match the number of" 
                 f" selections ({len(self._selection)}).")
```

src/graphnet/data/datamodule.py
```diff
@@ -0,0 +1,429 @@
...
+
+            if self._test_selection is not None:
+                try:
+                    # Check that the number of dataset paths is equal to the
+                    # number of test selections.
+                    assert len(self._dataset_args["path"]) == len(
+                        self._test_selection
+                    )
+                except AssertionError:
+                    raise ValueError(
```
RasmusOrsoe, line 225: See previous comment on shortening this line

src/graphnet/data/datamodule.py
```diff
@@ -0,0 +1,429 @@
...
+            )
+        if "dataset" in self._test_dataloader_kwargs:
+            raise ValueError(
+                "`test_dataloader_kwargs` must not contain `dataset`"
+            )
+
+    def _resolve_selections(self) -> None:
+        if self._test_selection is None:
+            self.warning_once(
+                f"{self.__class__.__name__} did not receive an argument for `test_selection` and will therefore not have a prediction dataloader available."
```
RasmusOrsoe, line 247: This can be shortened like the `ValueError`s. Something like: 
```python
self.warning_once(f"{self.__class__.__name__} did not receive an" 
                  " argument for `test_selection` and will "
                  "therefore not have a prediction dataloader available.")
```

src/graphnet/data/datamodule.py
```diff
@@ -0,0 +1,429 @@
...
+                (
+                    self._train_selection,
+                    self._val_selection,
+                ) = self._split_selection(  # type: ignore
+                    self._selection
+                )
+
+        else:  # selection is None
+            # If not provided, we infer it by grabbing all event ids in the dataset.
+            self.info(
```
RasmusOrsoe, line 274: Same as `self.warning_once`

samadpls:
@RasmusOrsoe, thank you for reviewing and giving your feedback. I have implemented the necessary adjustments to address the comment. 😋

RasmusOrsoe (approved):
🥇

PR #645, opened 2023-12-09
Fix save_state_dict

AMHermansen:
Current implementation moves the entire model to cpu, whenever save_state_dict is called. This seems like an undesireable side effect of the method. This PR changes save_state_dict to only save the statedict, but keeps the model on the current device.

### DIFF
--- a/src/graphnet/models/model.py
+++ b/src/graphnet/models/model.py
@@ -49,7 +49,10 @@ class Model(
             self.info(
                 "It is recommended to use the .pth suffix for state_dict files."
             )
-        torch.save(self.cpu().state_dict(), path)
+        state_dict = self.state_dict()
+        for key, value in state_dict.items():
+            state_dict[key] = value.cpu()
+        torch.save(state_dict, path)
         self.info(f"Model state_dict saved to {path}")
 
     def load_state_dict(


RasmusOrsoe:
@AMHermansen thanks for this suggestion. I'm not sure we want this change though - If the state dict is saved to disk from gpu, it will require the model to be on gpu when the state dict is loaded in again, or an error will be thrown. see https://pytorch.org/tutorials/recipes/recipes/save_load_across_devices.html . 

Has this been a issue for you?

AMHermansen:
> @AMHermansen thanks for this suggestion. I'm not sure we want this change though - If the state dict is saved to disk from gpu, it will require the model to be on gpu when the state dict is loaded in again, or an error will be thrown. see https://pytorch.org/tutorials/recipes/recipes/save_load_across_devices.html .
> 
> Has this been a issue for you?

The suggested change in this PR only removes the side effect from the current `save_state_dict` implementation to not move the model to cpu when this is called. This is done by copying the state_dict to cpu and then saving the copy. The reason for this implementation is to make saving models more streamlined, my current understanding from the example scripts is that `save_model_config` and `save_state_dict` is the intended way to save graphnet models. If you however want to save a model like this during training, you will run into problems, since the model will be moved away from the accelerator.

RasmusOrsoe (approved):
@AMHermansen sorry, I glanced over this too quickly. I was under the impression that the state dict was saved on whatever device it happened to be on; upon looking at the code again I see that's not the case.

PR #646, opened 2023-12-12
Change argument `gnn` to `architecture` 

RasmusOrsoe:
This is the second part in the series of PRs introducing normalizing flows to graphnet. 

As we begin to support model architectures that are not inherently GNNs, our current synonymous use of `gnn` in place of `architecture` can become misleading. This PR aims to make the distinction clearer and intuitive.

**What this PR does:**
This PR updates references to `gnn` with `architecture` in places where we really do mean `architecture`.  

model architectures are now stored under `self._architecture` in `StandardModel`. This is a **breaking** change for state dicts saved in a previous version of graphnet using `StandardModel.save_state_dict()`.

**To mitigate these breaking changes, the PR does the following:** 
`DeprecationWarning` is given when passing `gnn` argument to `StandardModel`. This means `ModelConfig` files compiled before this PR is still compatible. I have started a thread that compiles deprecated items for the next release in #647.

state dicts that can be "updated" to this new naming scheme by using the following snippet:

```python
import torch
import dill
from copy import deepcopy

def convert_state_dict(path: str,
                       save_path: str) -> None:
    """Update old state dicts.

    Converts fields named with `_gnn` in the state dict given by `path`
    into fields named `_architecture` and saves the new state dict as `save_path`.

    Args:
        path: Path to old state dict you want to convert.
        save_path: Path to save the new state dict to. Must contain full
                   file name.
    """
    
    assert path.endswith('.pth'), "Please point to a specific .pth file."
    assert save_path.endswith('.pth'), "Please include filename and extension."

    # Load old state dict
    original_state_dict = torch.load(path,
                                     map_location='cpu',
                                     pickle_module = dill)
    
    # Make a carbon-copy
    new_state_dict = deepcopy(original_state_dict)

    # Replace old entries in copy 
    for old_key in original_state_dict.keys():
        if '_gnn' in old_key:
            new_key = old_key.replace('_gnn', '_architecture')
            new_state_dict[new_key] = new_state_dict.pop(old_key)
    
    # Save updated copy
    torch.save(obj = new_state_dict,
               f = save_path,
               pickle_module=dill)
```

### DIFF
--- a/src/graphnet/models/standard_model.py
+++ b/src/graphnet/models/standard_model.py
@@ -33,7 +33,6 @@ class StandardModel(Model):
         *,
         graph_definition: GraphDefinition,
         architecture: GNN = None,
-        gnn: Optional[GNN] = None,
         tasks: Union[StandardLearnedTask, List[StandardLearnedTask]],
         optimizer_class: Type[torch.optim.Optimizer] = Adam,
         optimizer_kwargs: Optional[Dict] = None,
@@ -51,20 +50,8 @@ class StandardModel(Model):
         assert isinstance(tasks, (list, tuple))
         assert all(isinstance(task, StandardLearnedTask) for task in tasks)
         assert isinstance(graph_definition, GraphDefinition)
-
-        # deprecation warnings
-        if (architecture is None) & (gnn is not None):
-            architecture = gnn
-            # Code continues after warning
-            self.warning(
-                """DeprecationWarning: Argument `gnn` will be deprecated in GraphNeT 2.0. Please use `architecture` instead."""
-            )
-        elif (architecture is None) & (gnn is None):
-            # Code stops
-            raise TypeError(
-                "__init__() missing 1 required keyword-only argument: 'architecture'"
-            )
         assert isinstance(architecture, GNN)
+
         # Member variable(s)
         self._graph_definition = graph_definition
         self._architecture = architecture
--- a/src/graphnet/models/standard_model.py
+++ b/src/graphnet/models/standard_model.py
@@ -32,7 +32,8 @@ class StandardModel(Model):
         self,
         *,
         graph_definition: GraphDefinition,
-        gnn: GNN,
+        architecture: GNN = None,
+        gnn: Optional[GNN] = None,
         tasks: Union[StandardLearnedTask, List[StandardLearnedTask]],
         optimizer_class: Type[torch.optim.Optimizer] = Adam,
         optimizer_kwargs: Optional[Dict] = None,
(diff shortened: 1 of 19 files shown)


src/graphnet/models/standard_model.py
```diff
@@ -50,11 +51,24 @@ def __init__(
         assert isinstance(tasks, (list, tuple))
         assert all(isinstance(task, StandardLearnedTask) for task in tasks)
         assert isinstance(graph_definition, GraphDefinition)
-        assert isinstance(gnn, GNN)
+
+        # deprecation warnings
+        if (architecture is None) & (gnn is not None):
+            architecture = gnn
+            # Code continues after warning
+            self.warning(
```
AMHermansen, line 59: Any reason why you're using the logger and not ```warnings.warn(""Argument `gnn` will be deprecated in GraphNeT 2.0. Please use `architecture` instead., DeprecationWarning)```. To me this feels like an avoidable warning made by a user. See [python documentation on logging](https://docs.python.org/3/howto/logging.html)

AMHermansen (commented):
Hello @RasmusOrsoe,

I think this is a well motivated change to GraphNeT.

There exists a fine-tuning callback in lightning which freezes all parameters of a module named `backbone`. Have you considered renaming `_architecture` to `backbone` to allow support for this callback? See [BackboneFinetuning](https://lightning.ai/docs/pytorch/stable/api/lightning.pytorch.callbacks.BackboneFinetuning.html)

I'm thinking for backward compatibility would it be possible to automatically change the `load_state_dict` method to change "old" state_dicts to the new format, or possibly add a new method to model called `load_old_state_dict` which can load the old state_dict?

I have some minor comments, let me know what you think.

RasmusOrsoe:
@AMHermansen thanks for pointing my attention to this callback from lightning. I think it does exactly what we will soon need, so even though I dislike the term `backbone`, I've renamed `_architecture` to `backbone`. 

I've added a module `deprecation_tools.py` under `graphnet.utilities` and put in an altered version of the snippet for state dict entry renaming. This function is now called in `Model.load_state_dict` such that older state dicts will have continued support. When it renames an entry, a deprecation warning is given.

On the matter of `self.warning` vs. `warnings.warn`: For reasons I did not care to investigate deeply, pre-commit hooks (at least mine) fail to parse the files if I use warnings.warn. Figuring out exactly what causes this is beyond the scope of this PR in my opinion.

AMHermansen (approved):
I think it looks good.

I've left a comment you can think about :-)

PR #651, opened 2024-01-21
Log current lr in `StandardModel` 

RasmusOrsoe:
This PR changes the following in `graphnet.models.StandardModel`:

- Adds the global optimizer learning rate as a logged quantity, which makes it appear both in `wandb` and in the `ProgressBar`.
- Sets `on_step=True` for the logging of training loss. 

This change makes it easier to monitor the progress of the training from the progressbar information alone.


ArturoLlorente (approved):
Looks nice!

I also explored the possibility to use `self.trainer.lr_scheduler_configs[0].scheduler.get_last_lr()` which gives automaticly the LR of all optimizers, but I think this solution is better since it will work even if there is no scheduler defined.

PR #652, opened 2024-01-22
Update `QUESO` energy model and unit test

RasmusOrsoe:
The current state_dict in `graphnet/models/pretrained/icecube/upgrade/QUESO/total_neutrino_energy` is not identical to the one used in IceCube Upgrades processing pipeline.

This PR does the following:

1. Updates the state dict in `graphnet/models/pretrained/icecube/upgrade/QUESO/total_neutrino_energy` to the correct one
2. Updates the QUESO integrity unit tests (A bug here was found when performing 1.)
3. Makes `GraphNeTI3Module` inherit from `graphnet.utilities.logging.Logger` and adds a `warning` when the inference module is run on events with no pulses in the pulsemap (and therefore pads prediction entries with `NaN`)


Aske-Rosted (approved):
Looks good to me.

PR #653, opened 2024-01-23
Rnn attention

Aske-Rosted:
This PR adds RNN_TITO model, and the necessary sub modules. This model has been developed specifically to tackle the issue of having large amounts of DOM activation's or pulses within the high energy domain. Many of the models suggested in the _Kaggle_ competition simply limited the max amount of pulses so that the models could fit inside memory. 

The RNN TITO model is based on applying a _reccurent neural network_ per node treating the pulse data as time series data, effectively reducing the number of nodes from _# pulses_ to _# activated DOMs_. The output of the RNN is joined with the  DOM summary features `x,y,z first-hit-time, total-energy`. The output of this module is then used as input to a model based on the first place _Kaggle_ competition winner _TITO_. 

This PR also allows for the learnable _fourier transform_ (or _sinusodial positional embedding_ as the module is named inside the code) as it was implemented in the 2nd place _Kaggle_ competition solution. 

The PR also allows for more customization of the _dynedge tito_ model. 

A schematic of the model can be seen below.

[DOM RNN model.pdf](https://github.com/graphnet-team/graphnet/files/14019168/DOM.RNN.model.pdf)


src/graphnet/models/components/embedding.py
```diff
@@ -0,0 +1,27 @@
+"""Classes for performing embedding of input data."""
+import torch
+
+
+class SinusoidalPosEmb(torch.nn.Module):
+    """Sinusoidal positional embedding layer."""
+
+    def __init__(self, dim: int = 16, M: int = 10000) -> None:
```
RasmusOrsoe, line 8: M -> m for consistency.

src/graphnet/models/components/embedding.py
```diff
@@ -0,0 +1,27 @@
...
+        """Construct `SinusoidalPosEmb`.
+
+        Args:
+            dim: Embedding dimension.
+            M: Number of frequencies.
+        """
+        super().__init__()
+        self.dim = dim
+        self.M = M
```
RasmusOrsoe, line 17: self.m = m

src/graphnet/models/components/embedding.py
```diff
@@ -0,0 +1,27 @@
...
+import torch
+
+
+class SinusoidalPosEmb(torch.nn.Module):
+    """Sinusoidal positional embedding layer."""
+
+    def __init__(self, dim: int = 16, M: int = 10000) -> None:
+        """Construct `SinusoidalPosEmb`.
+
+        Args:
```
RasmusOrsoe, line 11: I think a nice fat doc string here would be good. It's a bit hard to read this code and understand what it does and why this is motivated. Perhaps you could spell it out here and point to the kaggle paper for details?

src/graphnet/models/rnn/node_rnn.py
```diff
@@ -0,0 +1,85 @@
...
+    @save_model_config
+    def __init__(
+        self,
+        nb_inputs: int,
+        hidden_size: int,
+        num_layers: int,
+        RNN_dropout: float = 0.5,
+        embedding_dim: int = 0,
+    ) -> None:
+        """Construct `NodeTimeRNN`.
```
RasmusOrsoe, line 35: Class is called `Node_RNN`

src/graphnet/models/rnn/node_rnn.py
```diff
@@ -0,0 +1,85 @@
...
+    NodeAsDOMTimeSeries), which is not standard in the graphnet framework.
+    """
+
+    @save_model_config
+    def __init__(
+        self,
+        nb_inputs: int,
+        hidden_size: int,
+        num_layers: int,
+        RNN_dropout: float = 0.5,
```
RasmusOrsoe, line 32: i'd recommend just calling this "dropout" as you only have one dropout method here, as far as I can see. You can clarify in the doc string that this dropout is used for the RNN

src/graphnet/models/rnn/node_rnn.py
```diff
@@ -0,0 +1,85 @@
...
+
+        if self._embedding_dim != 0:
+            self._nb_inputs = self._embedding_dim * 2 * nb_inputs
+
+        self._rnn = torch.nn.GRU(
+            num_layers=self._num_layers,
+            input_size=self._nb_inputs,
+            hidden_size=self._hidden_size,
+            batch_first=True,
+            dropout=RNN_dropout,
```
RasmusOrsoe, line 60: `dropout` see comment above

src/graphnet/models/rnn/node_rnn.py
```diff
@@ -0,0 +1,85 @@
...
+        )
+        self._emb = SinusoidalPosEmb(dim=self._embedding_dim)
+
+    def forward(self, data: Data) -> torch.Tensor:
```
RasmusOrsoe, line 64: As it is written, this class is very specifically tailored to your use-case and choice in `graph_definition`. I think we could greatly increase the re-usability of this method by changing the input assumption.  Instead of assuming `data` with a named field called `time_series`, we could instead let this method receive a generic `x : Tensor` and `cutter: Tensor` (although lets find a better name. Happy to help here, but I am not entirely sure what cutter represents), and instead of returning a `Data` object, it would return a `rnn_out: Tensor` (then your type hints is also correct ;-) ) .  We could then move this method into `components` and have users apply this technique inside their `backbone`'s.

src/graphnet/models/graphs/nodes/nodes.py
```diff
@@ -211,3 +214,110 @@ def _construct_nodes(self, x: torch.Tensor) -> Data:
...
+        if self._max_activations is not None:
+            counts[counts > self._max_activations] = self._max_activations
+            time_series = [
+                image[: self._max_activations] for image in time_series
+            ]
+        time_series = np.concatenate(time_series)
+        # apply inverse hyperbolic sine to charge values (handles zeros unlike log scaling)
+        time_series[:, 0] = np.arcsinh(5 * time_series[:, 0]) / 5
+
+        return Data(
```
RasmusOrsoe, line 318: Is there any way we could avoid having this `NodeDefinition` write custom fields to the `Data` object? Models downstream that looks for these would only be compatible with this NodeDefinition, which reduces reusability.

src/graphnet/models/graphs/nodes/nodes.py
```diff
@@ -211,3 +214,110 @@ def _construct_nodes(self, x: torch.Tensor) -> Data:
...
+        unique_sensors, counts = np.unique(
+            x[:, self._id_columns], axis=0, return_counts=True
+        )
+        # sort DOMs and pulse-counts
+        sort_this = np.concatenate(
+            [unique_sensors, counts.reshape(-1, 1)], axis=1
+        )
+        sort_this = lex_sort(x=sort_this, cluster_columns=self._id_columns)
+        unique_sensors = sort_this[:, 0 : unique_sensors.shape[1]]
+        counts = sort_this[:, unique_sensors.shape[1] :].flatten().astype(int)
```
RasmusOrsoe, line 294: So -  what is `cutter`?

RasmusOrsoe (commented):
Hey @Aske-Rosted - thanks for this PR.

I have left a few comments and questions; most important of those are on the subject of re-usability of the new cool methods you introduce here. As the code stands, it almost entirely depends on a single node definition with custom fields injected into the graph objects, and I think with a little bit of refactoring, we could make sure that your new code will play well with existing modules. Please give a read and let me know what you think!

src/graphnet/models/gnn/RNN_tito.py
```diff
@@ -0,0 +1,125 @@
...
+            post_processing_layer_sizes=self._post_processing_layer_sizes,
+            readout_layer_sizes=self._readout_layer_sizes,
+            n_head=self._n_head,
+            nb_neighbours=self._nb_neighbours,
+        )
+
+    def forward(self, data: Data) -> torch.Tensor:
+        """Apply learnable forward pass of the RNN and tito model."""
+        data = self._rnn(data)
+        readout = self._dynedge_tito(data)
```
RasmusOrsoe, line 123: `self._dynedge_tito` could in principle be any of our existing backbones, right?

Aske-Rosted: yes.

src/graphnet/models/graphs/nodes/nodes.py
Aske-Rosted: So the whole reason for this is due to the data objects inability to store jagged arrays, which arise from the differing number of DOM activations within an event. the cutter is an object that contains the number of DOM activations pr dom this enables us to concatenate all the time-series data and still be able to separate them into after loading into the model. I agree that this is not a very elegant solution and kinda _hacky_, however I did not find a better way of doing it, which was allowed within the framework of the **pytorch_geometric Data** object.

Aske-Rosted: If you have any other ideas of how to approach this I would love to get your input.

src/graphnet/models/rnn/node_rnn.py
Aske-Rosted: I feel like it makes it more obvious at a glance to have it named **RNN_dropout**. Considering future models that might have dropout layers in different sub-modules then we would have to resort to this kind of naming convention anyways.

RasmusOrsoe: Alright, but then lets rename it to `rnn_dropout` to be consistent with the snake_case naming scheme in graphnet.

src/graphnet/models/graphs/nodes/nodes.py
RasmusOrsoe: I'm happy to provide some concrete suggestions on how to achieve this, but I will need a detailed description on what these fields actually contain:

examples/04_training/05_train_RNN_TITO.py
```diff
@@ -0,0 +1,266 @@
...
+                azimuth_key="injection_azimuth", zenith_key="injection_zenith"
+            )
+        },
+    )
+
+    # Building model
+    backbone = RNN_TITO(
+        nb_inputs=graph_definition.nb_outputs,
+        nb_neighbours=8,
+        RNN_layers=2,
```
ArturoLlorente, line 119: input args not updated with `camel_case`.

src/graphnet/models/rnn/node_rnn.py
```diff
@@ -0,0 +1,134 @@
...
+
+
+class Node_RNN(GNN):
+    """Implementation of the RNN model architecture.
+
+    The model takes as input the typical DOM data format and transforms it into
+    a time series of DOM activations pr. DOM. before applying a RNN layer and
+    outputting the an RNN output for each DOM. This model is in its current
+    state not intended to be used as a standalone model. Furthermore, it needs
+    to be used with a time-series dataset and a "cutter" (see
```
RasmusOrsoe, line 24: Doc string needs to be updated

src/graphnet/models/rnn/node_rnn.py
RasmusOrsoe: Still an issue

RasmusOrsoe (commented):
Hey @Aske-Rosted! I think your recent changes are great. Only missing things are now some superficial things in the doc strings and variable naming. See the remaining comments.

examples/04_training/05_train_RNN_TITO.py
```diff
@@ -0,0 +1,266 @@
...
+        RNN_hidden_size=64,
+        RNN_dropout=0.5,
```
RasmusOrsoe: `rnn_..`

RasmusOrsoe (approved):
🚀

PR #656, opened 2024-02-03
bump requirement files to `torch=2.2`

RasmusOrsoe:
This PR changes the torch version suggested in the `requirements/` to `2.2`, which was released a few days ago. Without this change, automated workflows will fail.


Aske-Rosted (approved):
Nothing to add.

kayla-leonard:
When trying to use graphnet after re-installing, I get a segfault because of a version mis-match in the torch gpu requirements. I updated the version to match in the requirements and the issue is resolved. See #664

PR #658, opened 2024-02-08
`predict_as_dataframe` bugfix + syntax error in `minkowski.py`

### LINKED ISSUE #654: `predict_as_dataframe: ValueError`
**Describe the bug**
When dataloaders are construction with a collate function that removes events from batches, like our default collate function does, the automated inference of node-level and graph-level attributes fail.

**To Reproduce**
1. Pick any training example
2. Overwrite the standard collate function with one that removes at least one event in the dataset.
3. run `model.predict_as_dataframe(dataloader)`

**Expected behavior**
The automated inference of whether the quantities in "`additional_attributes`" are node-level or graph-level variables should work without error.

**Full traceback**
```
  File "/lustre/hpc/icecube/stuttard/workspace/fridge/processing/samples/oscNext/selection/level6_GNN/L6_dynedge_predict.py", line 49, in <module>
    model.predict(
  File "/lustre/hpc/icecube/stuttard/workspace/fridge/processing/samples/oscNext/selection/level6_GNN/L6_dynedge_model.py", line 1153, in predict
    results_val = self.model.predict_as_dataframe(val_dataloader, **predict_kw)
  File "/lustre/hpc/icecube/stuttard/workspace/icetray/oscNext/ext/graphnet/src/graphnet/models/standard_model.py", line 458, in predict_as_dataframe
    data = np.concatenate(
  File "<__array_function__ internals>", line 200, in concatenate
ValueError: all the input array dimensions except for the concatenation axis must match exactly, but along dimension 0, the array at index 0 has size 999 and the array at index 1 has size 12459
```

**Additional context**

I believe the source of the error can be found in the following lines of code in `predict_as_dataframe`:

```python
for batch in dataloader:
            for attr in attributes:
                attribute = batch[attr]
                if isinstance(attribute, torch.Tensor):
                    attribute = attribute.detach().cpu().numpy()

                # Check if node level predictions
                # If true, additional attributes are repeated
                # to make dimensions fit
                if len(predictions) != len(dataloader.dataset):
                    if len(attribute) < np.sum(
                        batch.n_pulses.detach().cpu().numpy()
                    ):
                        attribute = np.repeat(
                            attribute, batch.n_pulses.detach().cpu().numpy()
                        )
                        try:
                            assert len(attribute) == len(batch.x)
                        except AssertionError:
                            self.warning_once(
                                "Could not automatically adjust length"
                                f"of additional attribute {attr} to match length of"
                                f"predictions. Make sure {attr} is a graph-level or"
                                "node-level attribute. Attribute skipped."
                            )
                            pass
                attributes[attr].extend(attribute)
```
Specifially the check  `if len(predictions) != len(dataloader.dataset)` is troublesome here, because if a `collate_fn` have removed one or more events from the batch, this check will return `True`, resulting in the error.

RasmusOrsoe:
This PR does two things

1. Fixes #654 
2. Fixes what appears to be a lurking syntax error in `graphnet/models/graphs/edges/minkowski.py` which have been making my local mypy upset recently. This was introduced along with the original PR in #630 but surprisingly the tests passed just fine there. The code in question is
```python
distance_mat = compute_minkowski_distance_mat(
                x_masked := x[batch][mask[batch]],
                x_masked,
                self.c,
                self.space_coords,
                self.time_coord,
            )
```
specially ` x_masked := x[batch][mask[batch]]`. This PR changes those lines to:
```python
x_masked = x[batch][mask[batch]]
distance_mat = compute_minkowski_distance_mat(
    x=x_masked,
    y=x_masked,
    c=self.c,
    space_coords=self.space_coords,
    time_coord=self.time_coord,
)
```
which I believe is the intended arguments. Perhaps @AMHermansen can confirm?

### DIFF
--- a/requirements/torch_cpu.txt
+++ b/requirements/torch_cpu.txt
@@ -1,2 +1,2 @@
 --find-links https://download.pytorch.org/whl/cpu
---find-links https://data.pyg.org/whl/torch-2.1.0+cpu.html
+--find-links https://data.pyg.org/whl/torch-2.2.0+cpu.html
--- a/requirements/torch_gpu.txt
+++ b/requirements/torch_gpu.txt
@@ -1,4 +1,4 @@
 # Contains packages requirements for GPU installation
 --find-links https://download.pytorch.org/whl/torch_stable.html
 torch==2.1.0+cu118
---find-links https://data.pyg.org/whl/torch-2.1.0+cu118.html
+--find-links https://data.pyg.org/whl/torch-2.2.0+cu118.html
--- a/requirements/torch_macos.txt
+++ b/requirements/torch_macos.txt
@@ -1,2 +1,2 @@
 --find-links https://download.pytorch.org/whl/torch_stable.html
---find-links https://data.pyg.org/whl/torch-2.1.0+cpu.html
+--find-links https://data.pyg.org/whl/torch-2.2.0+cpu.html
--- a/src/graphnet/models/graphs/edges/minkowski.py
+++ b/src/graphnet/models/graphs/edges/minkowski.py
@@ -69,12 +69,13 @@ class MinkowskiKNNEdges(EdgeDefinition):
         row = []
         col = []
         for batch in range(x.shape[0]):
+            x_masked = x[batch][mask[batch]]
             distance_mat = compute_minkowski_distance_mat(
-                x_masked := x[batch][mask[batch]],
-                x_masked,
-                self.c,
-                self.space_coords,
-                self.time_coord,
+                x=x_masked,
+                y=x_masked,
+                c=self.c,
+                space_coords=self.space_coords,
+                time_coord=self.time_coord,
             )
             num_points = x_masked.shape[0]
             num_edges = min(self.nb_nearest_neighbours, num_points)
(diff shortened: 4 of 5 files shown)


AMHermansen:
> specially ` x_masked := x[batch][mask[batch]]`. This PR changes those lines to:
Is an "assignment expression" introduced with PEP 572, in python 3.8. Which is also our lowest support python version.

> ```python
> x_masked = x[batch][mask[batch]]
> distance_mat = compute_minkowski_distance_mat(
>     x=x_masked,
>     y=x_masked,
>     c=self.c,
>     space_coords=self.space_coords,
>     time_coord=self.time_coord,
> )
> ```
> 
> which I believe is the intended arguments. Perhaps @AMHermansen can confirm?
The proposed change would run the same way as the current code. I'm fine with changing it.

In regards to your local mypy is it possible that it believes you're running a python3.7 version? That would explain why it complains about the syntax.

RasmusOrsoe:
@AMHermansen thanks for the clarification! I would be in favor of keeping the proposed change here.

RasmusOrsoe:
@AMHermansen @Aske-Rosted If there are no further comments, could we get this approved?

Aske-Rosted (approved):

PR #660, opened 2024-02-11
Icemix kaggle solution

ArturoLlorente:
Implement models defined in Kaggle second solution. [icecube-2nd-place](https://github.com/DrHB/icecube-2nd-place). The following was performed:

 - Added `hlc` key to iceCube detector.
 - Modified `DynEdge` class to be able to fit new model definition.
 - Added new node definition `IceMixNodes` to calculate ice properties and random sampling of larger events.
 - Both GNN classes `DeepIce` and `DeepIceWithDynEdge` to reproduce different 5 models used in the competition. New PR with example and different model definition will be prepared.


RasmusOrsoe:
@ArturoLlorente Glad to see the PR is up! However, before we can review this, we need the unit tests to pass. It looks like the introduction of HLC as an additional variable makes some of the unit tests fail. If you need help addressing these, please let me know.

src/graphnet/models/gnn/icemix.py
```diff
@@ -0,0 +1,240 @@
...
+)
+from graphnet.models.gnn.dynedge import DynEdge
+from graphnet.models.gnn.gnn import GNN
+
+from torch_geometric.utils import to_dense_batch
+from torch_geometric.data import Data
+from torch import Tensor
+
+
+def convert_data(data: Data) -> tuple:
```
RasmusOrsoe, line 28: Return type should be `Tuple[Tensor, Tensor, Tensor]`

src/graphnet/models/gnn/icemix.py
```diff
@@ -0,0 +1,240 @@
...
+    Block,
+)
+from graphnet.models.gnn.dynedge import DynEdge
+from graphnet.models.gnn.gnn import GNN
+
+from torch_geometric.utils import to_dense_batch
+from torch_geometric.data import Data
+from torch import Tensor
+
+
```
RasmusOrsoe, line 27: I think this is a really cool function and that it could be a valuable utility function outside the context of this new model. 

I've refactored the function slightly and added additional doc strings to make it clearer what it actually does. We could consider moving this to `graphnet.models.utils.py` instead of having it in this file.

 Please see below a suggestion on how this function could look like. 

```python
def array_to_sequence(x: Tensor, 
                      batch_idx: LongTensor, 
                      padding_value: Tensor = torch.inf) -> Tuple[Tensor, 
                                                                  Tensor, 
                                                                  Tensor]:
    """Convert `x` of shape [n,d] into a padded sequence of shape [B, L, D].

       Where B is the batch size, L is the sequence length and D is the 
       features for each time step.

    Args:
        x: array-like tensor with shape `[n,d]` where `n` is the total number 
        of pulses in the batch and `d` is the number of  node features.
        batch_idx: a LongTensor identifying which row in `x` belongs to
                 which training example. 
                 E.g. `torch_geometric.data.Batch.batch`.
    Returns:
        x: Padded sequence with dimensions  [B, L, D].
        mask: A tensor that identifies masked entries in `x`.
               E.g. : `masked_entries = x[mask]`
        seq_length: A tensor containing the number of pulses in each event.
    """
 
    _, seq_length = torch.unique(batch_idx, return_counts=True)
    x_list = torch.split(x, seq_length.tolist())

    x = torch.nn.utils.rnn.pad_sequence(
        x_list, batch_first=True, padding_value=padding_value
    )
    mask = torch.ne(x[:, :, 1], torch.inf)
    x[~mask] = 0
    return x, mask, seq_length
```

src/graphnet/models/graphs/nodes/nodes.py
```diff
@@ -211,3 +212,125 @@ def _construct_nodes(self, x: torch.Tensor) -> Data:
...
+        Args:
+            input_feature_names: Column names for input features. Minimum
+            required features are z coordinate and hlc column names.
+            max_pulses: Maximum number of pulses to keep in the event.
+            z_name: Name of the z-coordinate column.
+            hlc_name: Name of the `Hard Local Coincidence Check` column.
+        """
+        super().__init__(input_feature_names=input_feature_names)
+
+        if input_feature_names is None:
```
RasmusOrsoe, line 243: You are hardcoding these variables because you need to construct a feature index in the `__init__` method that you can later search through  in `_construct_nodes`, right?

If so, I think this is a great hint that we should look into changing the interplay between `GraphDefinition` and `NodeDefinition` so this information is available by default. Because as your code is written, it will fail if this model is run on a different detector class without passing the input features explicitly. I think it's fair to not change this in this PR as it's already quite large, but we should follow this up soon.

src/graphnet/models/graphs/utils.py
```diff
@@ -158,3 +163,36 @@ def cluster_summarize_with_percentiles(
...
+        f_absorption: Function that takes a normalized depth
+                      and returns the corresponding normalized
+                      absorption length.
+    """
+    # Data from page 31 of https://arxiv.org/pdf/1301.5361.pdf
+    # Datum is from footnote 8 of page 29
+    df = pd.read_parquet(
+        os.path.join(DATA_DIR, "ice_properties/ice_transparency.parquet"),
+    )
+    df["z"] = df["depth"] - datum
```
RasmusOrsoe, line 190: I think the name "datum" is incredible confusing and I highly suspect it's the result of a machine translation going wrong. I realize you did not name this variable, but let's take this opportunity to get it right.  

From the code, I think it's clear that this variable is an offset in the scaling `z = (depth - offset)/500`. I believe this is the 2nd place teams attempt to transform global coordinates into icecube coordinates, such that the resulting interpolation functions can be evaluated using the scaled z-coordinate of the pulses. I.e `z = depth - offset` gives you the z-coordinate in icecube coordinates and the `/500` gives you the standardized feature, identical to the standardized z-coordinate of the pulses ( see the transform [here](https://github.com/graphnet-team/graphnet/blob/ee2addc6f495c68262f3fe36bea4a36b4ccabd98/src/graphnet/models/detector/icecube.py#L35)) . This means, however, as this code is written, that this function expects a very specific standardization in order to be correct (`/500`), which I think is problematic, because if one were to pass the `IceCubeDeepCore` detector instead, the scaling is (`/100`) see [here](https://github.com/graphnet-team/graphnet/blob/ee2addc6f495c68262f3fe36bea4a36b4ccabd98/src/graphnet/models/detector/icecube.py#L93).

src/graphnet/models/graphs/utils.py
```diff
@@ -158,3 +163,36 @@ def cluster_summarize_with_percentiles(
...
+                      and returns the corresponding normalized
+                      scattering length.
+        f_absorption: Function that takes a normalized depth
+                      and returns the corresponding normalized
+                      absorption length.
+    """
+    # Data from page 31 of https://arxiv.org/pdf/1301.5361.pdf
+    # Datum is from footnote 8 of page 29
+    df = pd.read_parquet(
+        os.path.join(DATA_DIR, "ice_properties/ice_transparency.parquet"),
```
RasmusOrsoe, line 188: I don't think this file is included in the PR

src/graphnet/models/graphs/utils.py
RasmusOrsoe: I think for the purposes of this PR, you could just slightly change the function - to make it clear that this is a potential problem. Like so:

```python
def ice_transparency() -> Tuple[interp1d, interp1d]:
    """ Returns interpolation functions for optical properties of IceCube.

        NOTE: The resulting interpolation functions assumes that the 
        Z-coordinate of pulse are scaled as `z = z/500`. 
        Any deviation from this scaling method results in inaccurate results. 

    Returns:
        f_scattering: Function that takes a normalized depth
                      and returns the corresponding normalized
                      scattering length.
        f_absorption: Function that takes a normalized depth
                      and returns the corresponding normalized
                      absorption length.
    """
    # Data from page 31 of https://arxiv.org/pdf/1301.5361.pdf
    # Datum is from footnote 8 of page 29
    df = pd.read_parquet(
        os.path.join(DATA_DIR, "ice_properties/ice_transparency.parquet"),
    )
    df["z"] = df["depth"] - 1950 # Should actually be 1948.2
    df["z_norm"] = df["z"] / 500
    df[
        ["scattering_len_norm", "absorption_len_norm"]
    ] = RobustScaler().fit_transform(df[["scattering_len", "absorption_len"]])

    f_scattering = interp1d(df["z_norm"], df["scattering_len_norm"])
    f_absorption = interp1d(df["z_norm"], df["absorption_len_norm"])
    return f_scattering, f_absorption
```

I think that would be OK as we currently do not have an easy way of including this information. In a future iteration, we can think of a more structured way of including this information, and to deal with the hard-coded feature names.

src/graphnet/models/gnn/icemix.py
```diff
@@ -0,0 +1,240 @@
...
+                    num_heads=dim // head_size,
+                    mlp_ratio=4,
+                    drop_path=0.0 * (i / (depth - 1)),
+                    init_values=1,
+                )
+                for i in range(depth)
+            ]
+        )
+        self.dyn_edge = DynEdge(
+            nb_inputs=9,  # 9 features, TBD by user
```
RasmusOrsoe, line 186: Should this not be an input to the model?

src/graphnet/models/gnn/icemix.py
```diff
@@ -0,0 +1,240 @@
...
+    x_list = torch.split(data.x, seq_length.tolist())
+    x = torch.nn.utils.rnn.pad_sequence(
+        x_list, batch_first=True, padding_value=torch.inf
+    )
+    mask = torch.ne(x[:, :, 1], torch.inf)
+    x[~mask] = 0
+    return x, mask, seq_length
+
+
+class DeepIce(GNN):
```
RasmusOrsoe, line 50: These two model classes are very similar in content, and I think we should merge them into one. Here's a pseudo-code suggestion on how to do so:

```python
class DeepIce(GNN):
    """DeepIce model."""

    def __init__(
        self,
        dim: int = 384,
        dim_base: int = 128,
        depth: int = 12,
        head_size: int = 32,
        depth_rel: int = 4,
        n_rel: int = 1,
        include_dynedge: bool = False,
        dynedge_args: Dict[str, Any] =  None,
        
    ):
        """Construct `DeepIce`.

        Args:
            dim: The latent feature dimension.
            dim_base: The base feature dimension.
            depth: The depth of the transformer.
            head_size: The size of the attention heads.
            depth_rel: The depth of the relative transformer.
            n_rel: The number of relative transformer layers to use.
            include_dynedge: If True, pulse-level predictions from `DynEdge`
                            will be added as features to the model.
                            Defaults to False.
            dynedge_args: Arguments to initialize DynEdge with.
                          If not given, DynEdge will be initalized with the
                          arguments that the model was originally used with
                          during the Neutrinos In Deep Ice Kaggle Competition.
                          If `include_dynedge` is False, this argument have
                          no impact. 
        """
        super().__init__(dim_base, dim)
        self.fourier_ext = FourierEncoder(dim_base, dim)
        self.rel_pos = SpacetimeEncoder(head_size)
        self.sandwich = nn.ModuleList(
            [
                Block_rel(dim=dim, num_heads=dim // head_size)
                for i in range(depth_rel)
            ]
        )
        self.cls_token = nn.Linear(dim, 1, bias=False)
        self.blocks = nn.ModuleList(
            [
                Block(
                    dim=dim,
                    num_heads=dim // head_size,
                    mlp_ratio=4,
                    drop_path=0.0 * (i / (depth - 1)),
                    init_values=1,
                )
                for i in range(depth)
            ]
        )
        self.n_rel = n_rel
        self._include_dynedge = include_dynedge
        if self._include_dynedge & dynedge_args is None:
            self.dynedge = DynEdge(
                nb_inputs=self._nb_inputs,  # 9 features, TBD by user
                nb_neighbours=nb_neighbours,
                post_processing_layer_sizes=[336, dim // 2],
                dynedge_layer_sizes=[
                    (128, 256),
                    (336, 256),
                    (336, 256),
                    (336, 256),
                ],
                global_pooling_schemes=None,
                icemix_encoder=True,
            )
        elif self._include_dynedge & dynedge_args is not None:
            self.dynedge = DynEdge(
                **dynedge_args
            )

    @torch.jit.ignore
    def no_weight_decay(self) -> Set:
        """cls_tocken should not be subject to weight decay during training."""
        return {"cls_token"}

    def forward(self, data: Data) -> Tensor:
        """Apply learnable forward pass."""
        x0, mask, seq_length = convert_data(data)
        x = self.fourier_ext(x0, seq_length)
        rel_pos_bias = self.rel_pos(x0)
        batch_size = mask.shape[0]
        attn_mask = torch.zeros(mask.shape, device=mask.device)
        attn_mask[~mask] = -torch.inf

        if self._include_dynedge:
            graph = self.dyn_edge(data)
            graph, _ = to_dense_batch(graph, data.batch)
            x = torch.cat([x, graph], 2)

        for i, blk in enumerate(self.sandwich):
            x = blk(x, attn_mask, rel_pos_bias)
            if self._include_dynedge:
                rel_pos_bias = None
            else:
                if i + 1 == self.n_rel:
                    rel_pos_bias = None

        mask = torch.cat(
            [
                torch.ones(
                    batch_size, 1, dtype=mask.dtype, device=mask.device
                ),
                mask,
            ],
            1,
        )
        attn_mask = torch.zeros(mask.shape, device=mask.device)
        attn_mask[~mask] = -torch.inf
        cls_token = self.cls_token.weight.unsqueeze(0).expand(
            batch_size, -1, -1
        )
        x = torch.cat([cls_token, x], 1)

        for blk in self.blocks:
            x = blk(x, None, attn_mask)

        return x[:, 0]
```

src/graphnet/models/gnn/dynedge.py
```diff
@@ -65,6 +66,10 @@ def __init__(
                 after global pooling. The alternative is to  added (distribute)
                 them to the individual nodes before any convolutional
                 operations.
+            icemix_encoder: Whether to use the IceCubeMix encoder. If `True`,
```
RasmusOrsoe, line 69: I think we should split this into three separate args:

```python
activation : Callable = torch.nn.LeakyReLU()
add_layer_norm: bool = False
skip_readout: bool = False
```
The kaggle-version of this model would then be a specific set of args, i.e `activation = torch.nn.GELU()`, `add_layer_norm = True` and `skip_readout = True`.

RasmusOrsoe (commented):
Hey @ArturoLlorente  - thank you for this much anticipated contribution! I think the PR is in good shape, but here is a few places where I have added suggestions. Please let me know if anything is unclear.

src/graphnet/models/gnn/icemix.py
ArturoLlorente: Nice comment, I think this is a better way to structure the function. However, I am a but concern about the padding value.
However, if the user chooses to specify a padding value, it should look like `x[~mask] = padding_value`, since this padding_value is the value that will be used to pad the output Tensor x.
For the intermediate step I chosed `torch.inf` to avoid false possitives when using `torch.ne`.
Some pseudo-code would be like this, let me know your comments:

```python
def array_to_sequence(x: Tensor, 
                      batch_idx: LongTensor, 
                      padding_value: Any = 0,
                      excluding_value: Any = torch.inf) -> Tuple[Tensor, 
                                                                  Tensor, 
                                                                  Tensor]:
    """Convert `x` of shape [n,d] into a padded sequence of shape [B, L, D].

       Where B is the batch size, L is the sequence length and D is the 
       features for each time step.

    Args:
        x: array-like tensor with shape `[n,d]` where `n` is the total number 
        of pulses in the batch and `d` is the number of  node features.
        batch_idx: a LongTensor identifying which row in `x` belongs to
                which training example. 
                E.g. `torch_geometric.data.Batch.batch`.
        padding_value: The value to use for padding.
        excluding_value: This parameter represents a unique value that should
                not be present in the input tensor 'x'
    Returns:
        x: Padded sequence with dimensions  [B, L, D].
        mask: A tensor that identifies masked entries in `x`.
               E.g. : `masked_entries = x[mask]`
        seq_length: A tensor containing the number of pulses in each event.
    """
    
    if torch.any(torch.eq(x, excluding_value)):
        raise ValueError(f'Transformation cannot be made because input tensor '
                         f'`x` contains at least one element equal to '
                         f'excluding value {excluding_value}.')
    
    _, seq_length = torch.unique(batch_idx, return_counts=True)
    x_list = torch.split(x, seq_length.tolist())
    
    x = torch.nn.utils.rnn.pad_sequence(
        x_list, batch_first=True, padding_value=excluding_value
    )
    mask = torch.ne(x[:, :, 1], excluding_value)
    x[~mask] = padding_value
    return x, mask, seq_length
```

src/graphnet/models/graphs/utils.py
ArturoLlorente: You are right, this function was barely modified compared to the one used in [icecube-2nd-place](https://github.com/DrHB/icecube-2nd-place).

I share your concern about the lack of generalization of the ice properties in general. 
Maybe it is a good idea to make the small changes you suggested, and have it deleted before the _GraphNet 2.0.0_ release

ChenLi2049:
Hi, thanks for the great work. I have few suggestions on naming:

1. Instead of `dim`, it can be `hidden_dim` or `d_model`, which is commonly used for Transformer models. I personally prefer `hidden_dim`, which provides more information.
2. Instead of `dim_base`, it can be `seq_length`, which is commonly used for Transformer models. And in IceCube, this is the number of pulses in an event, thus "sequence length". This naming convention is more intuive in my opinion.

RasmusOrsoe (approved):
🚀

PR #661, opened 2024-02-12
define SWA and EMA model

ArturoLlorente:
Stochastic Weight Averaging (SWA) is a regularization technique in PyTorch that improves the generalization performance of deep neural networks by computing the average of multiple weights characterized by different points of the optimization trajectory. It is better explained in the article: [PyTorch 1.6 now includes Stochastic Weight Averaging](https://pytorch.org/blog/pytorch-1.6-now-includes-stochastic-weight-averaging/).

In this model, it is also possible to define EMA. Exponential Moving Average (EMA) is a widely known technique to reduce the training time by reducing the number of weight updates needed.

Implementation was made following the example: [SWA and EMA pytorch](https://pytorch.org/docs/stable/optim.html#weight-averaging-swa-and-ema)

### DIFF
--- a/src/graphnet/models/__init__.py
+++ b/src/graphnet/models/__init__.py
@@ -10,3 +10,4 @@ GNN
 
 from .model import Model
 from .standard_model import StandardModel
+from .standard_averaged_model import StandardAveragedModel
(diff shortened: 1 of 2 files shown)


RasmusOrsoe:
@ArturoLlorente could you update this PR with the latest changes from `main`? Then I'll review

RasmusOrsoe (approved):
🚀 !

PR #662, opened 2024-02-17
I3 converter bugs

Aske-Rosted:
I encountered two bugs when trying to create files with the _new_ general dataconverter.

1. The `output_files` list was instantiated in a way where the folder for the output files where checked for existing files before these had been created. 
2. The `_create_file_name` were created in a way that did not allow for multiple "." to be in the input file name (which is often the case for our I3 files).
3. The `create_table_and_save_to_sql` in ` _merge_databases` for the `SQLiteWriter` was instantiated without a `default_type`, which means that it defaults to `NOT NULL` effectively not allowing for missing values when merging the databases. 

This pull request should fix these issues.

### DIFF
--- a/src/graphnet/data/dataconverter.py
+++ b/src/graphnet/data/dataconverter.py
@@ -165,7 +165,7 @@ class DataConverter(ABC, Logger):
         for ext in self._file_reader._accepted_file_extensions:
             if file_name.endswith(ext):
                 file_name_without_extension = file_name.replace(ext, "")
-        return file_name_without_extension
+        return file_name_without_extension.replace(".i3", "")
 
     @final
     def _assign_event_no(
--- a/src/graphnet/data/dataconverter.py
+++ b/src/graphnet/data/dataconverter.py
@@ -89,11 +89,14 @@ class DataConverter(ABC, Logger):
         # in the directory
         input_files = self._file_reader.find_files(path=input_dir)
         self._launch_jobs(input_files=input_files)
-        self._output_files = glob(
+        self._output_files = [
             os.path.join(
-                self._output_dir, f"*{self._save_method.file_extension}"
+                self._output_dir,
+                self._create_file_name(file)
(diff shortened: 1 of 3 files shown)


RasmusOrsoe:
@Aske-Rosted Thanks for sharing this. I noticed that there are a few unit tests that are not passing. Could you take a look at those?

Aske-Rosted:
> @Aske-Rosted Thanks for sharing this. I noticed that there are a few unit tests that are not passing. Could you take a look at those?

I have tried to take a look at them they are all about "ModuleNotFoundError: No module named 'torch_geometric.utils.homophily'" which is not a module I have introduced in this PR so I do not really understand why these are showing up here...

RasmusOrsoe:
@Aske-Rosted I've pushed a fix here #663

Aske-Rosted:
> I encountered two bugs when trying to create files with the _new_ general dataconverter.
> 
>     1. The `output_files` list was instantiated in a way where the folder for the output files where checked for existing files before these had been created.
> 
>     2. The `_create_file_name` were created in a way that did not allow for multiple "." to be in the input file name (which is often the case for our I3 files).
> 
>     3. The `create_table_and_save_to_sql` in ` _merge_databases` for the `SQLiteWriter` was instantiated without a `default_type`, which means that it defaults to `NOT NULL` effectively not allowing for missing values when merging the databases.
> 
> 
> This pull request should fix these issues.

I found one more issue for which I implemented a solution. The Issue was that using the Generic extractor to try and extract a value from an I3PODHolder (https://software.icecube.wisc.edu/icetray/main/projects/icetray/i3podholder.html) would not be extracted properly. The data would per table be extracted as a python dict with with the key being "value". The solution I have implemented catches this exact instance and extracts the value of the dict.

RasmusOrsoe (approved):
@Aske-Rosted Thank you for this!

PR #664, opened 2024-02-23
Update torch version in gpu requirements to fix mis-match (2.2)

kayla-leonard:
See #656


RasmusOrsoe (approved):
@kayla-leonard thank you very much for correcting this oversight!

Issue #650, opened 2024-01-08
Implement more descriptive extra_repr methods for GraphNeT Models


AMHermansen:
**Is your feature request related to a problem? Please describe.**
Currently when printing a GraphNeT Model we default to the bare-bones pytorch `module.__repr__`, which only prints the layers in a module. Since all GraphNeT Models contain a config_dict with information for how it was constructed we can provide a much more detailed description of the object print. This is especially useful for models containing no learnable parameters like `Detector` and `GraphDefinition`.

For example when trying to print a `GraphDefinition` object you get `GraphDefinition()` regardless of how the GraphDefinition was constructed.

**Describe the solution you'd like**
Implement an `extra_repr` method in Model which provide a more detailed description of the object.

samadpls:
The following is the output produced by running the extra_reprt implementation:

```
GraphDefinition(
  GraphDefinition(
      class_name=GraphDefinition
      arguments={'detector': ModelConfig(class_name='IceCube86', arguments={}), 'node_definition': ModelConfig(class_name='NodesAsPulses', arguments={'input_feature_names': None}), 'edge_definition': None, 'input_feature_names': None, 'dtype': torch.float32, 'perturbation_dict': None, 'seed': None, 'add_inactive_sensors': False, 'sensor_mask': None, 'string_mask': None, 'sort_by': None}
  )
  (_detector): IceCube86()
  (_node_definition): NodesAsPulses()
)
```

The implementation used to generate the output is as follows:

```python
def extra_repr(self) -> str:
    """Provide a more detailed description of the object print.

    Returns:
        str: A string representation containing detailed information about the object.
    """
    full_str = f"{self.__class__.__name__}(\n"
    for item, value in self._config.__dict__.items():
        if isinstance(value, Model):
            full_str += self._predindent_args(value)
        else:
            full_str += f"    {item}={value}\n"
    full_str += ")"
    return full_str


def _predindent_args(self, model: Model) -> str:
    """Indent nested model arguments.

    Args:
        model (Model): The nested model.

    Returns:
        str: Indented string representation of the nested model's arguments.
    """
    indented_str = model.extra_repr().replace("\n", "\n    ")
    return f"    {indented_str}\n"
```

here is how i create the code

```python
from graphnet.models.graphs.graph_definition import GraphDefinition
from graphnet.models.detector import IceCube86

graph_def = GraphDefinition(detector=IceCube86())
print(graph_def)
```

AMHermansen:
[...]

PR #665, opened 2024-02-23
Added `extra_repr` and `__repr__ ` methods to `Model` and `ModelConfig` classes

samadpls:
Implemented `extra_repr` and `__repr__` for GraphDefinition and ModelConfig classes
Resolved issue #650 
While Running Following Code
```python
from graphnet.models.graphs.graph_definition import GraphDefinition
from graphnet.models.detector import IceCube86

graph_def = graph_definition=GraphDefinition(detector=IceCube86())
print(graph_def)
```
### output
```bash
GraphDefinition(
  GraphDefinition(
  {
      'arguments': {
          'detector': ModelConfig(
  {
  
  }
  ),
          'node_definition': ModelConfig(
  {
      'input_feature_names': None,
  }
  ),
          'edge_definition': None,
          'input_feature_names': None,
          'dtype': torch.float32,
          'perturbation_dict': None,
          'seed': None,
          'add_inactive_sensors': False,
          'sensor_mask': None,
          'string_mask': None,
          'sort_by': None,
      },
  })
  (_detector): IceCube86()
  (_node_definition): NodesAsPulses()
)
```

### DIFF
--- a/src/graphnet/models/graphs/graph_definition.py
+++ b/src/graphnet/models/graphs/graph_definition.py
@@ -447,3 +447,28 @@ class GraphDefinition(Model):
         for key, fn in custom_label_functions.items():
             graph[key] = fn(graph)
         return graph
+
+    def extra_repr_recursive(self, dictionary: dict, indent: int = 4) -> str:
+        """Recursively format a dictionary for extra_repr."""
+        result = "{\n"
+        for key, value in dictionary.items():
+            if key == "class_name":
+                continue
+            result += " " * indent + f"'{key}': "
+            if isinstance(value, dict):
+                result += self.extra_repr_recursive(value, indent + 4)
+            elif isinstance(value, Model):
+                result += value.__repr__()
+            else:
+                result += repr(value)
+            result += ",\n"
+        result += " " * (indent - 4) + "}"
+        return result
+
+    def extra_repr(self) -> str:
+        """Provide a more detailed description of the object print.
+
+        Returns:
+            str: A string representation containing detailed information about the object.
+        """
+        return f"{self.__class__.__name__}(\n{self.extra_repr_recursive(self._config.__dict__)})"
--- a/src/graphnet/models/graphs/graph_definition.py
+++ b/src/graphnet/models/graphs/graph_definition.py
@@ -447,28 +447,3 @@ class GraphDefinition(Model):
         for key, fn in custom_label_functions.items():
             graph[key] = fn(graph)
         return graph
-
-    def extra_repr_recursive(self, dictionary: dict, indent: int = 4) -> str:
-        """Recursively format a dictionary for extra_repr."""
-        result = "{\n"
-        for key, value in dictionary.items():
-            if key == "class_name":
-                continue
-            result += " " * indent + f"'{key}': "
-            if isinstance(value, dict):
-                result += self.extra_repr_recursive(value, indent + 4)
-            elif isinstance(value, Model):
-                result += value.__repr__()
-            else:
-                result += repr(value)
-            result += ",\n"
-        result += " " * (indent - 4) + "}"
-        return result
-
-    def extra_repr(self) -> str:
-        """Provide a more detailed description of the object print.
-
-        Returns:
-            str: A string representation containing detailed information about the object.
-        """
-        return f"{self.__class__.__name__}(\n{self.extra_repr_recursive(self._config.__dict__)})"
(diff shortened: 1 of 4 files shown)


samadpls:
Hi @AMHermansen, Regarding:
> Could you also include how a `Model` object which contains learnable parameters is printed? I think one of the subclasses of `Task` would be nice, so we can see how this `extra_repr` works with the components from torch?

Here is the Output from running `test_model_config.py`
```bash
StandardModel(
  (_graph_definition): KNNGraph(
    KNNGraph(
    {
        'arguments': {
            'detector': ModelConfig(
    {
    
    }
    ),
            'node_definition': ModelConfig(
    {
        'input_feature_names': None,
    }
    ),
            'input_feature_names': ['dom_x', 'dom_y', 'dom_z', 'dom_time', 'charge', 'rde', 'pmt_area'],
            'dtype': torch.float32,
            'perturbation_dict': None,
            'seed': None,
            'nb_nearest_neighbours': 8,
            'columns': [0, 1, 2],
        },
    })
    (_detector): IceCubeDeepCore()
    (_edge_definition): KNNEdges()
    (_node_definition): NodesAsPulses()
  )
  (backbone): DynEdge(
    (_activation): LeakyReLU(negative_slope=0.01)
    (_conv_layers): ModuleList(
      (0): DynEdgeConv(nn=Sequential(
        (0): Linear(in_features=38, out_features=128, bias=True)
        (1): LeakyReLU(negative_slope=0.01)
        (2): Linear(in_features=128, out_features=256, bias=True)
        (3): LeakyReLU(negative_slope=0.01)
      ))
      (1-3): 3 x DynEdgeConv(nn=Sequential(
        (0): Linear(in_features=512, out_features=336, bias=True)
        (1): LeakyReLU(negative_slope=0.01)
        (2): Linear(in_features=336, out_features=256, bias=True)
        (3): LeakyReLU(negative_slope=0.01)
      ))
    )
    (_post_processing): Sequential(
      (0): Linear(in_features=1043, out_features=336, bias=True)
      (1): LeakyReLU(negative_slope=0.01)
      (2): Linear(in_features=336, out_features=256, bias=True)
      (3): LeakyReLU(negative_slope=0.01)
    )
    (_readout): Sequential(
      (0): Linear(in_features=1024, out_features=128, bias=True)
      (1): LeakyReLU(negative_slope=0.01)
    )
  )
  (_tasks): ModuleList(
    (0): EnergyReconstruction(
      (_loss_function): LogCoshLoss()
      (_affine): Linear(in_features=128, out_features=1, bias=True)
    )
  )
)
```

src/graphnet/models/graphs/graph_definition.py
```diff
@@ -447,3 +447,28 @@ def _add_custom_labels(
...
+        result += " " * (indent - 4) + "}"
+        return result
+
+    def extra_repr(self) -> str:
+        """Provide a more detailed description of the object print.
+
+        Returns:
+            str: A string representation containing detailed information about the object.
+        """
+        return f"{self.__class__.__name__}(\n{self.extra_repr_recursive(self._config.__dict__)})"
```
AMHermansen, line 474: I think perhaps it makes sense to move this into `Model`, that way we this more elaborate repr in most of our components. What are you thoughts?

samadpls:
Thanks for your suggestion. I think the new change with more details is great. I updated the code to match your recommendations.

samadpls:
Hi @AMHermansen, thanks for the suggestion. However, the test keeps failing even after making these changes until I modify the assert.
from
```python
assert repr(model) == repr(constructed_model)
```
to 
```python
assert constructed_model.extra_repr() == model.extra_repr()
```

AMHermansen:
> Hi @AMHermansen, thanks for the suggestion. However, the test keeps failing even after making these changes until I modify the assert. from
> 
> ```python
> assert repr(model) == repr(constructed_model)
> ```
> 
> to
> 
> ```python
> assert constructed_model.extra_repr() == model.extra_repr()
> ```

I believe this new assert would pass trivially since both `extra_repr` return the empty string.
I suspect the root cause, is that the `verbose_print` flag is not being applied recursively, so only the outermost `Model` object changes the flag.
I think the easiest way to fix this, is to implement a utility function, which recursively changes the `verbose_print` flag, I think something like,
```python
class Model(...):
    verbose_print = True

    def set_verbose_print_recursively(self, verbose_print: bool):
        for module in self.modules:
            if isinstance(module, Model):
                module.verbose_print = verbose_print
        self.verbose_print = verbose_print
```
Should work.

AMHermansen:
Looks good, I'll merge it :raised_hands:

samadpls:
thanks :)

PR #672, opened 2024-03-13
Weight norm per file.

Aske-Rosted:
Adds the option to normalize the weights of a database, such that the total sum of event weights in the file is equal to this number, useful when training using multiple databases.


RasmusOrsoe (approved):
🥇

PR #678, opened 2024-03-14
combined_extractor

Aske-Rosted:
This pr adds the `CombinedExtractor` class, which can be used to combine other extractor classes renaming the name of the extractor. For the SQLite databases this allows to save the information from multiple extractors into the same table.

### DIFF
--- a/src/graphnet/data/extractors/combine_extractors.py
+++ b/src/graphnet/data/extractors/combine_extractors.py
@@ -1,0 +1,34 @@
+"""Module for combining multiple extractors into a single extractor."""
+from graphnet.data.extractors.icecube.i3extractor import I3Extractor
+from typing import List, Dict
+
+from icecube import icetray
+
+
+class CombinedExtractor(I3Extractor):
+    """Class for combining multiple extractors.
+
+    This class is used to combine multiple extractors into a single extractor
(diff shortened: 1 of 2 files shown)


src/graphnet/data/extractors/combine_extractors.py
```diff
@@ -0,0 +1,34 @@
...
+
+    This class is used to combine multiple extractors into a single extractor
+    with a new name.
+    """
+
+    def __init__(self, extractors: List[I3Extractor], extractor_name: str):
+        """Construct CombinedExtractor.
+
+        Args:
+        extractors: List of extractors to combine.
```
RasmusOrsoe, line 19: I think we should mention explicitly here, that the user must provide a list of extractors that are compatible. E.g. that combining a truth table with a pulsemap will fail, because one contains event-level information while the other contains pulse-level information. E.g.:

`"extractors: List of extractors to combine. The extractors must all return data on the same level; e.g. all event-level data or pulse-level data. Mixing tables that contain event-level and pulse-level information will fail."`

RasmusOrsoe (changes requested):
I think this is a clever way of using the existing code! I left one comment related to doc strings.

PR #679, opened 2024-03-14
add_checkpoint_handling

Aske-Rosted:
When using check-pointing during training the save format is slightly different from the state dict, that is it has some additional information. This pr adds the ability to automatically deploy using a model checkpoint by extracting the state dict from the checkpoint.


RasmusOrsoe (approved):
👍

PR #680, opened 2024-03-31
 Integration of #660 and Example Definition

ArturoLlorente:
This PR continues the integration of #660 by defining an example and making necessary changes due to incompatibilities with the Prometheus (and any) detector. Now the model can be run without ice properties and without HLC feature.

Changes Made:

 - Flexible Embeddings: The embeddings have been modified to allow flexibility in feature input definition. Now, the number of features is defined by the user.
 - Updated Argument: The `activation` argument is now a string to resolve an error encountered while attempting to save the model into a config file.
 - Enhanced Event Subsampling: If no `hlc_name` is defined, the event subsampling will default to random.
 - Offset and scaling of the ice properties z coordinate now can be defined by the user.


examples/04_training/06_train_icemix_model.py
```diff
@@ -0,0 +1,277 @@
+"""Example of training Model.
+
+This example is based on Icemix solution proposed in
+https://github.com/DrHB/icecube-2nd-place.git (2nd place solution). There are a
+few things to note about this example:
```
RasmusOrsoe, line 5: Looks like you forgot to finish this docstring

examples/04_training/06_train_icemix_model.py
```diff
@@ -0,0 +1,277 @@
...
+    }
+
+    graph_definition = KNNGraph(
+        detector=Prometheus(),
+        node_definition=IceMixNodes(
+            input_feature_names=features,
+            max_pulses=128,
+            z_name="sensor_pos_z",
+            hlc_name=None,
+            add_ice_properties=True,
```
RasmusOrsoe, line 89: `prometheus` is a reference to  an open-source simulation of `ORCA150`, which is a water-based detector. So I think we should set `add_ice_properties = False` here, since you're training from scratch.

RasmusOrsoe (commented):
Thank you for adding this! I left a few superficial comments.

examples/04_training/06_train_icemix_model.py
ArturoLlorente: Removed. 
For further familiarization with the solution I think is enough with #681 .

RasmusOrsoe (approved):

PR #682, opened 2024-04-03
Integration of LiquidO

RasmusOrsoe:
This PR adds [LiquidO](https://neutrinos.portales.ciemat.es/liquido) as a supported experiment in GraphNeT.

The integration includes

- Data conversion code; `LiquidOReader` and associated `H5Extractor`s. An example folder for LiquidO is added.
- Detector Class `LiquidO_v1` - A preliminary detector configuration + geometry table

I also removed two print statements in `ParquetDataset` that snug past our previous review.


src/graphnet/data/constants.py
```diff
@@ -30,6 +30,7 @@ class FEATURES:
         "t",
     ]
     KAGGLE = ["x", "y", "z", "time", "charge", "auxiliary"]
+    LIQUIDO = ["sipmID", "sipm_x", "sipm_y", "sipm_z", "t", "var"]
```
Aske-Rosted, line 33: Would it not be better to keep x,y,z,t as the first entries for consistency?

Aske-Rosted (commented):
1 small comment otherwise looks fine.

src/graphnet/data/constants.py
RasmusOrsoe: Hey @Aske-Rosted! I would ordinarily agree, but `x` is a reserved name in our data representations - i.e. `data.x` gives us the entire input to models (see [here](https://github.com/graphnet-team/graphnet/blob/16bf0cede70f5575762d2e969ced1fb10ba9aa5b/src/graphnet/models/graphs/graph_definition.py#L435)). So it is actually a bit unfortunate that the naming in the competition used a generic xyzt naming scheme. Also, LiquidO does not have PMTs like our usual experiments do; they use scintillating fibers arranged in a grid-like structure, and the xyz currently represents the position of the silicon photomultiplier (SiPM) that read out the signal from a given fiber (one attached to each end), which make out points on "the boundary" of the detector and not inside the detector volume. I think adding `sipm_` is a fair way to make this distinction clearer.
image
(image from [here](https://neutrinos.portales.ciemat.es/liquido))

Aske-Rosted: Sorry I was a bit unclear I don't think there is any problem with actual names, bur thought that the order i.e having it be `LIQUIDO = [ "sipm_x", "sipm_y", "sipm_z","sipmID", "t", "var"]`
might be slightly more standard, although it is a very minor thing. I'll approve the PR and leave this at your discretion.

Aske-Rosted (approved):
looks good , see comment.

src/graphnet/data/constants.py
RasmusOrsoe: Sorry I misunderstood. I changed the order!

PR #687, opened 2024-04-04
remove pisa-related code

### LINKED ISSUE #671: Should we remove PISA-related code?
Currently, our library contain code that relies on PISA, an icecube software package for low energy oscillation analyses.

Because of closed data policies and execution times, this code cannot be subject to unit tests. In total, this is makes up around 1.200 lines of codes that we don't have tests for, which hurts our code coverage.

I think we should consider removing this code because:

- I don't think anyone is actively using it
- I don't think graphnet should contain code that we can't test
- The current fitting code in (https://github.com/graphnet-team/graphnet/blob/main/src/graphnet/pisa/fitting.py) doesn't  allow the users to include noise and muons, making it's usability very narrow. 
- Removing this code would contribute towards #102

RasmusOrsoe:
closes #671 by deleting pisa-related code.


Aske-Rosted (approved):
:scissors:

PR #689, opened 2024-04-09
`CuratedDataset` - A method for sharing pre-converted datasets

### LINKED ISSUE #545: Add repository of open datasets, and easy means to access/download them
Plan is to use University of Copenhagen's [ERDA](https://www.erda.dk/) for hosting and allowing for download/rsync using a python utility function.

Related to #546.

RasmusOrsoe:
This PR adds functionality for sharing pre-converted and configured datasets as importable `DataModule`s with possibly fixed train, val and test selections. Closes #545.  Intended usage is shown below:

```python
from graphnet.models.detector.prometheus import ORCA150
from graphnet.datasets import TestDataset

# Pick GraphDefinition
graph_definition = KNNGraph(detector = ORCA150())

# Instantiate CuratedDataset - It will download and unzip, build dataloaders
data_module = TestDataset(download_dir='',
                    graph_definition=graph_definition,
                    train_dataloader_kwargs = {'batch_size': 2,
                                                "num_workers": 1})

# Get description of Dataset - this is an auto-generated text using the class properties:
data_module.description()

# Get Dataloaders
train_dataloader = data_module.train_dataloader
val_dataloder = data_module.val_dataloader

# Go Train
```
The resulting terminal output is:

```
TestDataset contains data from ARCA Prometheus Simulation and was added to GraphNeT by Rasmus F. Ørsøe. 

 COMMENTS ON USAGE: 
 Rasmus F. Ørsøe: This Dataset should be used for unit tests only. Simulation produced by Stephan Meighen-Berger, U. Melbourne. 
 
 DATASET DETAILS: 
 pulsemaps: ['photons'] 
 truth table: mc_truth 
 input features: ['sensor_pos_x', 'sensor_pos_y', 'sensor_pos_z', 't']
 pulse truth: None 
 event truth: ['interaction', 'initial_state_energy', 'initial_state_type', 'initial_state_zenith', 'initial_state_azimuth', 'initial_state_x', 'initial_state_y', 'initial_state_z'] 
 Number of training events: 9 
 Number of validation events: 1 
 Number of test events: 0 
```

In this PR, the following functionality is added:

-  `graphnet.datasets` is added; it is intended to hold these pre-converted datasets.
- `CuratedDataset`; it is essentially a wrapper of `GraphNeTDataModule` with static inputs. It exposes two abstractmethods `prepare_data` (logic for downloading, unzipping, etc, dataset files) and `_prepare_args` (prepares Dataset arguments).  `CuratedDataset` also have a rather long series of `@property` set to help document what the dataset contains. These properties are used to auto-generate a simple terminal print with details on the dataset.
- `ERDAHostedDataset`  - a child class of `CuratedDataset`. It has a specific definition of `prepare_data` that works with ERDA share links, allowing us to distribute public datasets via ERDA.
- `TestDataset` - a child of  `ERDAHostedDataset` that is used to run unit tests on this new functionality.

### DIFF
--- a/src/graphnet/datasets/test_dataset.py
+++ b/src/graphnet/datasets/test_dataset.py
@@ -60,5 +60,5 @@ class TestDataset(ERDAHostedDataset):
             "features": features,
             "truth": truth,
         }
-        selection = [0, 1, 2, 3, 4, 6, 7, 8, 9]  # event 5 is empty
+        selection = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9]  # event 5 is empty
         return dataset_args, selection, None
--- a/src/graphnet/data/dataset/dataset.py
+++ b/src/graphnet/data/dataset/dataset.py
@@ -634,7 +634,7 @@ class Dataset(
         if len(features):
             node_features = features
         else:
-            node_features = np.array([]).reshape((0, len(self._features) - 1))
+            node_features = np.array([]).reshape((0, len(self._features)))
 
         assert isinstance(features, np.ndarray)
         # Construct graph data object
--- a/tests/datasets/test_erda_hosted_dataset.py
+++ b/tests/datasets/test_erda_hosted_dataset.py
@@ -6,6 +6,7 @@ from graphnet.models.detector.prometheus import ORCA150
 from graphnet.models.graphs import KNNGraph
 from graphnet.datasets import TestDataset
 from graphnet.training.utils import DataLoader
+from graphnet.constants import TEST_DATA_DIR
 
 
 def iterate_dataloader(dataloader: DataLoader) -> None:
@@ -28,7 +29,7 @@ def test_erda_hosted_dataset(backend: str) -> None:
     """
     graph_definition = KNNGraph(detector=ORCA150())
     data_module = TestDataset(
-        download_dir="",
+        download_dir=TEST_DATA_DIR,
         graph_definition=graph_definition,
         backend=backend,
         train_dataloader_kwargs={"batch_size": 3, "num_workers": 1},


src/graphnet/data/curated_datamodule.py
```diff
@@ -0,0 +1,282 @@
...
+    methods.
+    """
+
+    def __init__(
+        self,
+        graph_definition: GraphDefinition,
+        download_dir: str,
+        truth: Optional[List[str]] = None,
+        features: Optional[List[str]] = None,
+        backend: str = "parquet",
```
Aske-Rosted, line 32: Are we slowly changing over to the default backend being parquet? Or is it just that the example dataset here is parquet? I was under the impression that sqlite was currently the "default" and what most people used?

src/graphnet/datasets/test_dataset.py
```diff
@@ -0,0 +1,64 @@
...
+
+        dataset_args = {
+            "truth_table": self._truth_table,
+            "pulsemaps": self._pulsemaps,
+            "path": dataset_path,
+            "graph_definition": self._graph_definition,
+            "features": features,
+            "truth": truth,
+        }
+        selection = [0, 1, 2, 3, 4, 6, 7, 8, 9]  # event 5 is empty
```
Aske-Rosted, line 63: We should be able to handle empty events though? wouldn't it be better to keep event 5 in for the testing of this.

Aske-Rosted (commented):
Code looks good and much more simple than I would have guessed. Left a couple of questions.

src/graphnet/data/curated_datamodule.py
RasmusOrsoe: There is not really much thought behind this default value. It's not intended to be read as a suggestion on what we recommend people to use. The code becomes a bit simpler if we allow for this argument to default to one of the backends, and by chance I chose `parquet`.

The parquet backend takes up ~8x less space and has a constant query speed, making it the "recommended" backend for very large datasets, or for people with tight space quotas. Because it does not offer fast random access to the entire dataset, it's somewhat difficult to work with, and is thus a bit more advanced.

sqlite has the fastest query speeds for small-medium sized datasets, and has full random access, but takes up quite a bit of space, and is therefore not scalable into very large datasets. It is quite easy to work with.

src/graphnet/datasets/test_dataset.py
RasmusOrsoe: Selection updated!

RasmusOrsoe:
@Aske-Rosted thank you very much for giving this a read! I've updated the selection and commented on the default value of the `backend` argument. Please let me know if anything is unclear still.

Aske-Rosted:
> @Aske-Rosted thank you very much for giving this a read! I've updated the selection and commented on the default value of the `backend` argument. Please let me know if anything is unclear still.

All good

Aske-Rosted (approved):
👍

PR #693, opened 2024-04-10
Add all public Prometheus Detector classes

RasmusOrsoe:
This PR adds A `Detector` for each of the incoming simulation datasets from Prometheus. This includes:

- ARCA
- Icecube86
- TRIDENT
- Baikal-GVD
- IceCube-Upgrade
- IceCubeGen2
- IceCubeDeepCore
- ORCA
-  and two "demo" detectors

I plan to add the datasets in a separate PR in the future, when the full simulation is available.


src/graphnet/models/detector/prometheus.py
```diff
@@ -38,5 +38,302 @@ def _t(self, x: torch.tensor) -> torch.tensor:
...
+        return x / 100
+
+    def _sensor_pos_z(self, x: torch.tensor) -> torch.tensor:
+        return x / 1000
+
+    def _t(self, x: torch.tensor) -> torch.tensor:
+        return x / 1.05e04
+
+
+class IceCubeGen2(Detector):
```
Aske-Rosted, line 311: Am i understanding correctly that in this configuration we are "simply" using the features available in both the old and new detector modules? and excluding stuff like directionality of the specific PMT? Which would also require us to be able to handle heterogeneous graphs [https://pytorch-geometric.readthedocs.io/en/latest/tutorial/heterogeneous.html]

Aske-Rosted (approved):
One more general question, but the code itself looks fine.

src/graphnet/models/detector/prometheus.py
RasmusOrsoe: Hey! Simulation from Prometheus "stops" at the module-level, so it does not simulate DOMs etc. All we have are photons. We don't know where on the module, although I think Prometheus could perhaps add that information in the output if configured to do so. But the simulation we're expecting will not have that.

PR #686, opened 2024-04-03
Bugfix to #685 and #683

RasmusOrsoe:
Closes #683, closes #685 .

#683 was solved by switching a single line from polars `df.explode(columns)` which utilizes multiprocessing to a pure numpy-based solution. This numpy-based solution gave a significant speed up to `ParquetDataset.__getitem__`. In #677 I found the  `ParquetDataset.__getitem__` to be ~1.8 times slower than it's SQLite counterpart on a 1 million event sample. Following this PR,  `ParquetDataset.__getitem__` is ~1.2 times slower than its SQLite counterpart on the same sample.

Issue #683, opened 2024-04-03
`ParquetDataset` Hangs unexpectedly long when `__getitem__` is called


RasmusOrsoe:
**Describe the bug**
When working on #682 I noticed that `ParquetDataset` hangs unexpectedly long in the `__getitem__` function call, when used in a DataLoader - but not outside of it.

The line that hangs is [df.explode(columns)](https://github.com/graphnet-team/graphnet/blob/0c2486f1fbe2c6483393e04dd2d156e92d77a527/src/graphnet/data/dataset/parquet/parquet_dataset.py#L288) which relies on `polars` that may utilize multiprocessing to execute the line.

The bug appears to be resolved when setting the torch.multiprocessing context to "`spawn`".

PR #694, opened 2024-04-10
Update DataModule

RasmusOrsoe:
This PR revisits the bug documented in #683 which I thought was previously fixed in #686.

Because Polars uses optimization strategies when executing code, it will behave slightly differently under the hood depending on the data. In some cases it will use multithreading while in other cases it won't. #683 is caused by thread lock from polars multithreaded operations, and I thought in #686 that I had removed the line causing the issue, but when revisiting this on a different dataset, I found that the issue was now caused from other lines of polars code.

Issues with polars and "forking" multiprocessing context is mentioned [here](https://docs.pola.rs/user-guide/misc/multiprocessing/).

I have found what appears to be a definitive solution to the bug. By simply specifying

```python
dataloader = DataLoader(...,
                        multiprocessing_context = "spawn",
                        ..)

```
when creating dataloaders with the `ParquetDataset`, the threadlocking disappears.

This PR addresses this specific requirement to the `ParquetDataset` by checking for the argument and providing (hopefully) helpful logger messages (see screenshot).
image

I also took the liberty of having `test_dataloader_kwargs` and the `val_dataloader_kwargs` default to `train_dataloader_kwargs` in cases where there is a test/val dataloader but no arguments for their respective dataloaders.

### DIFF
--- a/src/graphnet/data/datamodule.py
+++ b/src/graphnet/data/datamodule.py
@@ -118,6 +118,10 @@ class GraphNeTDataModule(pl.LightningDataModule, Logger):
                 )
             else:
                 self._test_dataloader_kwargs = {}
+        else:
+            self._train_dataloader_kwargs = train_dl_args
+            self._validation_dataloader_kwargs = val_dl_args
+            self._test_dataloader_kwargs = test_dl_args or {}
 
     def _add_context(
         self, dataloader_args: Dict[str, Any], dataloader_type: str
--- a/src/graphnet/data/datamodule.py
+++ b/src/graphnet/data/datamodule.py
@@ -24,9 +24,12 @@ class GraphNeTDataModule(pl.LightningDataModule, Logger):
         dataset_args: Dict[str, Any],
         selection: Optional[Union[List[int], List[List[int]]]] = None,
         test_selection: Optional[Union[List[int], List[List[int]]]] = None,
-        train_dataloader_kwargs: Optional[Dict[str, Any]] = None,
-        validation_dataloader_kwargs: Optional[Dict[str, Any]] = None,
-        test_dataloader_kwargs: Optional[Dict[str, Any]] = None,
+        train_dataloader_kwargs: Dict[str, Any] = {
+            "batch_size": 2,
+            "num_workers": 1,
+        },
+        validation_dataloader_kwargs: Dict[str, Any] = None,
+        test_dataloader_kwargs: Dict[str, Any] = None,
         train_val_split: Optional[List[float]] = [0.9, 0.10],
         split_seed: int = 42,
     ) -> None:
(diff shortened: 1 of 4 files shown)


src/graphnet/data/datamodule.py
```diff
@@ -72,6 +78,78 @@ def __init__(
...
+        """Handle assignment of `multiprocessing_context` arg to loaders.
+
+        Datasets relying on threaded libraries often require the
+        multiprocessing context to be set to "spawn". This method will check
+        the arguments for this entry and throw an error if the field is already
+        assigned to a wrong value. If the value is not specified, it is added
+        automatically with a log entry.
+        """
+        arg = "multiprocessing_context"
+        if arg in dataloader_args:
```
ArturoLlorente, line 130: Maybe it is better to define the case where `num_workers=0`, then no multiprocessing is needed, and this arg will trigger the error:

```
    dataloader = DataLoader(
  File "~/python3.10/site-packages/torch/utils/data/dataloader.py", line 262, in __init__
    self.multiprocessing_context = multiprocessing_context
  File "~/python3.10/site-packages/torch/utils/data/dataloader.py", line 421, in __setattr__
    super().__setattr__(attr, val)
  File "~/python3.10/site-packages/torch/utils/data/dataloader.py", line 410, in multiprocessing_context
    raise ValueError('multiprocessing_context can only be used with '
ValueError: multiprocessing_context can only be used with multi-process loading (num_workers > 0), but got num_workers=0
```

src/graphnet/data/datamodule.py
```diff
@@ -72,6 +78,78 @@ def __init__(
...
+
+        See: https://docs.pola.rs/user-guide/misc/multiprocessing/
+        """
+        if val_dl_args is None:
+            self.info(
+                "No `val_dataloader_kwargs` given. This arg has "
+                "been set to `train_dataloader_kwargs` automatically."
+            )
+            val_dl_args = deepcopy(train_dl_args)
+        if (test_dl_args is None) & (self._test_selection is not None):
```
ArturoLlorente, line 100: Here `val_dl_args` and `test_dl_args` have a main difference with `train_dl_args`, which is the argument `suffle`.
 I understand that it has to be set to `False` for both of them and `True` for the trainins.
Is it not going to be a problem if no `args` are speified for the validation/test?

ArturoLlorente (commented):
I made some few comments. Looks good!

RasmusOrsoe:
@ArturoLlorente thank you very much for pointing out these two errors! I have added a few commits that addresses them, and appended a few checks in the unit test of the DataModule to verify that the shuffle argument is set as expected.

ArturoLlorente (approved):
nothing else to add. Looks good!

PR #696, opened 2024-04-17
First Documentation Update

RasmusOrsoe:
This PR is the first in a series of updates to the documentation for graphnet hosted [here]([graphnet-team.github.io/graphnet/](https://graphnet-team.github.io/graphnet/)). Closes #420 .

The main change in this PR is an **installation matrix** (see screenshot) that allows users to click their way to installation commands. To make it easier to review these changes, I've zipped the html files  [html.zip](https://github.com/graphnet-team/graphnet/files/15012148/html.zip). When you unzip the folder and open the `index.html`, a local version of the documentation page should open.

image

To accommodate the different cuda versions, the `torch_gpu.txt` file under `graphnet/requirement` has been split into two, each representing a supported cuda version. The versions chosen here are from what is supported by PyG.

Minor changes:

The structure of the documentation is changed slightly under `graphnet/docs`. Instead of relying on one-off python scripts to convert static `.md` files into the `.rst` files (which is what `sphinx` requires), the files are now written in plain `.rst`. These conversion scripts have been deleted. The structure of the documentation is now:
```rst
   installation/install.rst #manual
   api/graphnet.rst #auto-generated
   about/about.rst #manual
   contribute/contribute.rst #manual
```
which we can expand upon in future pull requests.

PR #697, opened 2024-04-19
update documentation workflow

RasmusOrsoe:
Merging #696 introduced [an error](https://github.com/graphnet-team/graphnet/actions/runs/8750400161/job/24013868071) to the documentation workflow, because I forgot to update the github workflow file. 

Error is caused by the removal of the three python scripts that convert `.md` to `.rst`. The github workflow definition in #696 still looks for these scripts, but because they're removed, the workflow fails.

This PR updates the workflow definition by removing references to these (now unnecessary) scripts, and adds a new workflow that checks that documentation compiles without errors, such that future pull requests with errors in documentation will fail PR checks.


Aske-Rosted (approved):
Not that sharp in the workflow and Docker parts of the code but looks good to me.

PR #698, opened 2024-04-24
Add developer notes to documentation

RasmusOrsoe:
This PR moves the contents of `GETTING_STARTED.md` and `CONTRIBUTING.md` into our documentation page, and additional text and corrections have been added.

The github readme has been updated such that the installation section points to the new installation matrix in the documentation.

To make it easier to review this PR, I've included the html files in 
html.zip . If you unzip and open `index.html` a local version of the documentation page should appear in your browser.

[html.zip](https://github.com/graphnet-team/graphnet/files/15151319/html.zip)

Closes #401

PR #700, opened 2024-04-29
Update torch and macos version

RasmusOrsoe:
Lately workflows on PR have failed for `macos` and now recently in #698 for all `build` workflows, despite no changes are made to the related code.

These errors are caused by two dependency changes:

`torch` recently moved from `2.2` -> `2.3` and our `setup.py` currently has `torch>2.1`. `PyG` is not compatible with `torch==2.3` yet, causing the build errors in #698.

`macos-latest` recently moved from `macos-13` -> `macos-14 (ARM)` causing the macos errors we've seen recently.

This PR fixes these issues by setting the `macos` workflow to run on `macos-13` and changing our torch versioning to `torch>=2.1,<2.3`

### DIFF
--- a/.github/actions/install/action.yml
+++ b/.github/actions/install/action.yml
@@ -31,11 +31,11 @@ runs:
       shell: bash
     - name: Install dependencies
       run: |
-        pip install --upgrade pip>=20
-        pip install wheel setuptools==59.5.0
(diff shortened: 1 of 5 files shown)


ArturoLlorente (approved):
look good!

PR #701, opened 2024-05-01
Add public datasets

RasmusOrsoe:
This PR adds three public datasets from the incoming prometheus simulation for P-ONE, Baikal-GVD and TRIDENT. Each dataset contains roughly 1 million track events. These datasets are the "small" version - larger versions will be added in the future.

The PR also adds 
- "Track" label.
- P-ONE detector class and geometry table


ArturoLlorente (approved):
nothing else to add!

Issue #527, opened 2023-05-26
Use unsupervised learning to learn representations in data and Monte Carlo


asogaard:

RasmusOrsoe:
Hey @AMHermansen I've now taken a deeper look at this, and I must admit I still struggle to follow your diagram. I _think_ that your diagram depicts the usual two-stage approach for using autoencoders where one first trains the encoder-decoder pair in a unsupervised setting, and then secondly finetune one of components of the encoder-decoder pair on a supervised learning task. Assuming this, I think the intended workflow is straight forward:

The first stage pre-trains the encoder-decoder pair according to some scheme defined in an `AutoEncoding` task, and the encoder and decoder is saved separately for later use.

```python
from graphnet.models import AutoEncoder
from graphnet.models.gnn import DynEdge
from graphnet.models.detector import IceCube86
from graphnet.models.graphs import KNNGraph
from graphnet.models.task import AutoEncoding


# Define autoencoder
model = AutoEncoder(graph_definition = KNNGraph(),
                    detector = IceCube86()
                    encoder = DynEdge(..),
                    decoder = DynEdge(..),
                    task = AutoEncoding()
                    )

# Train autoencoder pair (first stage)
model.fit(..)

# Save encoder and decoder seperately
model.save_config(path = '..') # Saves the config for encoder and decoder seperately
model.save_state_dict(path = '')# Saves the state_dict for encoder and decoder seperately

```

In a second session, one can load in a pre-trained component of the encoder-decoder pair and fine tune it on a physics task: 


```python
 # .. In a different session (2nd stage)

from graphnet.models.gnn import DynEdge
from graphnet.models import StandardModel
from graphnet.models.detector import IceCube86
from graphnet.models.graphs import KNNGraph
from graphnet.models.task import EnergyReconstruction

pretrained_model = DynEdge.from_config('encoder_config.yml')
pretrained_model.load_state_dict('encoder_state_dict')

model = StandardModel(graph_definition = KNNGraph(),
                    detector = IceCube86()
                    gnn = pretrained_model,
                    task = EnergyReconstruction())

model.fit(..)

results = model.predict_as_dataframe(...)

# Save finetuned model
model.save_config(path = '..') 
model.save_state_dict(path = '')
```

I think the technical work required to achieve this syntax is reasonable.  @AMHermansen Could you elaborate on how your current approach differs from this?

AMHermansen:
[...]

PR #707, opened 2024-05-03
Refactor standard model

RasmusOrsoe:
Refactors `StandardModel` such that the easy syntax (`.fit`, `.predict_as_dataframe`, etc) is stored in a seperate class `EasySyntax`, allowing different `Model` s to use it without the restrictions of `StandardModel`. Refactor is done according to the discussion in #527


Aske-Rosted (approved):
Looks good to me.

PR #715, opened 2024-05-14
add isolated nodes

Aske-Rosted:
Adds a simple graph definition which simply does not construct a edge definition. Useful when the connection between the nodes is either not needed or constructed later in the model during run time.

### DIFF
--- a/src/graphnet/models/components/embedding.py
+++ b/src/graphnet/models/components/embedding.py
@@ -84,6 +84,7 @@ class FourierEncoder(LightningModule):
         super().__init__()
 
         self.sin_emb = SinusoidalPosEmb(dim=seq_length, scaled=scaled)
+        self.aux_emb = nn.Embedding(2, seq_length // 2)
         self.sin_emb2 = SinusoidalPosEmb(dim=seq_length // 2, scaled=scaled)
 
         if n_features < 4:
@@ -92,7 +93,7 @@ class FourierEncoder(LightningModule):
                 f"{n_features} features."
             )
         elif n_features >= 6:
-            self.aux_emb = nn.Embedding(2, seq_length // 2)
+
             hidden_dim = 6 * seq_length
         else:
             hidden_dim = int((n_features + 0.5) * seq_length)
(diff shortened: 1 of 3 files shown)


src/graphnet/models/graphs/graphs.py
```diff
@@ -54,3 +54,41 @@ def __init__(
             perturbation_dict=perturbation_dict,
             seed=seed,
         )
+
+
+class IsolatedNodes(GraphDefinition):
+    """A Graph representation where each node is isolated."""
```
RasmusOrsoe, line 60: Perhaps it would be more informative if we changed the doc string to say 

```
"""A Data representation without edge assignment.

I.e the resulting representation is created without an EdgeDefinition."""
```

src/graphnet/models/graphs/graphs.py
```diff
@@ -54,3 +54,41 @@ def __init__(
             perturbation_dict=perturbation_dict,
             seed=seed,
         )
+
+
+class IsolatedNodes(GraphDefinition):
```
RasmusOrsoe, line 59: Maybe we could rename this from `IsolatedNodes`  to `EdgelessGraph`, or something else that indicates exactly how this representation is different from existing representations?

RasmusOrsoe (commented):
Hey @Aske-Rosted. I left a couple of comments related to naming.

src/graphnet/models/components/embedding.py
```diff
@@ -93,6 +92,7 @@ def __init__(
                 f"{n_features} features."
             )
         elif n_features >= 6:
+            self.aux_emb = nn.Embedding(2, seq_length // 2)
```
RasmusOrsoe, line 95: Looks like your recent commit added this line. Was that intentional?

RasmusOrsoe (commented):
Thanks for renaming the class! I noticed that changes to embedding.py snuck into your recent commit too. Was that intentional?

src/graphnet/models/components/embedding.py
Aske-Rosted: Ah no, sorry that was for another PR.

RasmusOrsoe (approved):
🚀

PR #716, opened 2024-05-14
RMSE von Mises-Fisher 3d loss

Aske-Rosted:
Adds a combination of the RMSE loss and the von Mises-Fisher 3d loss, with the Mises-Fisher weighted with a factor 0.05. Similar to what was used in the 2nd place solution in the kaggle competition.

### DIFF
--- a/src/graphnet/training/loss_functions.py
+++ b/src/graphnet/training/loss_functions.py
@@ -71,6 +71,8 @@ class MSELoss(LossFunction):
         """Implement loss calculation."""
         # Check(s)
         assert prediction.dim() == 2
+        if target.dim() != prediction.dim():
+            target = target.squeeze(1)
         assert prediction.size() == target.size()
 
         elements = torch.mean((prediction - target) ** 2, dim=-1)
@@ -453,6 +455,8 @@ class EnsembleLoss(LossFunction):
         loss_functions: List[LossFunction],
         loss_factors: List[float] = None,
         prediction_keys: Optional[List[List[int]]] = None,
+        *args: Any,
+        **kwargs: Any,
     ) -> None:
         """Chain multiple loss functions together.
 
@@ -482,6 +486,7 @@ class EnsembleLoss(LossFunction):
             self._prediction_keys: Optional[List[List[int]]] = prediction_keys
         else:
             self._prediction_keys = None
+        super().__init__(*args, **kwargs)
 
     def _forward(self, prediction: Tensor, target: Tensor) -> Tensor:
         """Calculate loss using multiple loss functions.
@@ -504,11 +509,11 @@ class EnsembleLoss(LossFunction):
         ):
             if k == 0:
                 elements = self._factors[k] * loss_function._forward(
-                    prediction=prediction[prediction_key], target=target
+                    prediction=prediction[:, prediction_key], target=target
                 )
             else:
                 elements += self._factors[k] * loss_function._forward(
-                    prediction=prediction[prediction_key], target=target
+                    prediction=prediction[:, prediction_key], target=target
                 )
         return elements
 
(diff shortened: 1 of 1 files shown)


src/graphnet/training/loss_functions.py
```diff
@@ -443,3 +443,33 @@ def _forward(self, prediction: Tensor, target: Tensor) -> Tensor:
         kappa = prediction[:, 3]
         p = kappa.unsqueeze(1) * prediction[:, [0, 1, 2]]
         return self._evaluate(p, target)
+
+
+class RMSEVonMisesFisher3DLoss(VonMisesFisherLoss):
+    """von Mises-Fisher loss function vectors in the 3D plane."""
+
+    def _forward(self, prediction: Tensor, target: Tensor) -> Tensor:
```
RasmusOrsoe, line 451: Looks like a lot of the code in the `_forward` pass is copied from ` VonMisesFisher3DLoss`, and we already have an implementation of RMSELoss, so I think we should try to refactor this loss function to avoid repeating ourselves.

How about we refactor this new loss function like so:

```
class EnsembleLoss(LossFunction):
    """Chain multiple loss functions together. """

    def __init__(self, loss_functions: List[LossFunction], loss_factors : List[float] = None) -> None:
        """Chain multiple loss functions together.

            Optionally apply a weight to each loss function contribution.

            E.g. Loss = RMSE*0.5 + LogCoshLoss*1.5

        Args:
            loss_functions: A list of loss functions to use.
                Each loss function contributes a term to the overall loss.
            loss_factors: An optional list of factors that will be mulitplied
            to each loss function contribution. Must be ordered according
            to `loss_functions`. If not given, the weights default to 1.
        """

        if loss_factors is None:
            # add weight of 1 - i.e no discrimination
            loss_factors = np.repeat(1,len(loss_functions)).tolist() 

        assert len(loss_functions) == len(loss_functions)
        self._factors = loss_factors
        self._loss_functions = loss_functions

        """Calculate loss using multiple loss functions.

        Args:
            prediction: Output of the model.
            target: Target tensor, extracted from graph object.

        Returns:
            Elementwise loss terms. Shape [N,]
        """
        for k, loss_function in enumerate(self._loss_functions):
            if k == 0:
                elements = self._factors[k]*loss_function._forward(prediction = prediction, 
                                              target = target)
            else:
                elements += self._factors[k]*loss_function._forward(prediction = prediction, 
                                              target = target)
        return elements
    
class RMSEVonMisesFisher3DLoss(EnsembleLoss):
    """ Combine the VonMisesFisher3DLoss with RMSELoss."""

    def __init__(self, vmfs_factor: float = 0.05) -> None:
        """ VonMisesFisher3DLoss with a RMSE penality term.

            The VonMisesFisher3DLoss will be weighted with `vmfs_factor`.

        Args:
            vmfs_factor: A factor applied to the VonMisesFisher3DLoss term.
            Defaults ot 0.05.
        """

        super().__init__(loss_functions= [RMSELoss(), 
                                          VonMisesFisher3DLoss()], 
                                          loss_factors=[1, vmfs_factor])
```

RasmusOrsoe (commented):
Hey @Aske-Rosted - thanks for adding this. I think it's a cool idea to utilize multiple loss functions in one go. I have some comments regarding repeating / copying existing code. It's important that we do not repeat ourselves - i.e have multiple copies of essentially the same code in multiple places, as it make maintaining and updating the code more difficult. I tried to put together a suggested refactor that will generalize to other use cases. Let me know what you think!

src/graphnet/training/loss_functions.py
Aske-Rosted: I very much agree.

Using the code you provided as a starting point I coded up a solution. Due to the varying number of predictions used by the different loss functions this also had to be addressed by the solution. There were also some other minor changes that had to be done

- Adaptively squeezing the dimension of the target for RMSE
- Adding a super init of the `LossFunction` class in the `EnsembleLoss` class

I am unsure whether or not it is necessary to allow for args and kwargs in the EnsembleLoss init. I followed how it was done in the `CrossEntropyLoss` class.

PR #717, opened 2024-05-14
weights utilities + fix

Aske-Rosted:
Adds 2 options for the weight fitter 

- a choice to employ an automatic log10 binning
- a choice to cap the max single event weight to be x * sum(all_weights) (1>x>0)

Adds an option for the `BjoernLow` class
- determine x_low as a percentile of the truth. x_low in this case has to be given as a value between 0 and 1. 

+ a fix so that the db_count_norm works as intended.


RasmusOrsoe (commented):
@Aske-Rosted  - I think the changes look good. Two of the doc string lines are too long though ( see code climate). Could you shorten those? The other codeclimate warnings are fine.

RasmusOrsoe (approved):

Issue #718, opened 2024-05-15
ISeeCube triggering an exception


Aske-Rosted:
**Describe the bug**

**To Reproduce**
Run a script that triggers the `get_all_grapnet_classes` class. for an example running 
`graphnet/examples/04_training/03_train_dynedge_from_config.py`
triggers the error for me.

**Expected behavior**
No error message and running as usual.

**Error message**
```
Exception has occurred: RuntimeError
operator torchvision::nms does not exist
```

**Additional context**
The error message happens during the `from torchscale.architecture.encoder import Encoder` call in the relatively recently added ISeeCube model. It happens when running `model.from_config()` via the `get_all_grapnet_classes` class. (graphnet/src/graphnet/utilities/config/parsing.py). I suspect that there might be a missing dependency or something of the sort but cannot say for sure and the error message is somewhat ambiguous. 

also just noted the typo in `get_all_grapnet_classes` ...

I have also checked that I do not have conflicting versions in the nvidia libraries 

> nvidia-cublas-cu11            11.11.3.6
> nvidia-cuda-cupti-cu11        11.8.87
> nvidia-cuda-nvrtc-cu11        11.8.89
> nvidia-cuda-runtime-cu11      11.8.89
> nvidia-cudnn-cu11             8.7.0.84
> nvidia-cufft-cu11             10.9.0.58
> nvidia-curand-cu11            10.3.0.86
> nvidia-cusolver-cu11          11.4.1.48
> nvidia-cusparse-cu11          11.7.5.86
> nvidia-ml-py                  12.550.52
> nvidia-nccl-cu11              2.19.3
> nvidia-nvtx-cu11              11.8.86

And the torch libraries

> torch                         2.2.0+cu118
> torch_cluster                 1.6.3+pt22cu118
> torch_geometric               2.5.3
> torch_scatter                 2.1.2+pt22cu118
> torch_sparse                  0.6.18+pt22cu118
> torchmetrics                  1.4.0
> torchscale                    0.2.0
> torchvision                   0.17.0+rocm5.7

ChenLi2049:
Hi, I'm able to reproduce the bug, and here's the full error report. I'm working on it.
```Bash
graphnet [MainProcess] WARNING  2024-05-15 09:26:04 - has_icecube_package - `icecube` not available. Some functionality may be missing.
[...]

PR #719, opened 2024-05-16
requirements update

Aske-Rosted:
this small update seeks to solve the issued mentioned in #718 by fixing the version of torchvision in the requirement files.


ChenLi2049 (approved):
I tested cu118 and cu121. They both work.

Issue #720, opened 2024-05-17
Automatic ensembling of databases


Aske-Rosted:
**Automatic ensembling of databases when loading from config**
For some use cases having several separate databases rather than one large db is preferable. However our current load from config only allows for a single database.

The feature should allow for the path entry in the config files to be either a path or a list of path. In the case where the path is a list of path the databases should be _merged_. Any potential selections should be applied to all of the databases.

PR #721, opened 2024-05-17
Automatic ensemble new

Aske-Rosted:
the following is an implementation of the feature mentioned in #720. 

I had to include a type ignore on line 68. which I personally was not too happy with, but I did not manage to satisfy the pre-commit hooks without it. 

The solution uses the pre-existing `EnsembleDataset` class to automatically combine databases from a list of databases.

### DIFF
--- a/src/graphnet/data/dataloader.py
+++ b/src/graphnet/data/dataloader.py
@@ -65,7 +65,9 @@ class DataLoader(torch.utils.data.DataLoader):
             )
 
             if isinstance(config.path, list):
-                datasets: Union[Dict[str, Dataset], Dict[str, EnsembleDataset]] = {}  # type: ignore
+                datasets: Union[
+                    Dict[str, Dataset], Dict[str, EnsembleDataset], Dict
+                ] = {}
                 dataset_col: Dict[str, list] = {}
                 for key in config.selection.keys():
                     dataset_col[key] = []
--- a/tests/utilities/test_dataset_config.py
+++ b/tests/utilities/test_dataset_config.py
@@ -269,22 +269,3 @@ def test_dataset_config_files(backend: str) -> None:
         )
         == 0
     )
-
-
-@pytest.mark.order(6)
-@pytest.mark.parametrize("backend", ["sqlite"])
-def test_multiple_dataset_config_dict_selection(backend: str) -> None:
-    """Test constructing Dataset with multiple data paths."""
-    # Arrange
-    config_path = CONFIG_PATHS[backend]
-
-    # Single dataset
-    config = DatasetConfig.load(config_path)
-    dataset = Dataset.from_config(config)
-    # Construct multiple datasets
-    config_ensemble = DatasetConfig.load(config_path)
-    config_ensemble.path = [config_ensemble.path, config_ensemble.path]
-
-    ensemble_dataset = Dataset.from_config(config)
-
-    assert len(dataset) * 2 == len(ensemble_dataset)
--- a/tests/utilities/test_dataset_config.py
+++ b/tests/utilities/test_dataset_config.py
@@ -269,3 +269,22 @@ def test_dataset_config_files(backend: str) -> None:
         )
         == 0
     )
+
+
+@pytest.mark.order(6)
+@pytest.mark.parametrize("backend", ["sqlite"])
+def test_multiple_dataset_config_dict_selection(backend: str) -> None:
+    """Test constructing Dataset with multiple data paths."""
+    # Arrange
+    config_path = CONFIG_PATHS[backend]
+
+    # Single dataset
+    config = DatasetConfig.load(config_path)
+    dataset = Dataset.from_config(config)
+    # Construct multiple datasets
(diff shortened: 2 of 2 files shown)


src/graphnet/data/dataloader.py
```diff
@@ -63,23 +63,51 @@ def from_dataset_config(
                 "`shuffle` is automatically inferred from the selection name, "
                 "and thus should not specified as an argument."
             )
-            datasets = Dataset.from_config(config)
+
+            if isinstance(config.path, list):
+                datasets: Union[Dict[str, Dataset], Dict[str, EnsembleDataset]] = {}  # type: ignore
```
RasmusOrsoe, line 68: I think the reason for your problem with mypy could be that you define `datasets` to contain both `Dataset` and `EnsembleDataset`, but in the typehints of `DataLoader` the `dataset` is only type hinted to be `dataset: Dataset`. See [here](https://github.com/graphnet-team/graphnet/blob/d0d987474f1f4e76e53384b46a9d1950d3672d56/src/graphnet/data/dataloader.py#L31). I think if you changed that type hint to `dataset: Union[Dataset, EnsembleDataset]`, the type hinting should be fine. Did you try this?

RasmusOrsoe (commented):
Hey @Aske-Rosted  - Thank you very much for adding this new feature. I added a hint on what might be causing the mypy challenges that you mentioned. 

Given that this is new functionality, could you be persuaded to add a small unit test for this [here](https://github.com/graphnet-team/graphnet/blob/main/tests/utilities/test_dataset_config.py)? I added an example of how such a test could look like below:

@```
pytest.mark.order(6)
@pytest.mark.parametrize("backend", ["sqlite"])
def test_dataset_config_dict_selection(backend: str) -> None:
    """Test constructing Dataset with multiple data paths."""
    # Arrange
    config_path = CONFIG_PATHS[backend]

    # Single dataset
    config = DatasetConfig.load(config_path)
    dataset = Dataset.from_config(config)
    # Construct multiple datasets
    config_ensemble = DatasetConfig.load(config_path)
    config_ensemble.path = [config_ensemble.path, config_ensemble.path]

    ensemble_dataset = Dataset.from_config(config)
    
    assert len(dataset)*2 == len(ensemble_dataset)
```

Aske-Rosted:
> Hey @Aske-Rosted - Thank you very much for adding this new feature. I added a hint on what might be causing the mypy challenges that you mentioned.
> 
> Given that this is new functionality, could you be persuaded to add a small unit test for this [here](https://github.com/graphnet-team/graphnet/blob/main/tests/utilities/test_dataset_config.py)? I added an example of how such a test could look like below:
> 
> ```
> @pytest.mark.order(6)
> @pytest.mark.parametrize("backend", ["sqlite"])
> def test_dataset_config_dict_selection(backend: str) -> None:
>     """Test constructing Dataset with multiple data paths."""
>     # Arrange
>     config_path = CONFIG_PATHS[backend]
> 
>     # Single dataset
>     config = DatasetConfig.load(config_path)
>     dataset = Dataset.from_config(config)
>     # Construct multiple datasets
>     config_ensemble = DatasetConfig.load(config_path)
>     config_ensemble.path = [config_ensemble.path, config_ensemble.path]
> 
>     ensemble_dataset = Dataset.from_config(config)
>     
>     assert len(dataset)*2 == len(ensemble_dataset)
> ```

Hey Rasmus looked into the suggestion for a bit but since this automatic ensembling is happing during the dataloader, and not in the `dataset.from_config(config)`  call the current test suggestion does not test the ensembling code. This does however illuminate a bit of an issue with the current implementation of the automatic ensembling. That is you can have a dataset config file that works fine as long as you only feed it to a dataloader, however if you try to manually load the dataset from the config using the dataset class it will fail due to the list of files rather than a single file...

RasmusOrsoe:
@Aske-Rosted thanks for pointing this out. It's an essential function to be able to load datasets from their respective configuration files, so I think we need to make sure that this new usage doesn't break the core intention of the files.

I had a quick look at the code again, and I think if you just moved (with slight modifications) the new logic to handle multiple paths into `Dataset.from_config` (see https://github.com/Aske-Rosted/graphnet/blob/7baa60d2fd729e84214a9f1c5a1d0882cfeac14a/src/graphnet/data/dataset/dataset.py#L107) the problem would be solved. What do you think?

Aske-Rosted:
> @Aske-Rosted thanks for pointing this out. It's an essential function to be able to load datasets from their respective configuration files, so I think we need to make sure that this new usage doesn't break the core intention of the files.
> 
> I had a quick look at the code again, and I think if you just moved (with slight modifications) the new logic to handle multiple paths into `Dataset.from_config` (see https://github.com/Aske-Rosted/graphnet/blob/7baa60d2fd729e84214a9f1c5a1d0882cfeac14a/src/graphnet/data/dataset/dataset.py#L107) the problem would be solved. What do you think?

I agree that is probably the more appropriate location to handle multiple file paths. I will have a look at it.

Aske-Rosted:
Was fixed with PR #775

PR #725, opened 2024-05-20
add missing `abs` to `Track` label

RasmusOrsoe:
In our current implementation of the `Track` label, there is a missing `torch.abs` - resulting in a faulty label.

This PR adds the `torch.abs`.


Aske-Rosted (approved):
👍

PR #726, opened 2024-05-20
Cast single path to list of paths in `DataConverter.merge_files`

### LINKED ISSUE #711: DataConverter.merge_files loops over letters in string when filepath is given as string
Normally, DataConverter.merge_files should be called with a list of strings. If I (accidentally) only supply one filepath as a string, it will not raise an exception, but instead will loop over the letters in the string and create empty files having the letter as a name. E.g., the code

```
input_dir = "data"
converter.merge_files(files=input_dir)
```

will create empty files "d", "a", and "t" in the current working directory. It would be better if an exception is raised.

RasmusOrsoe:
closes #711

### DIFF
--- a/src/graphnet/data/dataconverter.py
+++ b/src/graphnet/data/dataconverter.py
@@ -336,8 +336,7 @@ class DataConverter(ABC, Logger):
             # Proceed to merge specified by user.
             if isinstance(files, str):
                 # We shouldn't merge a single file?
-                self.info(f"Got just a single file {files}."
-                           "Merging skipped.")
+                self.info(f"Got just a single file {files}. Merging skipped.")
                 return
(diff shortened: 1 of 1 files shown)


Aske-Rosted (commented):
While the issue mentions that the user should be made aware of issue by raising a more sensible warning. This "fixes" the users mistake but ultimately runs a merge_files script on a single file. Is there some scenario where this action makes sense? Or is there something I am missing?

RasmusOrsoe:
@Aske-Rosted thanks for pointing this out. That is why we have reviews! I updated the code to prompt an info message and return without merging.

Aske-Rosted (approved):
Looks good.

Issue #722, opened 2024-05-17
FourierEncoder unused parameters. 


Aske-Rosted:
**Bug**
The bug happens if less than 6 features is used in the FourierEncoder class.  The script will _terminate_  with the following error

> 
RuntimeError: It looks like your LightningModule has parameters that were not used in producing the loss returned by training_step. If this is intentional, you must enable the detection of unused parameters in DDP, either by setting the string value `strategy='ddp_find_unused_parameters_true'` or by setting the flag in the strategy with `strategy=DDPStrategy(find_unused_parameters=True)`

This happens because self.aux_emb is initialized whether or not it is used.

PR #727, opened 2024-05-21
unused params fix

Aske-Rosted:
Fixes an issue in the `FourierEncoder` class where an `embedding` is intantiated but unused leading to the runtime error reported in issue #722


RasmusOrsoe:
@Aske-Rosted looks like black is failling and that there is a merge conflict

RasmusOrsoe (approved):
Sorry for the delay!

PR #728, opened 2024-05-29
Jammy flow integration

RasmusOrsoe:
This PR adds support for Normalizing Flows via the [jammy_flows](https://github.com/thoglu/jammy_flows) package, and therefore supersedes #649. The benefit of using `jammy_flows `is that it contains many different normalizing flows, and that we avoid maintaining that code ourselves :-).

The package is not listed as a direct dependency but used as an optional support package. 

Specifically, this PR introduces the following major changes:

1.  `StandardFlowTask` now uses jammy_flows to construct pdfs of any kind that it supports. These pdfs can be both conditional and non-conditional. Conditional flows can be conditioned on latent model output, event-level information or pulse-level information. 
2. A new model class is added: ` NormalizingFlow` which work with the `StandardFlowTask`. Usage is similar to `StandardModel`.
3. An example of training a conditional flow is added under `examples/04_training/07_train_normalizing_flow.py`
4. `has_jammy_flows_package()` is added under `graphnet.utils.imports` to check if its installed, and is used in a few places to make sure that the code runs also for people who choose not to install jammy_flows.

Minor changes:
1. `repeat_labels` is added as an argument to `GraphDefinition` - if `True`, event-level information, .e.g `energy` is repeated row-wise to match the number of pulses in the event. This feature was added in this PR to make it possible to build flows that learn pulse-level pdfs conditioned on event-level information.
2.  Installation matrix is updated to provide a note on the installation of jammy flows
3. Github workflows is adjusted to run **with** jammy_flows installed.
4. `**kwargs` for `Trainer` is added for `predict`-methods in `EasySyntax` to allow the same level of control over `Trainer` arguments as we have for `.fit`

### DIFF
--- a/src/graphnet/models/easy_model.py
+++ b/src/graphnet/models/easy_model.py
@@ -292,6 +292,7 @@ class EasySyntax(Model):
         dataloader: DataLoader,
         gpus: Optional[Union[List[int], int]] = None,
         distribution_strategy: Optional[str] = "auto",
+        **trainer_kwargs,
     ) -> List[Tensor]:
         """Return predictions for `dataloader`."""
         self.inference()
@@ -305,6 +306,7 @@ class EasySyntax(Model):
             gpus=gpus,
             distribution_strategy=distribution_strategy,
             callbacks=callbacks,
+            **trainer_kwargs,
         )
 
         predictions_list = inference_trainer.predict(self, dataloader)
@@ -325,6 +327,7 @@ class EasySyntax(Model):
         additional_attributes: Optional[List[str]] = None,
         gpus: Optional[Union[List[int], int]] = None,
         distribution_strategy: Optional[str] = "auto",
+        **trainer_kwargs,
     ) -> pd.DataFrame:
         """Return predictions for `dataloader` as a DataFrame.
 
@@ -357,6 +360,7 @@ class EasySyntax(Model):
             dataloader=dataloader,
             gpus=gpus,
             distribution_strategy=distribution_strategy,
+            **trainer_kwargs,
         )
         predictions = (
             torch.cat(predictions_torch, dim=1).detach().cpu().numpy()
--- a/src/graphnet/models/easy_model.py
+++ b/src/graphnet/models/easy_model.py
@@ -16,7 +16,6 @@ import pandas as pd
 from pytorch_lightning.loggers import Logger as LightningLogger
 
 from graphnet.training.callbacks import ProgressBar
-from graphnet.models.graphs import GraphDefinition
 from graphnet.models.model import Model
 from graphnet.models.task import StandardLearnedTask
 
@@ -292,7 +291,7 @@ class EasySyntax(Model):
         dataloader: DataLoader,
         gpus: Optional[Union[List[int], int]] = None,
         distribution_strategy: Optional[str] = "auto",
-        **trainer_kwargs,
+        **trainer_kwargs: Any,
     ) -> List[Tensor]:
         """Return predictions for `dataloader`."""
         self.inference()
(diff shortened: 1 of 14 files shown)


RasmusOrsoe:
@Aske-Rosted checks are now passing. Please let me know if you have any questions

src/graphnet/models/graphs/graph_definition.py
```diff
@@ -62,9 +63,14 @@ def __init__(
             add_inactive_sensors: If True, inactive sensors will be appended
                 to the graph with padded pulse information. Defaults to False.
             sensor_mask: A list of sensor id's to be masked from the graph. Any
-                sensor listed here will be removed from the graph. Defaults to None.
-            string_mask: A list of string id's to be masked from the graph. Defaults to None.
+                sensor listed here will be removed from the graph.
+                    Defaults to None.
```
Aske-Rosted, line 67: Did some unintended formatting happen here?

src/graphnet/models/task/task.py
```diff
@@ -380,66 +386,85 @@ def _forward(self, x: Union[Tensor, Data]) -> Tensor:  # type: ignore
...
-    def _forward(self, x: Tensor, jacobian: Tensor) -> Tensor:  # type: ignore
-        # Leave it as is.
-        return x
+    @property
+    def default_prediction_labels(self) -> List[str]:
+        """Return default prediction labels."""
+        return self._default_prediction_labels
+
+    def nb_inputs(self) -> Union[int, None]:  # type: ignore
+        """Return number of conditional inputs assumed by task."""
```
Aske-Rosted, line 434: does non-conditional inputs also exist or what is the reason for the distinction of "conditional"?

src/graphnet/models/easy_model.py
Aske-Rosted, line 1: Are these changes unrelated to jammy flows?

src/graphnet/models/graphs/graphs.py
Aske-Rosted, line 1: Are these changes also unrelated to Jammy flows?

Aske-Rosted (approved):
LGTM - A few questions here and there that do not necessarily require fixing.

src/graphnet/models/easy_model.py
Aske-Rosted: reread the minor changes and these changes are described there you can disregard this question.

src/graphnet/models/graphs/graphs.py
Aske-Rosted: reread the minor changes and these changes are described there you can disregard this question.

src/graphnet/models/task/task.py
```diff
@@ -380,66 +386,85 @@ def _forward(self, x: Union[Tensor, Data]) -> Tensor:  # type: ignore
...
-        self._regularisation_loss = 0  # Reset
-        x = self._forward(x, jacobian)
+        # Manually cast pdf to correct dtype - is there a better way?
+        self._flow = self._flow.to(self.dtype)
+        # Get target values
+        labels = get_fields(data=data, fields=self._target_labels)
+        labels = labels.to(self.dtype)
+        # Set the initial parameters of flow close to truth
+        # This speeds up training and helps with NaN
+        if self._initialized is False:
```
jvaracarbonell, line 464: Just a quick note: I noticed that this part could override the learned weights whenever they are loaded from a state dict, which might affect further training or inference. It could be worth adding a global boolean to check if the weights have been loaded.  I believe this could be done[ here](https://github.com/graphnet-team/graphnet/blob/dc7fa4f4515b28b1ea37838402f0c41ee0619ce3/src/graphnet/deployment/deployment_module.py#L65). Otherwise, it might be good at least to pass `and self.training` , which should work when the model is set to evaluation mode for predictions.

src/graphnet/models/task/task.py
RasmusOrsoe: The flow can be trained both as a conditional flow, where it's conditioned on a latent representation of an event e.g. `p(target|latent_rep) `  which is the probability of seeing `target` for event represented by` latent_rep`, and as an unconditional probability distribution e.g. `p(target)` which would just learn the overall distribution of the variable in the dataset.

src/graphnet/models/task/task.py
RasmusOrsoe: @jvaracarbonell thanks for pointing this out! I've added the check

PR #739, opened 2024-08-15
ParticleNet

IvanMM27:
ParticleNeT model for Jet Tagging with particle cloud like data used in KM3NeT implemented succesfully in GraphNeT. It uses all utilities from GraphNeT, mainly DynEdgeConv blocks to build its architecture. References:
- https://arxiv.org/pdf/1902.08570
- https://ecap.nat.fau.de/wp-content/uploads/2023/06/2023-06_LukasHennig_MSc_ORCA_TauID_GNNs.pdf

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -109,6 +109,7 @@ jobs:
           pip show torch-cluster
           pip show torch-sparse
           pip show torch-scatter
+          pip show numpy
       - name: Run unit tests and generate coverage report
         run: |
           set -o pipefail  # To propagate exit code from pytest
--- a/examples/01_icetray/01_convert_i3_files.py
+++ b/examples/01_icetray/01_convert_i3_files.py
@@ -1,6 +1,5 @@
 """Example of converting I3-files to SQLite and Parquet."""
 
-import os
 from glob import glob
 
 from graphnet.constants import EXAMPLE_OUTPUT_DIR, TEST_DATA_DIR
--- a/examples/04_training/01_train_dynedge.py
+++ b/examples/04_training/01_train_dynedge.py
@@ -16,9 +16,11 @@ from graphnet.models.graphs import KNNGraph
 from graphnet.models.task.reconstruction import EnergyReconstruction
 from graphnet.training.callbacks import PiecewiseLinearLR
 from graphnet.training.loss_functions import LogCoshLoss
-from graphnet.training.utils import make_train_validation_dataloader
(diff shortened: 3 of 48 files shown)


RasmusOrsoe (commented):
@IvanMM27 thank you for adding the ParticleNeT model!

In order to merge this pull request, we need the automatic checks to pass. As you might notice, quite a few of the checks are currently failing. In addition, this PR appears to contain not only changes related to the ParticleNeT model itself, but also code that looks like KM3NeT integration code.

I would suggest that this PR is split up into at least two smaller PR's to reduce the complexity. I.e. One pull request that contains the ParticleNeT model, and then a second pull request that contains the KM3NeT integration.

@IvanMM27 would that be alright with you?

Tagging @giogiopg for completeness

IvanMM27:
> @IvanMM27 thank you for adding the ParticleNeT model!
> 
> In order to merge this pull request, we need the automatic checks to pass. As you might notice, quite a few of the checks are currently failing. In addition, this PR appears to contain not only changes related to the ParticleNeT model itself, but also code that looks like KM3NeT integration code.
> 
> I would suggest that this PR is split up into at least two smaller PR's to reduce the complexity. I.e. One pull request that contains the ParticleNeT model, and then a second pull request that contains the KM3NeT integration.
> 
> @IvanMM27 would that be alright with you?
> 
> Tagging @giogiopg for completeness

@RasmusOrsoe I think it is better to re-do this pull request again as we want to push only the ParticleNeT model and not any of KM3NeT utilities. Would it be okay if I delete this one and we do another one? Apologies for the misunderstanding. 

@giogiopg

RasmusOrsoe:
> > @IvanMM27 thank you for adding the ParticleNeT model!
> > In order to merge this pull request, we need the automatic checks to pass. As you might notice, quite a few of the checks are currently failing. In addition, this PR appears to contain not only changes related to the ParticleNeT model itself, but also code that looks like KM3NeT integration code.
> > I would suggest that this PR is split up into at least two smaller PR's to reduce the complexity. I.e. One pull request that contains the ParticleNeT model, and then a second pull request that contains the KM3NeT integration.
> > @IvanMM27 would that be alright with you?
> > Tagging @giogiopg for completeness
> 
> @RasmusOrsoe I think it is better to re-do this pull request again as we want to push only the ParticleNeT model and not any of KM3NeT utilities. Would it be okay if I delete this one and we do another one? Apologies for the misunderstanding.
> 
> @giogiopg

Yes! That is totally fine!

PR #748, opened 2024-09-13
Add pulse merging functionality to `GraphDefinition`

RasmusOrsoe:


pweigel:
Just a minor comment/warning in `_find_photons_for_merging`: the time windows are assumed to begin at the start of the first pulse/photon, which can lead to some bias. In data, the start time can be between 0 and `merge_window` ns before the first pulse because the DAQ is always sampling (well, for our purposes). A second pulse that comes soon after may not be merged in data, but if the window start time is set to the time of the first pulse it may always be merged.

src/graphnet/models/graphs/graph_definition.py
```diff
@@ -464,3 +497,153 @@ def _add_custom_labels(
...
+                        f" {self._input_feature_names}."
+                        " Cannot do weighted sum."
+                    )
+                    weights = np.array([1])
+                # Handle numeric and non-numeric fields differently
+                if all(
+                    isinstance(value, (int, float))
+                    for value in values_to_merge
+                ):
+                    # alculate the mean for all attributes except charge
```
Aske-Rosted, line 585: small typo

Aske-Rosted (approved):
small typo but otherwise LGTM.

Issue #746, opened 2024-09-12
Wrong location for SQLiteDataset in the documentation


niklasmei:
**Describe the bug**
In the documentation according to https://graphnet-team.github.io/graphnet/datasets/datasets.html the path to import `SQLiteDataset` is not up to date. Therefore trying the import as suggested on the site mentioned above does not work.

**To Reproduce**
simply execute the following:

`from graphnet.data.sqlite import SQLiteDataset`

in the current version of graphnet.

**Expected behavior**
From the documentation I would expect the import to work without error. It works as expected with the correct import: 

`from graphnet.data.dataset.sqlite.sqlite_dataset import SQLiteDataset`

**Full traceback**
```
ImportError                               Traceback (most recent call last)
Cell In[9], [line 1](vscode-notebook-cell:?execution_count=9&line=1)
----> [1](vscode-notebook-cell:?execution_count=9&line=1) from graphnet.data.sqlite import SQLiteDataset

ImportError: cannot import name 'SQLiteDataset' from 'graphnet.data.sqlite' (/mnt/c/Users/nikla/Desktop/GraphNet/anaconda_graphnet_file/graphnet/src/graphnet/data/sqlite/__init__.py)
```

**Additional context**
This is not a problem with graphnet, but just an outdated suggestion in the documentation. You find it specifically in the _SQLiteDataset & ParquetDataset_ subection of the _Datasets in GraphNet_ section of the documentation (https://graphnet-team.github.io/graphnet/datasets/datasets.html)

RasmusOrsoe:
@niklasmei thanks for reporting this!

The file in question can be edited here: `docs/source/datasets/datasets.rst`

niklasmei:
Pull request to fix this issue should also resolve issue #750

PR #749, opened 2024-09-14
Changed the import path of SQLiteDataset in the documentation to the currently correct location.

niklasmei:
Import path for `SQLiteDataset` was wrong in the documentation. This is fixed here and at the same time resolves a similar issue for the `ParquetDataset`. An official issue for the parquet case will follow.


niklasmei:
this will close issues #746 and #750

RasmusOrsoe (approved):
@niklasmei thanks for this! 🚀

PR #760, opened 2024-10-28
Pre commit version

RasmusOrsoe:
Since the recent update of pre-commit to 4.0, `docformatter ` has failed in workflows across the globe, including in ours.

This PR sets `pre-commit <= 3.8` in setup.py and updates our hooks (flake8, mypy, etc) to recent versions.

I took this as an opportunity to do some spring cleaning, and to get flake8 to catch unused imports. The size of this PR might look overwhelming, but there is **no functional change** introduced here other than what's mentioned above.  The changes seen here are superficial changes to make the codebase pass the updated hooks (In total, ~108 code climate flags has been solved).

Closes #757. Tagging @AlexKurek for completeness, as you've worked on the unused imports in the past.

PR #763, opened 2024-11-01
Improvements for inelasticity reconstructions and other utilities

pweigel:
This pull request implements some changes for training inelasticity reconstructions and adds other utilities:

- The `I3TruthExtractor` can now determine starting events using the event vertex and the convex hull of the detector volume (using the GCD file). The convex hull can also be expanded with `extend_boundary`. Some other utilities in this class use the "boundaries" list, but I think these could be redone to use the convex hull. I should note I had to override the `set_gcd` function since the geometry is added after instantiation.
- Additionally, the `I3TruthExtractor` can now identify MC produced with LeptonInjector and get the relevant quantities.
- Added a new I3Filter that selects on `sub_event_stream`.
- New setting for `IceMixNodes` to enable/disable the random pulse sampling (default True).
- Added `VisibleInelasticityReconstruction` task
- Added `MAELoss`


RasmusOrsoe:
@pweigel thank you for this contribution! The problem with pre-commit is now fixed in ` main.` If you update the branch, the failing checks in this PR are likely to pass (see #760).

src/graphnet/data/extractors/icecube/i3truthextractor.py
```diff
@@ -13,7 +13,7 @@
 from graphnet.utilities.imports import has_icecube_package
 
 if has_icecube_package() or TYPE_CHECKING:
-    from icecube import (
+    from icecube import (  # noqa: F401
```
pweigel, line 16: I needed to add this because LeptonInjector isn't used directly, but it needs to be imported for icetray to recognize the "EventProperties" frame object type.

src/graphnet/data/extractors/icecube/i3truthextractor.py
```diff
@@ -78,15 +84,85 @@ def __init__(
             self._borders = [border_xy, border_z]
         else:
             self._borders = borders
+
+        self._extend_boundary = extend_boundary
         self._mctree = mctree
 
+    def set_gcd(self, i3_file: str, gcd_file: Optional[str] = None) -> None:
```
RasmusOrsoe, line 91: Limiting the adjustments of`set_gcd` to only `I3TruthExtractor` is a good idea, as the modifications are only needed here. But instead of copying the original function, let's solve this through inheritance. Here is some pseudo code that I think will be identical in function but omit the duplication:

```python
def set_gcd(self, i3_file: str, gcd_file: Optional[str] = None) -> None:
    """Extract GFrame and CFrame from i3/gcd-file pair.
           Information from these frames will be set as member variables of
           `I3Extractor.`
        Args:
            i3_file: Path to i3 file that is being converted.
            gcd_file: Path to GCD file. Defaults to None. If no GCD file is
                      given, the method will attempt to find C and G frames in
                      the i3 file instead. If either one of those are not
                      present, `RuntimeErrors` will be raised.
    """
    # Parent implementation of `set_gcd`
    super().set_gcd(i3_file = i3_file,
                    gcd_file = gcd_file)
    
    # Modifications specific to I3TruthExtractor
    # These modifications are needed to identify starting events
    coordinates = []
    for _, g in self._gcd_dict.items():
        if g.position.z > 1200:
            continue  # We want to exclude icetop
        coordinates.append([g.position.x, g.position.y, g.position.z])
    coordinates = np.array(coordinates)

    if self._extend_boundary != 0.0:
        center = np.mean(coordinates, axis=0)
        d = coordinates - center
        norms = np.linalg.norm(d, axis=1, keepdims=True)
        dn = d / norms
        coordinates = coordinates + dn * self._extend_boundary

    hull = ConvexHull(coordinates)

    self.hull = hull
    self.delaunay = Delaunay(coordinates[self.hull.vertices])
```

RasmusOrsoe (changes requested):
@pweigel thank you for these contributions!

I added a small comment related to the duplication of the `set_gcd`.

src/graphnet/data/extractors/icecube/i3truthextractor.py
pweigel: Good point! Made these changes in the last commit.

RasmusOrsoe (approved):
🚀

PR #766, opened 2024-11-07
Autofix: I3InferenceModule doesn't recognize a list of length 1

lattaai3:
This change modifies the I3InferenceModule to avoid spamming warnings when there's only one task. It now checks the length of the predictions list and uses warning_once to prevent repeated warnings. 
> [!CAUTION]
> Disclaimer: This fix was created by Latta AI and you should never merge before you check the correctness of generated code!
>
> The fix provided by Latta AI might not be complete and it can serve as an inspiration.

---
This bug was fixed for free by Latta AI - https://latta.ai/ourmission

If you no longer want Latta AI to attempt fixing issues on your repository, you can block this account.


pweigel:
Bad bot

Issue #765, opened 2024-11-07
Importing I3Tray directly is deprecated


pweigel:
**Describe the bug**
Using `I3Tray` directly is now deprecated, so it should be imported using `icecube.icetray` for the icecube-related features.

**To Reproduce**
1. `from I3Tray import I3Tray`

**Full traceback**
```
.../graphnet/src/graphnet/deployment/icecube/i3deployer.py:14: UserWarning: Using `import I3Tray` or `from I3Tray import *` is now considered deprecated. Please switch to using `from icecube.icetray import I3Tray`
  from I3Tray import I3Tray
```

pweigel:
Resolved with https://github.com/graphnet-team/graphnet/pull/772.

PR #772, opened 2024-11-22
Fix deprecated I3Tray imports

pweigel:
Fix for https://github.com/graphnet-team/graphnet/issues/765. Just changing `from I3Tray import I3Tray` to `from icecube.icetray import I3Tray` for a few files.


RasmusOrsoe (approved):
Thank you for this @pweigel ! 🚀

PR #777, opened 2024-12-17
Add GRIT model

pweigel:
### GRIT: "Graph Inductive Biases in Transformers without Message Passing"
This PR includes a new model based on the GRIT transformer. It uses novel methods for encoding graph information for use in sparse multi-head attention blocks. It uses a learned position encoding based on random walk probabilities, which enhances the model's expressivity.

PMLR: https://proceedings.mlr.press/v202/ma23c.html
Paper pre-print: https://arxiv.org/abs/2305.17589

image

Many layers/functions are adapted from the original repository: https://github.com/LiamMa/GRIT/tree/main. The original code uses `graphgym` to set up most of its modules, so I refactored some things to fit into graphnet. Many of the arguments have been relabeled to be more self-explanatory. In principle, other graph attention mechanisms could be used by replacing the GRIT MHA block. 

Since there are a lot of changes, I will quickly summarize the significant new additions and modifications to existing files:

- [src/graphnet/models/components/embedding.py](https://github.com/graphnet-team/graphnet/compare/main...pweigel:graphnet:grit#diff-f0f33604024b7cf450a886a8e94df9afe84760555fdafcc399377994994c92ee): Adds four encoding modules, which apply linear and RRWP encodings to nodes and edges.
- [src/graphnet/models/components/layers.py](https://github.com/graphnet-team/graphnet/compare/main...pweigel:graphnet:grit#diff-4eaa1b7d1c266a2439af0ff264d00ef274a5c7e97d9aa6388c9154e32d919cf1): Adds GRIT sparse multi-head attention, the core of the GRIT model, and the GRIT transformer layer.
- [src/graphnet/models/gnn/grit.py](https://github.com/graphnet-team/graphnet/compare/main...pweigel:graphnet:grit#diff-c3a2c42bf35e7e1f77b16930222ee8032099f2fa5b713f5192aca4066a096de6): Implementation of GRIT as a GNN model.
- [src/graphnet/models/graphs/edges/edges.py](https://github.com/graphnet-team/graphnet/compare/main...pweigel:graphnet:grit#diff-85c6e5df5fe062e8c382f4d5db4ac86526fc97f2b3b5fee85e3c8813dd26924d): Added a variant of the `KNNEdge` (`KNNDistanceEdges`) to include edge values corresponding to the distance between the node pair.
- [src/graphnet/models/graphs/graphs.py](https://github.com/graphnet-team/graphnet/compare/main...pweigel:graphnet:grit#diff-78ce7f838fa1e2ec5a57b5bd1820615dcb869bc3d3cedcc04121e79efdec2b4c): Added a variant of the `KNNGraph` to use the `KNNDistanceEdges` and compute the RRWP values.
- [src/graphnet/models/utils.py](https://github.com/graphnet-team/graphnet/compare/main...pweigel:graphnet:grit#diff-a14c152e62a2cefdb5448d6f2bdc30701e97319ceeb99416d7336fd55d9efe37): Added new utilities needed for the RRWP calculation and other miscellaneous tools used by the GRIT transformer layers.
- [examples/04_training/08_train_grit_model.py](https://github.com/graphnet-team/graphnet/compare/main...pweigel:graphnet:grit#diff-088fd8c5e18678c8be006b270440cd12a24292e2c7133e74cad1d6af2df5bf2e): Example training script using GRIT.

This model has many hyperparameters, but the defaults should provide a good starting point. It should be noted that the GPU memory required to train this model is quite high due to the use of global attention.


pweigel:
After some experimenting, I've found that no position encoding works quite well (skipping the RRWP encodings) and drastically reduces the GPU memory requirement. I'll add some options that allow users to do this.

src/graphnet/models/utils.py
```diff
@@ -116,3 +120,217 @@ def get_fields(data: Union[Data, List[Data]], fields: List[str]) -> Tensor:
...
+            out = out @ adj
+            pe_list.append(out)
+
+    pe = torch.stack(pe_list, dim=-1)  # n x n x k
+
+    abs_pe = pe.diagonal().transpose(0, 1)  # n x k
+
+    rel_pe = SparseTensor.from_dense(pe, has_value=True)
+    rel_pe_row, rel_pe_col, rel_pe_val = rel_pe.coo()
+    # rel_pe_idx = torch.stack([rel_pe_row, rel_pe_col], dim=0)
```
RasmusOrsoe, line 226: forgot a comment here

src/graphnet/models/utils.py
```diff
@@ -116,3 +120,217 @@ def get_fields(data: Union[Data, List[Data]], fields: List[str]) -> Tensor:
...
+    edge_index_full = torch.cat(negative_index_list, dim=1).contiguous()
+    return edge_index_full
+
+
+@torch.no_grad()
+def add_full_rrwp(
+    data: Data,
+    walk_length: int = 8,
+    attr_name_abs: str = "rrwp",  # name: 'rrwp'
+    attr_name_rel: str = "rrwp",  # name: ('rrwp_idx', 'rrwp_val')
```
RasmusOrsoe, line 172: forgot a comment here

src/graphnet/models/utils.py
```diff
@@ -116,3 +120,217 @@ def get_fields(data: Union[Data, List[Data]], fields: List[str]) -> Tensor:
...
+
+    Returns:
+        2D Tensor with shape (num_nodes, len(ksteps)) with RW landing probs
+    """
+    print(edge_index.shape)
+    if edge_weight is None:
+        edge_weight = torch.ones(edge_index.size(1), device=edge_index.device)
+    num_nodes = maybe_num_nodes(edge_index, num_nodes)
+    source = edge_index[0]
+    # dest = edge_index[1]
```
RasmusOrsoe, line 302: forgot a comment here

src/graphnet/models/utils.py
```diff
@@ -116,3 +120,217 @@ def get_fields(data: Union[Data, List[Data]], fields: List[str]) -> Tensor:
...
+
+    negative_index_list = []
+    for i in range(batch_size):
+        n = num_nodes[i].item()
+        size = [n, n]
+        adj = torch.ones(size, dtype=torch.short, device=edge_index.device)
+
+        adj = adj.view(size)
+        _edge_index = adj.nonzero(as_tuple=False).t().contiguous()
+        # _edge_index, _ = remove_self_loops(_edge_index)
```
RasmusOrsoe, line 160: forgot a comment here

src/graphnet/models/utils.py
```diff
@@ -116,3 +120,217 @@ def get_fields(data: Union[Data, List[Data]], fields: List[str]) -> Tensor:
...
+
+    Args:
+        data: Input data.
+        walk_length: Number of random walks for encoding.
+        attr_name_abs: Absolute position encoding name.
+        attr_name_rel: Relative position encoding name.
+        add_identity: Add identity matrix to position encoding.
+        spd: Use shortest path distances.
+    """
+    # device = data.edge_index.device
```
RasmusOrsoe, line 189: forgot a comment here

src/graphnet/models/components/layers.py
```diff
@@ -593,3 +602,393 @@ def forward(
...
+        clamp: float = 5.0,
+        dropout: float = 0.0,
+        activation: nn.Module = nn.ReLU,
+        edge_enhance: bool = True,
+    ):
+        """Construct 'GritSparseMHA'.
+
+        Args:
+            in_dim: Dimension of the input tensor.
+            out_dim: Dimension of theo output tensor.
```
RasmusOrsoe, line 629: theo -> the

src/graphnet/models/components/layers.py
```diff
@@ -593,3 +602,393 @@ def forward(
...
+            score = score + E_b
+
+        score = self.activation(score)
+        e_t = score  # ehat_ij
+
+        # Output edge
+        if E is not None:
+            wE = score.flatten(1)
+
+        # Complete attention ccaclculation
```
RasmusOrsoe, line 705: ccaclculation -> calculation

src/graphnet/models/components/layers.py
```diff
@@ -593,3 +602,393 @@ def forward(
...
+        attn_edge_enhance: bool = True,
+    ):
+        """Construct 'GritTransformerLayer'.
+
+        Args:
+            in_dim: Dimension of the input tensor.
+            out_dim: Dimension of theo output tensor.
+            num_heads: Number of attention heads.
+            dropout: Dropout layer probability.
+            norm: Normalization layer.
```
RasmusOrsoe, line 770: Would be informative to mention here that the normalization layer is assumed to be un-instantiated. E.g: 

`norm: Uninstantiated normalization layer. Must be either BatchNorm1d or BatchNorm1d`

src/graphnet/models/graphs/graphs.py
```diff
@@ -100,3 +108,203 @@ def __init__(
             seed=seed,
             **kwargs,
         )
+
+
+class KNNGraphRRWP(GraphDefinition):
+    """KNN Graph with relative random walk probabilities."""
```
RasmusOrsoe, line 114: Similarly to the other representation, I would suggest expanding the doc string so it's easier to see how this differs from existing representations and how the new fields are accessible. I think there are five new fields in the RRWP case. So here's an example:

"""
A KNN graph representation with Relative Random Walk Probabilities (RRWP).

Identical to KNNGraph but with five additional fields:  

`abs_pe = graph['abs_pe']` # Absolute positional encoding
`rrwp_index = graph['rrwp_index']` # rrwp index (which is used for slicing the vals?)
`rrwp_val = graph['rrwp_val']` # rrwp values
`degree= graph['deg']` # Degree of each node (num. of incoming edges)
`log_degree = graph['log_deg']` # Equal to torch.log10(graph['deg'] + 1)
"""

src/graphnet/models/graphs/graphs.py
```diff
@@ -100,3 +108,203 @@ def __init__(
...
+        """Forward pass."""
+        graph = super().forward(input_features, input_feature_names, **kwargs)
+        ksteps = torch.arange(1, self.walk_length)
+        graph.rwse = get_rw_landing_probs(
+            ksteps=ksteps, edge_index=graph.edge_index, edge_weight=None
+        )
+        return graph
+
+
+class KNNGraphNoPE(GraphDefinition):
```
RasmusOrsoe, line 251: Considering the very high level of similarity between this and `KNNGraph`, perhaps it would be beneficial to introduce a new argument to `KNNGraph` that would toggle between the vanilla KNNEdges and your new KNNDistanceEdges. For example, one could invent the argument `distance_as_edge_features: bool = False` (defaults to `KNNEdges`) and use `KNNDistanceEdges` if `True`.

src/graphnet/models/components/layers.py
```diff
@@ -593,3 +602,393 @@ def forward(
...
+        """Construct 'GritSparseMHA'.
+
+        Args:
+            in_dim: Dimension of the input tensor.
+            out_dim: Dimension of theo output tensor.
+            num_heads: Number of attention heads.
+            use_bias: Apply bias the key and value linear layers.
+            clamp: Clamp the absolute value of the attention scores to a value.
+            dropout: Dropout layer probability.
+            activation: Activation function.
```
RasmusOrsoe, line 634: Could mention here that the activation function is assumed to be un-instantiated. E.g.
```
"""
activation:  Reference to uninstantiated activation function. E.g. `torch.nn.LeakyReLU`
"""
```

src/graphnet/models/components/layers.py
```diff
@@ -593,3 +602,393 @@ def forward(
...
+        deg_scaler: bool = True,
+        activation: nn.Module = nn.ReLU,
+        norm_edges: bool = True,
+        update_edges: bool = True,
+        batch_norm_momentum: float = 0.1,
+        batch_norm_runner: bool = True,
+        rezero: bool = False,
+        enable_edge_transform: bool = True,
+        attn_bias: bool = False,
+        attn_dropout: float = 0.0,  # CHECK
```
RasmusOrsoe, line 758: Forgot a comment here

src/graphnet/models/components/layers.py
```diff
@@ -593,3 +602,393 @@ def forward(
...
+                norm(
+                    out_dim,
+                    track_running_stats=self.batch_norm_runner,
+                    eps=1e-5,
+                    momentum=self.batch_norm_momentum,
+                )
+                if norm_edges
+                else nn.Identity()
+            )
+        else:  # TODO: Maybe just set this to nn.Identity. -PW
```
RasmusOrsoe, line 847: I think preferred to raise the error if the user passes a non-compatible layer instead of passing warnings and defaulting to identity

src/graphnet/models/components/layers.py
```diff
@@ -593,3 +602,393 @@ def forward(
...
+
+        self.dropout1 = nn.Dropout(dropout)  # Post-attention dropout on x
+        self.dropout2 = nn.Dropout(dropout)  # Post-attention dropout on e
+        self.dropout3 = nn.Dropout(dropout)  # Post-FFN dropout on x
+
+    def forward(self, data: Data) -> Data:
+        """Forward pass."""
+        x = data.x
+        num_nodes = data.num_nodes
+        log_deg = get_log_deg(data)
```
RasmusOrsoe, line 880: Was there some particular reason why you needed the utility function to grab or calculate this quantity?

Naively I would've thought the degree could've been calculated directly in the forward pass like  so: 

```python
from torch_geometric.utils import degree
log_deg = torch.log10(degree(data.edge_index[0]) + 1)
```

src/graphnet/models/components/layers.py
```diff
@@ -593,3 +602,393 @@ def forward(
...
+    def __init__(
+        self,
+        in_dim: int,
+        out_dim: int,
+        num_heads: int,
+        dropout: float = 0.0,
+        norm: nn.Module = nn.BatchNorm1d,
+        residual: bool = True,
+        deg_scaler: bool = True,
+        activation: nn.Module = nn.ReLU,
```
RasmusOrsoe, line 750: Could mention here that the activation function is assumed to be un-instantiated. E.g.
```
"""
activation:  Reference to uninstantiated activation function. E.g. `torch.nn.LeakyReLU`
"""
```

src/graphnet/models/components/layers.py
```diff
@@ -593,3 +602,393 @@ def forward(
...
+
+        Args:
+            in_dim: Dimension of the input tensor.
+            out_dim: Dimension of theo output tensor.
+            num_heads: Number of attention heads.
+            dropout: Dropout layer probability.
+            norm: Normalization layer.
+            residual: Apply residual connections.
+            deg_scaler: Apply degree scaling after MHA.
+            activation: Activation function.
```
RasmusOrsoe, line 773: Could mention here that the activation function is assumed to be un-instantiated. E.g.
```
"""
activation:  Reference to uninstantiated activation function. E.g. `torch.nn.LeakyReLU`
"""
```

src/graphnet/models/gnn/grit.py
```diff
@@ -0,0 +1,160 @@
...
+        pad_to_full_graph: bool = True,
+        add_node_attr_as_self_loop: bool = False,
+        dropout: float = 0.0,
+        fill_value: float = 0.0,
+        norm: nn.Module = nn.BatchNorm1d,
+        attn_dropout: float = 0.2,
+        edge_enhance: bool = True,
+        update_edges: bool = True,
+        attn_clamp: float = 5.0,
+        activation: nn.Module = nn.ReLU,
```
RasmusOrsoe, line 54: Could mention here that the activation function is assumed to be un-instantiated. E.g.
```
"""
activation:  Reference to uninstantiated activation function. E.g. `torch.nn.LeakyReLU`
"""
```

src/graphnet/models/gnn/grit.py
```diff
@@ -0,0 +1,160 @@
...
+        hidden_dim: int,
+        nb_outputs: int = 1,
+        ksteps: int = 21,
+        n_layers: int = 10,
+        n_heads: int = 8,
+        pad_to_full_graph: bool = True,
+        add_node_attr_as_self_loop: bool = False,
+        dropout: float = 0.0,
+        fill_value: float = 0.0,
+        norm: nn.Module = nn.BatchNorm1d,
```
RasmusOrsoe, line 49: Would be informative to mention here that the normalization layer is assumed to be un-instantiated. E.g: 

`norm: Uninstantiated normalization layer. Must be either BatchNorm1d or BatchNorm1d`

src/graphnet/models/components/layers.py
```diff
@@ -593,3 +602,393 @@ def forward(
...
+        activation: nn.Module = nn.ReLU,
+        pooling: str = "mean",
+    ):
+        """Construct `SANGraphHead`.
+
+        Args:
+            dim_in: Input dimension.
+            dim_out: Output dimension.
+            L: Number of hidden layers.
+            activation: Activation function.
```
RasmusOrsoe, line 963: Could mention here that the activation function is assumed to be un-instantiated. E.g.
```
"""
activation:  Reference to uninstantiated activation function. E.g. `torch.nn.LeakyReLU`
"""
```

RasmusOrsoe (changes requested):
Hi @pweigel, thank you for this very clean contribution! 🎸 

I have added superficial comments, mostly about commented-out code and doc strings.

Have you, by chance, tested these different initial representations and their impact on the performance of the GRIT model in a neutrino telescope setting? The original authors appear to favor RRWP over RWSE (I presume they conclude RRPW > RWSE > NoPE), but that may be very problem-dependent.

pweigel:
Hey @RasmusOrsoe, thanks for taking a look. I think I've made all of the requested changes (and fixed a few other minor things). I haven't had the chance to benchmark the different encodings yet, but I plan to. The RRWP encodings are a bit memory-hungry, so I haven't had the chance to fully train a model beyond some tests to show that it works. Without the encodings, I've trained some models that look very good.

At some point in the near future, we should consider a better method of adding the different attributes (`graph.encoding`) in a modular fashion that doesn't require a new graph object. It could even be added as a part of `GraphDefinition`, where you pass some `pos_encoding=MyGraphPosEncoding()` and it uses the values/indices from the edge and node definitions to add the new attributes. It might be beyond the scope of this PR, but it would definitely be an enhancement.

RasmusOrsoe:
> Hey @RasmusOrsoe, thanks for taking a look. I think I've made all of the requested changes (and fixed a few other minor things). I haven't had the chance to benchmark the different encodings yet, but I plan to. The RRWP encodings are a bit memory-hungry, so I haven't had the chance to fully train a model beyond some tests to show that it works. Without the encodings, I've trained some models that look very good.
> 
> At some point in the near future, we should consider a better method of adding the different attributes (`graph.encoding`) in a modular fashion that doesn't require a new graph object. It could even be added as a part of `GraphDefinition`, where you pass some `pos_encoding=MyGraphPosEncoding()` and it uses the values/indices from the edge and node definitions to add the new attributes. It might be beyond the scope of this PR, but it would definitely be an enhancement.

Thanks! I think your idea of adding a positional encoding as a separate module, and storing its values in a dedicated field in the graph structures, is good! I also think this is the intended usage in PyG (see [here](https://pytorch-geometric.readthedocs.io/en/latest/get_started/introduction.html#data-handling-of-graphs)). We've not utilized this in the past because the distinction between having the position as a node feature or as a separately accessible graph feature didn't matter much for the GNN applications we've had so far. I think your use-case is a good example of how it can be beneficial!

It looks like the GitHub rollout to the new ubuntu version has affected your PR checks. @Aske-Rosted has solved this in the main branch (see #779). 

Could you merge the main branch into yours, so the checks pass? The checks should pass after that 💪

RasmusOrsoe (approved):
🚀

Issue #778, opened 2025-01-06
github workflow update


Aske-Rosted:
**Describe the bug**
Github is rolling out a new latest version of Ubuntu for the workflows (https://github.com/actions/runner-images/issues/10636). This breaks a bunch of our workflows with the error seen below . I received this rollout early on my personal repo, and found that the easiest fix for now is to enforce the version of Ubuntu to 22.04.

**Full traceback**
```
Prepare all required actions
Run ./.github/actions/install
Run PIP_FLAG_EDITABLE=`[[ false =~ (T|t)rue ]] && printf "%s\n" "-e " || echo " "`
Run pip install --upgrade pip>=20
Defaulting to user installation because normal site-packages is not writeable
Requirement already satisfied: wheel in /usr/lib/python3/dist-packages (0.42.0)
Collecting setuptools==59.5.0
  Downloading setuptools-59.5.0-py3-none-any.whl.metadata (5.0 kB)
Downloading setuptools-59.5.0-py3-none-any.whl (952 kB)
   ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 952.4/952.4 kB 28.0 MB/s eta 0:00:00
Installing collected packages: setuptools
Successfully installed setuptools-59.5.0
Run echo requirements/torch_cpu.txt    .[develop,torch]
requirements/torch_cpu.txt .[develop,torch]
Defaulting to user installation because normal site-packages is not writeable
Looking in links: https://download.pytorch.org/whl/cpu, https://data.pyg.org/whl/torch-2.2.0+cpu.html
Processing /home/runner/work/graphnet/graphnet
  Preparing metadata (setup.py): started
  Preparing metadata (setup.py): finished with status 'error'
  error: subprocess-exited-with-error
  
  × python setup.py egg_info did not run successfully.
  │ exit code: 1
  ╰─> [11 lines of output]
      Traceback (most recent call last):
        File "<string>", line 2, in <module>
        File "<pip-setuptools-caller>", line 14, in <module>
        File "/home/runner/.local/lib/python3.12/site-packages/setuptools/__init__.py", line 16, in <module>
          import setuptools.version
        File "/home/runner/.local/lib/python3.12/site-packages/setuptools/version.py", line 1, in <module>
          import pkg_resources
        File "/home/runner/.local/lib/python3.12/site-packages/pkg_resources/__init__.py", line 2172, in <module>
[...]

PR #779, opened 2025-01-06
fix ubuntu version

Aske-Rosted:
Enforces ubuntu to version 22.04, which is what we have been using so far. Fixes the issue mentioned in #778


RasmusOrsoe (approved):
🚀

PR #782, opened 2025-01-22
Disable affine option for LearnedTask

pweigel:
For cases where the final layer of a model already reduces the dimension to the number of outputs, you may not want the affine layer in the `LearnedTask`. Added a simple bool for disabling the layer.


RasmusOrsoe (approved):
@pweigel thank you for this contribution!

Aske-Rosted:
With `disable_affine=True` the IdentityTask will be able to run as discussed in #825

Aske-Rosted:
@pweigel Hi Phillip, thanks for this contribution, we believe that the checks that have been run might be a bit updated, would you have the time to re-run the checks and see if we are still compatible with main in the current version? This could be done by merging the updated main into the branch.

PR #783, opened 2025-01-27
add snowstorm_dataset and IceCubehosted class

sevmag:
- Addition of Curated Datasets hosted on the IceCube cluster. (Download requires Username and Password)
- Implementation of the curated IceCube-hosted Snowstorm dataset


src/graphnet/datasets/snowstorm_dataset.py
```diff
@@ -0,0 +1,159 @@
...
+        run_ids: List[int],
+        graph_definition: GraphDefinition,
+        download_dir: str,
+        truth: Optional[List[str]] = None,
+        features: Optional[List[str]] = None,
+        train_dataloader_kwargs: Optional[Dict[str, Any]] = None,
+        validation_dataloader_kwargs: Optional[Dict[str, Any]] = None,
+        test_dataloader_kwargs: Optional[Dict[str, Any]] = None,
+    ):
+        """Initialize SnowStorm dataset."""
```
RasmusOrsoe, line 48: I think we should explain the arguments here. Most can be repeated from the [parent class](https://github.com/graphnet-team/graphnet/blob/c31582f4b5f42bf99fb6db466342714af0ca3859/src/graphnet/data/curated_datamodule.py#L27) but `run_ids` is new.

src/graphnet/datasets/snowstorm_dataset.py
```diff
@@ -0,0 +1,159 @@
...
+        event_counts: Dict[str, int] = {}
+        event_counts = {}
+        for path in dataset_paths:
+
+            # Extract the ID
+            match = re.search(pattern, path)
+            assert match
+            run_id = match.group(1)
+
+            query_df = query_database(
```
RasmusOrsoe, line 91: How fast is this query in your experience? If the database is large, it might take minutes to execute. In that case, we could consider providing a `.parquet` file with the ids in them.

src/graphnet/datasets/snowstorm_dataset.py
```diff
@@ -0,0 +1,159 @@
...
+    """IceCube SnowStorm simulation dataset.
+
+    More information can be found at
+    https://wiki.icecube.wisc.edu/index.php/SnowStorm_MC#File_Locations
+    This is a IceCube Collaboration simulation dataset.
+    Requires a username and password.
+    """
+
+    _experiment = "IceCube SnowStorm dataset"
+    _creator = "Severin Magel"
```
RasmusOrsoe, line 26: Should probably mention @Aske-Rosted here :-)

src/graphnet/datasets/snowstorm_dataset.py
```diff
@@ -0,0 +1,159 @@
...
+        truth: Optional[List[str]] = None,
+        features: Optional[List[str]] = None,
+        train_dataloader_kwargs: Optional[Dict[str, Any]] = None,
+        validation_dataloader_kwargs: Optional[Dict[str, Any]] = None,
+        test_dataloader_kwargs: Optional[Dict[str, Any]] = None,
+    ):
+        """Initialize SnowStorm dataset."""
+        self._run_ids = run_ids
+        self._zipped_files = [
+            os.path.join(self._data_root_dir, f"{s}.tar.gz") for s in run_ids
```
RasmusOrsoe, line 51: I think there's currently 28 valid run id's, which is a subset of the overall available sets. I think the current code assumes the user knows this. Perhaps it would be wise to include an assertion that checks that the user-given ids actually exists.

RasmusOrsoe (commented):
Thank you for this very clean contribution @sevmag!

Before we proceed, do you have comments on this @Aske-Rosted?

src/graphnet/datasets/snowstorm_dataset.py
sevmag: Yes, I agree!

src/graphnet/datasets/snowstorm_dataset.py
sevmag: True! Changed it

src/graphnet/datasets/snowstorm_dataset.py
sevmag: Yeah, you are right. It depends on the files, but some take over 2 minutes, and it's probably the query.

sevmag: UPDATE: The bottleneck of initializing this `SnowStormDataset` implementation is mainly in the initialization called here (especially when selecting RunIDs with a lot of files)

https://github.com/graphnet-team/graphnet/blob/79d7baf1c5a52bf5384fc12b7e3d1f3e45a15e5c/src/graphnet/data/curated_datamodule.py#L76-L85

Contrary to my belief that it was the  `prepare_args` function snipped referenced above. Therefore, I decided to stick to this version, keeping in mind that one can make the `prepare_args` function more efficient.

RasmusOrsoe (approved):
Great contribution 🚀 ! @Aske-Rosted Any suggestions from you? To me this looks ready.

PR #785, opened 2025-02-12
fix readme link

Aske-Rosted:
readme link leads to a file that no longer exists, this fixes the target location.


RasmusOrsoe (approved):
🚀

Issue #786, opened 2025-02-17
Memory leak when using a list of SQLite files in GraphNeTDataModule


jvaracarbonell:
I encountered a memory leak when using a list of SQLite files as input to [GraphNeTDataModule](https://github.com/graphnet-team/graphnet/blob/79d7baf1c5a52bf5384fc12b7e3d1f3e45a15e5c/src/graphnet/data/datamodule.py#L19). This pull request addresses the issue and appears to resolve the leak.

PR #787, opened 2025-02-17
Fix memory leak.

jvaracarbonell:
This PR fixes [#786](https://github.com/graphnet-team/graphnet/issues/786).


RasmusOrsoe:
Hi @jvaracarbonell! Could you provide a few more details regarding the issue you opened in #787 and how this pull request solves it?

jvaracarbonell:
Hello @RasmusOrsoe! 

I encountered a memory leak when using the GraphNeTDataModule with a list of SQLite files as input. For example:

 ```
   files = glob.glob("/scratch/tmp/fvaracar/graphnet_training_data/NuMu/quality_cut/*/*/*db")
    
    dm = GraphNeTDataModule(
        dataset_reference = SQLiteDataset,
        selection=None,
        dataset_args={
            "truth": truth,
            "truth_table": truth_table,
            "features": features,
            "graph_definition": graph_definition,
            "pulsemaps": [config["pulsemap"]],
            "path": files,
            "index_column": "event_no",
            "labels": {"direction": Direction(
                azimuth_key="azimuth", zenith_key="zenith"
            )},
        },
        train_dataloader_kwargs={
            "batch_size": config["batch_size"],
            "num_workers": config["num_workers"],
            "shuffle": True,
        },
        test_selection = [None for _ in range(len(files))],
        test_dataloader_kwargs={
            "batch_size": config["batch_size"],
            "num_workers": config["num_workers"],
            "shuffle": True,
        },
    )

    training_dataloader = dm.train_dataloader
    validation_dataloader = dm.val_dataloader

```

My CPU RAM was steadily filling up with each iteration until my jobs were eventually killed. I noticed that closing the connection in the [query_table method of the SQLiteDataset module](https://github.com/jvaracarbonell/graphnet/blob/78cd8a2d826f10027e5518129ef8860d7d307b5c/src/graphnet/data/dataset/sqlite/sqlite_dataset.py#L84) prevents this memory leak.

I'm not entirely sure if this is the optimal solution, so any suggestions would be welcome.

By the way, I apologize for not following the [Graphnet contribution guidelines](https://graphnet-team.github.io/graphnet/contribute/contribute.html)—specifically, merging from my fork’s main branch instead of a separate branch—as the guidelines link did not seem to work when I submitted this.

RasmusOrsoe:
> Hello @RasmusOrsoe!
> 
> I encountered a memory leak when using the GraphNeTDataModule with a list of SQLite files as input. For example:
> 
> ```
>   files = glob.glob("/scratch/tmp/fvaracar/graphnet_training_data/NuMu/quality_cut/*/*/*db")
>    
>    dm = GraphNeTDataModule(
>        dataset_reference = SQLiteDataset,
>        selection=None,
>        dataset_args={
>            "truth": truth,
>            "truth_table": truth_table,
>            "features": features,
>            "graph_definition": graph_definition,
>            "pulsemaps": [config["pulsemap"]],
>            "path": files,
>            "index_column": "event_no",
>            "labels": {"direction": Direction(
>                azimuth_key="azimuth", zenith_key="zenith"
>            )},
>        },
>        train_dataloader_kwargs={
>            "batch_size": config["batch_size"],
>            "num_workers": config["num_workers"],
>            "shuffle": True,
>        },
>        test_selection = [None for _ in range(len(files))],
>        test_dataloader_kwargs={
>            "batch_size": config["batch_size"],
>            "num_workers": config["num_workers"],
>            "shuffle": True,
>        },
>    )
> 
>    training_dataloader = dm.train_dataloader
>    validation_dataloader = dm.val_dataloader
> ```
> 
> My CPU RAM was steadily filling up with each iteration until my jobs were eventually killed. I noticed that closing the connection in the [query_table method of the SQLiteDataset module](https://github.com/jvaracarbonell/graphnet/blob/78cd8a2d826f10027e5518129ef8860d7d307b5c/src/graphnet/data/dataset/sqlite/sqlite_dataset.py#L84) prevents this memory leak.
> 
> I'm not entirely sure if this is the optimal solution, so any suggestions would be welcome.
> 
> By the way, I apologize for not following the [Graphnet contribution guidelines](https://graphnet-team.github.io/graphnet/contribute/contribute.html)—specifically, merging from my fork’s main branch instead of a separate branch—as the guidelines link did not seem to work when I submitted this.

Hi @jvaracarbonell! Thanks for the details.  

In this pull request you close the sqlite connection in the query function, which means every example retrieval will begin with opening a sqlite connection and end with closing it. Because this function is called many times during training, the small computation overhead of closing and opening the connection adds up, and this suggested change will therefore affect other users negatively.  Closing a sqlite connection after a select statement shouldn't result in a significant release in memory, so I think we should dig a little deeper to figure out what you experienced, and preferably reproduce it.

When you instantiate the data module without a selection, the module will create a selection automatically for each of your databases. These selections are stored in memory per worker, and you should expect to see a steadily increasing memory usage until all selections are in memory. Perhaps this is what you saw.

How many databases did you use? How much ram did you specify for the job? How many workers did you use?

jvaracarbonell:
Hi @RasmusOrsoe. I am currently using 14 workers, 70-80 GB of RAM, and about 25,000 SQLite files as input. My main goal was to use this dataloader to avoid having to merge the individual SQLite files and to convert the i3 files to SQLite using multiple parallel jobs.

A few weeks ago, I submitted some jobs to monitor RAM usage versus training steps, and I noticed that the RAM was slowly filling up over iterations. However, after closing that connection, this issue did not occur again. This was mostly trial and error, so I should probably take some time to better understand how that part of the code really works when I get a chance.

RasmusOrsoe:
Aha! OK. The problem here is the  25,000 SQLite databases. 

When we began discussing using multiple databases for training, it was for stitching together a training dataset by recycling previous data conversions, so one did not need to convert the same files again. For example, one database would contain atmospheric muons, whereas one might contain neutrino interactions. I.e., the number of databases would be O(10), not O(10.000). 

The logic for handling the multiple connections is here: https://github.com/graphnet-team/graphnet/blob/941c9cd166880ed4898652a7ee5e117e1eed88ed/src/graphnet/data/dataset/sqlite/sqlite_dataset.py#L110 (and could use a polish..) 
As you can see, it initially creates a connection to all databases to avoid that overhead in the forward pass. So that means, in your case, you're creating 25.000 connections in one go. This would never work for O(10.000) connections as each connection count as a file descriptor, and systems have [a hard limit on the number of open files](https://www.baeldung.com/linux/limit-file-descriptors).

jvaracarbonell:
@RasmusOrsoe Thank you for your response! I think it makes sense to cancel this PR, or to move it to draft status until I come up with a better way to handle it.

There are three main reasons why I did not merge the files—and why I probably would prefer not to do so in the future:

1.- (Least important reason): To speed up processing time. If I could work directly on the individual files, why wait until they are merged.

2.- My dataset is quite large, and I cannot convert and merge all the files within the 48-hour window that condor usually allows (actually not even 1000 files from the high energies set). I previously tried merging files from several different jobs, but since I believe each job creates its own dataset, I think I encountered event index problems. Honestly, I did not look deeply into this because I did not consider it worth the time if I could simply use the individual files instead.

3.- (Most important reason): My dataset in .db format occupies well over 1TB of space. Since I was processing everything on the IceCube cluster, both the individual files and the merged files exceeded my disk quota. I have now requested additional space on /data/ana, which should help prevent this issue.

Maybe all these issues have quick fixes, but I thought it was not worth the time to investigate them if I could simply use the individual files.

PR #791, opened 2025-04-21
Sqlite string selection

Aske-Rosted:
This allows for using the sqlite3 backend for string selection instead of the pandas selection, there are slight differences between the two and the sqlite selections allows for more flexibility.

This was part of #789 but have been turned into it's own PR.


RasmusOrsoe:
Hi  @Aske-Rosted! Could you elaborate on the benefits of these modifications? Is it primarily that they allow you to run SQL in the selection field of the config?

PR #792, opened 2025-04-22
I3Filters

Aske-Rosted:
Adds two different filters that can be used when extracting from I3 files. 

TableFilter

> Checks if table exists in the events

ChargeFilter

> Checks that value in column is higher than given threshold standard column `"Homogenized_QTot"`


RasmusOrsoe (approved):
🚀

PR #794, opened 2025-04-22
Writer changes

Aske-Rosted:
This contains three changes to the dataconverter/writer. 

- Proper parsing of max table size
- Add capabilities for removing files after merging
- Add option for giving a list of files instead of a folder (This can be useful when converting files in a condor job on the cluster)


RasmusOrsoe (approved):
🚀

PR #795, opened 2025-04-22
Converter extractor changes

Aske-Rosted:
changes to the converter and extractor modules.

- Adds option to change the output dir of the merger
- Properly set the gcd file in the CombinedExtractor
- Adds option to exclude some of the values extracted by the extractor. 

### DIFF
--- a/src/graphnet/data/extractors/extractor.py
+++ b/src/graphnet/data/extractors/extractor.py
@@ -1,6 +1,6 @@
 """Base I3Extractor class(es)."""
 
-from typing import Any, Union
+from typing import Any, Union, Callable
 from abc import ABC, abstractmethod
 import pandas as pd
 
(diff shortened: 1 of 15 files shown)


src/graphnet/data/extractors/extractor.py
```diff
@@ -47,3 +68,8 @@ def __call__(self, data: Any) -> Union[dict, pd.DataFrame]:
     def name(self) -> str:
         """Get the name of the `Extractor` instance."""
         return self._extractor_name
+
+    def __init_subclass__(cls) -> None:
+        """Initialize subclass and apply the exclude decorator to __call__."""
+        super().__init_subclass__()
+        cls.__call__ = cls.exclude(cls.__call__)  # type: ignore
```
RasmusOrsoe, line 75: That is a pretty elegant solution!

RasmusOrsoe (approved):
🚀

PR #800, opened 2025-04-30
move examples dependencies into their own "extras" section.

nega0:
AwkwardArray doesn't build with CMake v4 and prevents building graphnet.
This moves it and `timers` into an examples "extra" since that's the
only place they're used.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -10,7 +10,6 @@ SETUP_REQUIRES = [
 ]
 
 INSTALL_REQUIRES = [
-    "awkward>=1.8,<2.0",
     "colorlog>=6.6",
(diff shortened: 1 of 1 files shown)


RasmusOrsoe:
Hi @nega0! Thank you for reporting this, and apologies for the slow response.

The usage of these libraries has decreased quite a bit (we used to use them more), and as you point out here, including these as dependencies in graphnet is unnecessary.  

I've created a pull request #808 that removes these two libraries. Given that this PR has failing checks, I think we should close this PR in favor of #808 as they achieve the same. Is that OK with you?

PR #802, opened 2025-05-20
I3dictvalueextractor

Aske-Rosted:
Extracts features from dictionary like data objects in the I3-files e.g.  `I3MCWeightDict`.


RasmusOrsoe (approved):
🚀

Issue #803, opened 2025-05-23
Undeterministic behaviour in deployment module


Aske-Rosted:
**Describe the bug**
Running the deployment module several times on the same I3 file will not produce the same prediction

**Expected behavior**
Deployment should be deterministic.

**Additional context**
I have already found a fix, in `class DeploymentModule` the model is correctly set to inference mode, however it is also necessary to call `self.model.train(mode=False)`.

PR #804, opened 2025-05-23
Set model.train(mode=False)

Aske-Rosted:
Simple fix to #803


RasmusOrsoe (approved):
Thanks for catching this @Aske-Rosted!

PR #805, opened 2025-05-28
Fix paper workflow

sevmag:
Upgrading the upload-artifact version to v4 as recommended in [this article](https://github.blog/changelog/2024-04-16-deprecation-notice-v3-of-the-artifact-actions/). Mitigates the failing of paper workflow.


RasmusOrsoe (approved):
@sevmag thanks for solving this :rocket:

PR #807, opened 2025-06-02
Mc labels

Aske-Rosted:
The final _big one_ with the two new extractors `I3HighestEparticleExtractor, I3Calorimetry`. There is a lot of supporting work that had to be done in order to enable these extractors/new labels. 

An example file is also included , which gives an example of how one might run their extraction.

see #789 for the original PR.

### DIFF
--- a/examples/01_icetray/05_convert_i3_files_advanced.py
+++ b/examples/01_icetray/05_convert_i3_files_advanced.py
@@ -54,7 +54,11 @@ def main(
     max_table_size: int = int(25e7),
     converter_class: str = "sqlite",
 ) -> None:
-    """Convert IceCube-Upgrade I3 files to intermediate `backend` format."""
+    """Convert IceCube I3 files tosqlite/parquet.
+
+    Advanced example script to convert IceCube I3 files to
+    sqlite/parquet.
+    """
     # Check(s)
     inputs = [f"{TEST_DATA_DIR}/i3/nugen_ftp-v3_level2"]
     outdir = f"{EXAMPLE_OUTPUT_DIR}/convert_i3_files/ic86_advanced"


src/graphnet/data/extractors/icecube/i3extractor.py
```diff
@@ -20,20 +25,28 @@ class I3Extractor(Extractor):
...
+    ):
         """Construct I3Extractor.
 
         Args:
             extractor_name: Name of the `I3Extractor` instance. Used to keep
                 track of the provenance of different data, and to name tables
                 to which this data is saved.
-            exclude: List of features to exclude from the extractor.
+            exclude: List of keys to exclude from the extracted data.
+            is_corsika: Boolean indicating if the event files being processed
```
RasmusOrsoe, line 41: Are there not some entry in the frame that could replace this argument?

src/graphnet/data/extractors/icecube/i3calorimetry.py
```diff
@@ -0,0 +1,237 @@
...
+    from icecube import (
+        icetray,
+        dataclasses,
+        MuonGun,
+        simclasses,
+    )  # pyright: reportMissingImports=false
+
+
+class I3Calorimetry(I3Extractor):
+    """Energy on appearance of all visible particles in the volume."""
```
RasmusOrsoe, line 22: Since this extracts the energy on appearance, maybe a more suitable name for the extractor would be `I3EnergyOnAppearanceExtractor`, or something similar?

examples/01_icetray/05_convert_i3_files_advanced.py
```diff
@@ -0,0 +1,179 @@
...
+    """Convert IceCube-Upgrade I3 files to intermediate `backend` format."""
+    # Check(s)
+    inputs = [f"{TEST_DATA_DIR}/i3/oscNext_muongun_level3_v02"]
+    outdir = f"{EXAMPLE_OUTPUT_DIR}/convert_i3_files/ic86_advanced"
+    gcd_rescue = glob(
+        f"{TEST_DATA_DIR}/i3/oscNext_muongun_level3_v02/*GeoCalib*"
+    )[0]
+
+    # Create hulls, these need to be global for the multiprocessing.
+    IceCube_hull_extended = GCD_hull(gcd_rescue, padding=padding)
```
RasmusOrsoe, line 63: `IceCube_hull_extended` -> icecube_hull_extended

examples/01_icetray/05_convert_i3_files_advanced.py
```diff
@@ -0,0 +1,179 @@
...
+    highest_e_particle = I3HighestEparticleExtractor(
+        extractor_name="HEP",
+        hull=IceCube_hull_extended,
+        daughters=True,
+        is_corsika=False,
+    )
+
+    # Comparisons to other reconstruction methods
+
+    SplineMPE = I3ParticleExtractor(extractor_name="OnlineL2_SplineMPE")
```
RasmusOrsoe, line 106: `SplineMPE` -> `splinempe` or `spline_mpe`

examples/01_icetray/05_convert_i3_files_advanced.py
```diff
@@ -0,0 +1,179 @@
...
+    LineFit = I3ParticleExtractor(extractor_name="LineFit")
```
RasmusOrsoe, line 107: `LineFit` -> `linefit`

examples/01_icetray/05_convert_i3_files_advanced.py
```diff
@@ -0,0 +1,179 @@
...
+        extractors=extractors,
+        outdir=outdir,
+        num_workers=workers,
+        gcd_rescue=gcd_rescue,
+        i3_filters=filters,
+        max_table_size=max_table_size,
+    )
+    # run the converter
+
+    print(f"converting {inputs} to {outdir}")
```
RasmusOrsoe, line 146: Lets use the logger, like in the other examples. I.e.

`logger.info(str)`

examples/01_icetray/05_convert_i3_files_advanced.py
```diff
@@ -0,0 +1,179 @@
...
+def main(
+    merge: bool,
+    remove: bool,
+    workers: int = 1,
+    padding: int = 250,
+    max_table_size: int = int(25e7),
+) -> None:
+    """Convert IceCube-Upgrade I3 files to intermediate `backend` format."""
+    # Check(s)
+    inputs = [f"{TEST_DATA_DIR}/i3/oscNext_muongun_level3_v02"]
```
RasmusOrsoe, line 56: Is this test file sufficient for the example, or should be add a new one?

examples/01_icetray/05_convert_i3_files_advanced.py
```diff
@@ -0,0 +1,179 @@
...
+
+
+def main(
+    merge: bool,
+    remove: bool,
+    workers: int = 1,
+    padding: int = 250,
+    max_table_size: int = int(25e7),
+) -> None:
+    """Convert IceCube-Upgrade I3 files to intermediate `backend` format."""
```
RasmusOrsoe, line 54: Lets update this string

src/graphnet/data/extractors/icecube/i3calorimetry.py
```diff
@@ -0,0 +1,237 @@
...
+        # Member variable(s)
+        self.hull = hull
+        self.mctree = mctree
+        self.mmctracklist = mmctracklist
+        self.daughters = daughters
+        # Base class constructor
+        super().__init__(extractor_name=extractor_name, **kwargs)
+
+    def __call__(self, frame: "icetray.I3Frame") -> Dict[str, Any]:
+        """Extract all the visible particles entering the volume."""
```
RasmusOrsoe, line 43: Extract **energy** for all ...

RasmusOrsoe (commented):
Hi @Aske-Rosted - Thank you for this great contribution!

I've taken a look and added a few minor comments. They mostly revolve around snakecase, docstrings and naming.

src/graphnet/data/extractors/icecube/i3extractor.py
```diff
@@ -86,8 +99,116 @@ def set_gcd(self, i3_file: str, gcd_file: Optional[str] = None) -> None:
...
+        frame: "icetray.I3Frame",
+        primaries: Union[
+            "dataclasses.ListI3Particle", "dataclasses.I3Particle"
+        ],
+    ) -> Union["dataclasses.ListI3Particle", "dataclasses.I3Particle"]:
+        """Check that primary energy exists for the particle(s).
+
+        If the primary energy is not available, the method will see if the
+        particle has a single daughter, and if so, return that
+        """
```
sevmag, line 123: Missing Args, description

src/graphnet/data/extractors/icecube/i3extractor.py
```diff
@@ -86,8 +99,116 @@ def set_gcd(self, i3_file: str, gcd_file: Optional[str] = None) -> None:
...
+            self.warning_once("Primary energy is nan checking daughters")
+            daughters = dataclasses.I3MCTree.get_daughters(
+                frame[self.mctree], primary
+            )
+            if len(daughters) == 0:
+                raise ValueError(
+                    "Primary energy is nan and no daughters found"
+                )
+            primary = dataclasses.ListI3Particle()
+            for daughter in daughters:
```
sevmag, line 158: In the docstring, you say "... will see if the particle has a single daughter ..." but here it seems as if you account for multiple daughters. Am I missing something?

src/graphnet/data/extractors/icecube/i3highesteparticleextractor.py
```diff
@@ -0,0 +1,750 @@
...
+        energies, lengths = np.array(
+            [[p.energy, p.length] for p in particles]
+        ).T
+        loc_max = np.argmax(energies)
+        # Inherit from the highest energy particle
+        bundle = particles[loc_max]
+
+        intersections = self.hull.surface.intersection(bundle.pos, bundle.dir)
+
+        assert not np.isnan(intersections.first), "No intersection found"
```
sevmag, line 239: Can it not happen that an event passes the `self.hull.rays_and_sphere_intersection_check` check, so intersecting with the sphere, but not the actual detector? In this case, this assertion would kill the process, right?

src/graphnet/data/extractors/icecube/i3calorimetry.py
```diff
@@ -0,0 +1,237 @@
...
+        for track in MuonGun.Track.harvest(frame[self.mctree], MMCTrackList):
+            assert track.is_track, "Track is not a track"
+
+            # Find distance to entrance and exit from sampling volume
+            intersections = self.hull.surface.intersection(
+                track.pos, track.dir
+            )
+            # Get the corresponding energies
+
+            e0 = track.get_energy(intersections.first)
```
sevmag, line 137: What will happen here for a starting track? I'm not familiar with MuonGun, but there shouldn't be any `intersections.first`, or it is not the entry but the exit, logically speaking.

src/graphnet/data/extractors/icecube/i3highesteparticleextractor.py
Aske-Rosted: I believe the case that you are describing is possible. I have however not run into it during my conversion, possible because of the buffer that my conversion employed. Since this whole thing is based on the assumption that the particles are in a relatively tight bundle (which I believe is the case for the corsika simulation) I believe that it would be okay to then simply "abort" and return the fill values for this event. Do you agree with this handling of the event?

sevmag: You could take the next highest energetic particle in the bundle iteratively until you have one that has an intersection, and include the misses in the length_mask. However, I see your point that this scenario is extremely rare, so maybe it's a bit overkill. I think you know best what to do. In case you use the default value, maybe mention this case somewhere so people are not confused when this happens.

src/graphnet/data/extractors/icecube/i3extractor.py
Aske-Rosted: There are ways to check this in the frame, but then you would have to check this for every frame, and the I3Extractor itself actually does not currently do anything on the per frame level.

src/graphnet/data/extractors/icecube/i3calorimetry.py
Aske-Rosted: The discerning difference between the `I3Calorimetry` approach and the `I3HighestEparticleExtractor` is that the `I3Calorimetry` approach tries to look at everything that enters the detector volume (only daughters of the primary if daughters=True). Whereas, the `I3HighestEparticleExtractor` only tries to reconstruct the _single_ highest energy detectable particle entering the detector. The _calorimetric_ approach provides several different types of energy labels which can be reconstructed and therefor energy on appearance. I will try to update the docstring to make this more clear.

examples/01_icetray/05_convert_i3_files_advanced.py
Aske-Rosted: The current file does produce any events meaning a db is not being constructed, it would be nice if we could add a file nugen file with an event passing the charge cut (charge cut could also be reduced).

examples/01_icetray/05_convert_i3_files_advanced.py
```diff
@@ -0,0 +1,179 @@
...
+
+if __name__ == "__main__":
+
+    if not has_icecube_package():
+        Logger(log_folder=None).error(ERROR_MESSAGE_MISSING_ICETRAY)
+    else:
+        # Parse command-line arguments
+        parser = ArgumentParser(
+            description="""
+Convert I3 files to an intermediate format.
```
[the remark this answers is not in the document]
Aske-Rosted: sqlite and parquet inputs have started to diverge slightly I can also add parquet but it means that some arguments such as `max_table_size` and `remove` are not used.

src/graphnet/data/extractors/icecube/i3calorimetry.py
Aske-Rosted: the intersections will in that case produce a negative value for intersections.first and I think the energy which will be used in that case is the energy of the particle at generation.

examples/01_icetray/05_convert_i3_files_advanced.py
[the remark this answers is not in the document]
RasmusOrsoe: Yes, but let's not have it diverge further :-D. Could you adjust it such that it is similar to the other conversion examples, i.e. backend agnostic? The current difference in arguments can be handled by checking if the backend is sqlite

Issue #817, opened 2025-08-12
Unexpected behaviour in I3Extractor.get_primaries


sevmag:
**Describe the bug**  

There are two issues arising in the following function:

<https://github.com/graphnet-team/graphnet/blob/2f61199a1280652e804e2aa7823d79688f1b5a19/src/graphnet/data/extractors/icecube/i3extractor.py#L166>  

# 1. Unexpected handling of searching for in-ice neutrinos if daughters==True

The issue arises when processing a NuGen file with the `daughters` option set to `True`.  

If there is no in-ice neutrino at the highest level of the `MCTree`, the following code block returns an empty `primaries` list:  

<https://github.com/graphnet-team/graphnet/blob/2f61199a1280652e804e2aa7823d79688f1b5a19/src/graphnet/data/extractors/icecube/i3extractor.py#L184-L195>  

This behaviour is expected. The neutrino can interact outside of the ice multiple times, and therefore, the neutrinos close to the root of the MCTree do not necessarily need to be in-ice. However, when this happens, the next step is to take *any* in-ice neutrino, as done here:  

<https://github.com/graphnet-team/graphnet/blob/2f61199a1280652e804e2aa7823d79688f1b5a19/src/graphnet/data/extractors/icecube/i3extractor.py#L197-L208>  

The problem is that this can include, for example, a `NuE` created at the very end of a muon track during its decay. Such particles should not be returned as primaries. Moreover, these particles—now included in the `primaries` list and located far down in the `MCTree`—can be daughters of other particles already in the `primaries` list, and they can even be particles that are not daughters of the primary neutrino but background particles. This can lead to unwanted side effects, such as double-counting energies and including energies from different particles in the `I3Calorimetry` extractor.

# 2.Unexpected handling of multiple found primaries for NuGen 

The problem concerns the following code block:

https://github.com/graphnet-team/graphnet/blob/2f61199a1280652e804e2aa7823d79688f1b5a19/src/graphnet/data/extractors/icecube/i3extractor.py#L214-L217

Imagine a NuGen simulation where multiple primaries are found. This can happen, e.g. when:

1.  daughters==False the primaries might be a list like `[neutrino, CR_PiPlus, CR_PiPlus, CR_PiPlus]`
2. daughters==True and the situation from the bug above arises, where then multiple in-ice neutrinos are found.

In my mind, we should not only return the highest energy particle in this list, but there should also be an option to return all of these primaries.

# Notes

[...]

PR #818, opened 2025-08-12
Fixing the get_primaries function of I3Extractor

sevmag:
This PR addresses the problem in #817.

As a fix to Problem 1 explained in #817, I introduce a function `find_in_ice_daughters` that recursively looks for in-ice particles in the subtree of the primary neutrino.

This solution makes sure that ...
- one finds the in-ice particle, no matter how many interactions the primary neutrino undergoes outside of ice. 
-  only the first in-ice particle is returned, mitigating the issue of returning a `primaries` list of particles where one of the particles might be a daughter of another (as explained in #817)
- no particle that is a daughter of a background particle is returned

As a fix to Problem 2 explained in #817, I introduce the argument `highest_energy_primary` that allows to return all primaries found instead of only the highest-energy one in the case of a NuGen file

### DIFF
--- a/src/graphnet/data/extractors/icecube/i3extractor.py
+++ b/src/graphnet/data/extractors/icecube/i3extractor.py
@@ -229,17 +229,14 @@ class I3Extractor(Extractor):
                         self.mctree,
                     )
 
-                    # This is not expected to happen
-                    if len(primaries) == 0:
-                        self.warning_once(
-                            "No in-ice neutrino found for NuGen event, "
-                            "returning no primaries."
-                        )
-                        return dataclasses.ListI3Particle([])
+                # This is not expected to happen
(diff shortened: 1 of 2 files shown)


src/graphnet/data/extractors/icecube/i3extractor.py
```diff
@@ -194,27 +209,71 @@ def get_primaries(
...
+
+                    # recursively search for in-ice neutrino daughters
+                    primaries = self.find_in_ice_daughters(
+                        frame,
+                        primary_nus,
+                        self.mctree,
                     )
-                ]
+
+                    # This is not expected to happen
```
Aske-Rosted, line 232: Is this still true, do all NuGen events have a InIce neutrino generated by thedaughter particles?

Issue #821, opened 2025-08-20
Bug in I3HighestEparticleExtractor


sevmag:
**Describe the bug**

This issue lists a few bugs of the I3HighestEparticleExtractor

# 1. Negative visible length

The visible length as computed for tracks can be negative if both intersection points are negative. To my knowledge, this happens when the particle never intersects the hull. (I'm not too sure about that)

https://github.com/graphnet-team/graphnet/blob/f3faff18f6caeddb7159b9d02910e17e33b8e9ad/src/graphnet/data/extractors/icecube/i3highesteparticleextractor.py#L336-L338

# 2. Track filtering only for top-level primaries

Currently, filtering for tracks that are in the subtrees of the primaries is handled here:

https://github.com/graphnet-team/graphnet/blob/f3faff18f6caeddb7159b9d02910e17e33b8e9ad/src/graphnet/data/extractors/icecube/i3highesteparticleextractor.py#L186-L192

The primaries are coming from the `get_primary` function, which can also return particles that are not in the head of the MCTree. 

**the `get_primaries` funciton:**
https://github.com/graphnet-team/graphnet/blob/f3faff18f6caeddb7159b9d02910e17e33b8e9ad/src/graphnet/data/extractors/icecube/i3extractor.py#L166

In the scenario that the returned primaries are not in the head of the MCTree, the filtering logic will always filter out the tracks even if they are descendants of particles in the `primaries` list. This is because `frame[self.mctree].get_primary()` will only give particles at the head level.

same issue occurs here:

https://github.com/graphnet-team/graphnet/blob/f3faff18f6caeddb7159b9d02910e17e33b8e9ad/src/graphnet/data/extractors/icecube/i3highesteparticleextractor.py#L399-L414

This problem leads to the energy proxies calculated by this extractor to be 0 or inaccurate if the particles in the primaries are not in the top level of the MCTree. 

# 3. Wrong return in `get_bundle_HEP`

This returns a single particle

https://github.com/graphnet-team/graphnet/blob/f3faff18f6caeddb7159b9d02910e17e33b8e9ad/src/graphnet/data/extractors/icecube/i3highesteparticleextractor.py#L232

However, the function is expected to return a Tuple as seen here:

https://github.com/graphnet-team/graphnet/blob/f3faff18f6caeddb7159b9d02910e17e33b8e9ad/src/graphnet/data/extractors/icecube/i3highesteparticleextractor.py#L672

[...]

PR #822, opened 2025-08-20
Fix I3HighestEparticleExtractor

sevmag:
This PR addresses the bugs explained in #821.

# Solving Issues

## 1. Negative visible length
set visible length to 0 if it would be negative

## 2. Track filtering only for top-level primaries
Look, if a track particle is in the subtree of any primary, not if the head-level primary is in the `primaries` list

## 3. Wrong return in `get_bundle_HEP`
return the right tuple

Additionally, I introduced the `get_descendants` function, which accelerates this code snippet:

https://github.com/graphnet-team/graphnet/blob/f3faff18f6caeddb7159b9d02910e17e33b8e9ad/src/graphnet/data/extractors/icecube/i3highesteparticleextractor.py#L399-L414

by not iterating through all particles, but only the descendants of the primaries

### DIFF
--- a/src/graphnet/data/extractors/icecube/i3highesteparticleextractor.py
+++ b/src/graphnet/data/extractors/icecube/i3highesteparticleextractor.py
@@ -429,9 +429,9 @@ class I3HighestEparticleExtractor(I3Extractor):
                 visible_length,
                 -1,
             )
-        else:
-            energies = e_p[0]
-            particles = e_p[1]
+
+        energies = e_p[0]
+        particles = e_p[1]
 
         pos, direc, lengths = self.get_pos_dir_length(particles)
         pos = pos + direc * lengths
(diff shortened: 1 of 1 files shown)


Aske-Rosted (commented):
Looks good to me! Nice contribution, thanks.

sevmag:
I made some more commits. They address further issues I explained in #821. Let me know if you agree, @Aske-Rosted.

Aske-Rosted:
> I made some more commits. They address further issues I explained in #821. Let me know if you agree, @Aske-Rosted.

I agree with the new changes as well, I also ran into the muongun issues a while back, but managed to remove them with the logic statements I had in play, guess some of the changes to the logic made them reappear...

Aske-Rosted (approved):

PR #828, opened 2025-09-05
Add ESSnuSB paper to publications list

kaareendrup:
Add the [recently published paper](https://iopscience.iop.org/article/10.1088/1748-0221/20/08/P08030) on ESSnuSB event classification to the list of publications. I put it at the bottom, but we can move it around if you want.


Aske-Rosted (approved):
Congratulations on the paper! 🎉 

I do not believe that there is any particular order to the publications list so looks fine to me.

PR #851, opened 2025-12-01
BERT-style mask prediction pretraining

niklasmei:
This adds a module for unsupervised pretraining using BERT-style mask prediction.

I originally made it as to pretrain a model that returns a latent view of input sequences along with a single vector to represent the sequences. This was the case because I used a model based on DeepIce, where I had the cls-token and the processed sequence. Some variables still reflect the original use in their names.

In the version here I made it optional to provide a vector that summarizes the input data, which is only used to predict some summary feature (the total charge in an event in the standard case).

It is important that the model that is to be pretrained does not change the number of sequence elements beyond providing an additional summary vector, like a cls-token. Other than that this pretraining module should be indifferent to the model that is pretrained.

### DIFF
--- a/src/graphnet/models/pretraining_maskpred.py
+++ b/src/graphnet/models/pretraining_maskpred.py
@@ -1,4 +1,7 @@
-"""Exp. feature: Self-supervised pretraining using BERT-style mask prediction."""
+"""Experimental feature.
+
+Self-supervised pretraining using BERT-style mask prediction.
+"""
 
 from typing import Any, Tuple, Union, List, Type, Optional, Dict
 import os


examples/pretraining_maskpred_example.py
```diff
@@ -0,0 +1,117 @@
...
+        )
+
+    def forward(self, data: Data) -> Tuple[Tensor, Tensor]:
+        """Forward pass."""
+        x = self.net(data.x)
+        x_rep = scatter(src=x, index=data.batch, dim=0, reduce="max")
+        return x, x_rep
+
+
+class simple_target_gen(Model):
```
Aske-Rosted: This class does not seem to be used so please remove it.

examples/pretraining_maskpred_example.py
```diff
@@ -0,0 +1,117 @@
...
+    for batch in dataloader:
+        data = batch
+        break
+
+    dummy_model = simple_model()
+
+    # encoder: Model,
+    # bert_task: UnsupervisedTask,
+    # encoder_out_dim: Optional[int] = None,
+    # need_charge_rep: bool = False,
```
Aske-Rosted: Seems like leftovers from previous iterations, please remove.

examples/pretraining_maskpred_example.py
```diff
@@ -0,0 +1,117 @@
...
+    )
+
+    out = model(data)
+    print(out)
+
+    # for training
+    # model.fit(train_dataloader=dataloader, max_epochs=10, gpus=1)
+
+    # for saving
+    # model.save_pretrained_model('some/path')
```
Aske-Rosted, line 91: Are these comments intended?

src/graphnet/models/pretraining_maskpred.py
```diff
@@ -0,0 +1,317 @@
...
+        return self.backbone
+
+    def save_pretrained_model(self, save_path: str) -> None:
+        """Automates the saving of the pretrained encoder."""
+        model = self.backbone
+
+        run_name = "pretrained_model"
+
+        save_path = os.path.join(save_path, run_name)
+        print("saving to", save_path)
```
Aske-Rosted: Use the available logging rather than print statements.

src/graphnet/models/pretraining_maskpred.py
```diff
@@ -0,0 +1,317 @@
...
+        final_loss: str = "mse",
+        default_hidden_dim: int = 1000,
+        default_nb_linear: int = 5,
+    ) -> None:
+        """Construct the loss calc."""
+        super().__init__()
+        if mask_pred_net is None:
+            print(
+                "no custom net for mask prediction specified; using a standard net"
+            )
```
Aske-Rosted, line 151: Use the available logging rather than print statements.

src/graphnet/models/pretraining_maskpred.py
```diff
@@ -0,0 +1,317 @@
...
+        self.ratio = masked_ratio
+        self.hlc_pos = hlc_pos
+        self.masked_feat = masked_feat
+        self.learned_value = learned_masking_value
+
+        if self.learned_value:
+            print(
+                """warning: can currently only mask adjacent features,
+                  e.g. only (x,y,z) or only (t,q) but not e.g. (x,t,q)"""
+            )
```
Aske-Rosted, line 89: Use the available logging rather than print statements.

examples/pretraining_maskpred_example.py
niklasmei: Yes I commented the training and the saving so that the example file runs even if someone does not have a gpu available and also without actually saving a model that does not actually do anything. I can uncomment or remove if you want me to

niklasmei:
Latest comments should be addressed now. When you can have another look please. Thank you for your support.

Aske-Rosted (approved):
Looks good to me now.

Issue #834, opened 2025-09-29
### [Performance] Explore alternatives to SQLite/Parquet for fast data access


RasmusOrsoe:
**Description**  
Fast access to experiment file formats is essential. Alternatives such as LMDB or Hugging Face datasets may offer better performance in some scenarios. The current Parquet dataset is broken and unlikely to work satisfactorily. Benchmarking is needed to determine tradeoffs.

Potential candidates: LMBD, Hugging Face Dataset, memorymapped .npy arrays (PolarBERT)
 
Some of these formats provide fast random access (like SQLite), while others is read sequentially and therefore require randomization on-write. As a result, the user experience is different. We should consider if/how we can support both regimes.

**Acceptance Criteria**  
- [ ] Benchmark storage footprint and query speeds
- [ ] Assess feasibility of storing data representations, not just raw data  
- [ ] Document benchmarking results and recommendations

Issue #820, opened 2025-08-18
Add `Dataset` and `Writer` modules for the `lmdb` format


RasmusOrsoe:
**Is your feature request related to a problem? Please describe.**
[lmdb](https://en.wikipedia.org/wiki/Lightning_Memory-Mapped_Database) is a memory-mapped database format that provides random access to individual training events faster than SQLite.

The main benefits of `lmdb`: Faster query speeds, lower storage footprint, serialized data allows for an elegant implementation of storing arbitrary `DataRepresentations` (see #781).  

<img width="1391" height="388" alt="Image" src="https://github.com/user-attachments/assets/9403664d-d97c-41fc-bdd7-fbea8670b8c7" />

<img width="590" height="314" alt="Image" src="https://github.com/user-attachments/assets/490cdf6a-78e0-4ad5-8c49-efb32740ba63" />

For larger datasets, and particularly large events, SQLite can become prohibitive as a dataset format.

**Describe the solution you'd like**
* A `LMDBWriter` that outputs events in `lmdb` format. Should accept a serialization method (msgpack, for example). Should optionally accept a DataRepresentation - if given, representations are pre-computed and serialized using `dill` or similar pickle methods (see #781). A field in the database should contain relevant information regarding the serializer, such that the file is a self-contained object that users can read without prior knowledge of the serializer used. 
*  A `LMDBDataset` that is compatible with the lmdb database format. Should automatically check for which serializer was used, so the user doesn't have to guess. Should be able to retrieve pre-computed data representations.

PR #852, opened 2025-12-09
Add `lmdb` as alternative file format

### LINKED ISSUE #781: Graph construction before training
I've found that the calculations for constructing graphs can be a cpu bottleneck in training (especially when not many cpus are available). It would be very useful if the graphs could be constructed ahead of time/before training to streamline dataloading. I suspect this will be even more of an issue for complicated graph encodings.

This could be a step after the data processing step (e.g. i3 files to parquet) and will be dependent on the graph definition.

RasmusOrsoe:
This PR extends the list of supported backends to include `lmdb`, thereby addressing #834 and closing #820.

**The main benefits of LMDB are threefold**: It requires roughly half the space of SQLite, it has significantly faster random access for larger events than SQLite, and it provides a generic way of pre-computing data representations. 

**The downsides are subjective**: No SQL syntax, and accessing large subsets of the dataset in one go is also slow.

### Major Changes

- Adds `LMDBWriter`:
   The writer outputs .lmdb databases, where entries are key-value pairs. The keys are created on the index column (similar to the primary key in sqlite) and the values associated with the entry is _all_ extracted data for the given event. The values are serialized, and several common serialization methods (`json`,  `pickle`, etc) are supported. 

  To make the files self-contained, the databases contain a `__meta__` entry with information on the serialization method used, and utility functions are added that will identify the correct method and use it for deserialization on queries. As a result, the user doesn't need to know the serialization method in order to read the files. Below is an example of a query:
  ```python
  from graphnet.data.utilities.lmdb_utilities import query_database
  
  lmdb_path = "~/merged/merged.lmdb"
  event_no = 1
  
  # The query function automatically detects the serializer used and will deserialize the blop
  # Result is a dict with all table entries for a single event
  # Every entry in the dict is a `table`. E.g. result["truth"] 
  
  result = query_database(lmdb_path, event_no)
  ```
  I profiled the query speeds in a usual data loading scenario vs. the event size, and found the following relationship
  query_vs_n
  The query speed includes both event-level truth and the pulsemap (deserialized), and is repeated 100 times for each event. Real-time computation of representations and computational overhead of establishing connections are not included. From the figure it can be seen that for large events, `lmdb` offers a significant speed-up.  

  Additionally, the `LMDBWriter` accepts a list of `DataRepresentations` - and if provided - the representations are calculated and stored in the file alongside other extracted data. Another meta field is written to the files that contain the config files of the representations, allowing users to re-instantiate the data representation modules used to compute the representations. A utility function for retrieving these is added. As such, this PR also closes #781. An example of retrieving the data representation from the meta data can be seen below

  ```python
  from graphnet.data.utilities.lmdb_utilities import query_database, get_data_representation_from_metadata

  event_no = 0
  lmdb_path = "~/merged/merged.lmdb"
  # Return list of available representations
  query_database(database = lmdb_path, index = event_no )["data_representations"].keys()
  
  # Returns perhaps dict_keys(['KNNGraph', 'GraphDefinition'])
  
  # Get the data representation for 'KNNGraph' from the metadata
  data_representation = get_data_representation_from_metadata(lmdb_path, "KNNGraph")
  ```
  It is assumed that the data representation used is part of the users graphnet installation - i.e. exotic representations that are not yet part of the library will fail. There is no robust way around this.
- Adds `LMDBDataset`
  The dataset is compatible with the .lmdb files and largely identical to the existing `SQLiteDataset`. It supports str-selections and has a "pre-computed" mode, where the user may choose to query pre-computed data representations instead of calculating them in real-time. 

- Adds `SQLiteToLMDBConverter`
  A pre-configured converter that converts existing sqlite databases to lmdb format, similar to our `ParquetToSQLiteConverter`. This converter also accepts a list of data representations, allowing your to export to lmdb alongside pre-computed representations. 

### Minor Changes
- Expanded the test suite to include the lmdb backend for unit testing
- Replaced deprecated converters in the relevant test suite
- Icetray conversion example adjusted to include the lmdb backend (which is now the default format in the example)
- Added utility functions  (graphnet.data.utilities.lmdb_utilities) - for querying events etc
- Minor updates of the documentation. I kept these very small to not bloat the PR further. 
- Updated the `GraphNeTDataModule` to support the lmdb backend

Tagging @astrojarred @giogiopg @Aske-Rosted @sevmag and @pweigel as we've all discussed various aspects of this in the past.


RasmusOrsoe:
@Aske-Rosted  @giogiopg - I've updated the branch with the recent commits from main. I've also fixed the error message and expanded the conversion example to include the option of adding precomputed representations.

astrojarred:
This is really cool, I can make a point to test this with MAGIC data as well.

astrojarred:
Hi @RasmusOrsoe, I can confirm that this works with my preliminary MAGIC data reader class! 
Really nice work.

The following code runs without issue, and the converted data can subsequently be loaded in with your `LMDBDataset`:

```py
reader = MAGICParquetReader(
    apply_cleaning=True,
    ...
)
reader.set_extractors(default_magic_mc_extractors())

converter = DataConverter(
    file_reader=reader,
    save_method=LMDBWriter(index_column="event_id"),
    extractors=default_magic_mc_extractors(),
    outdir=outdir,
    index_column="event_id",
    num_workers=10,
)

converter(input_dir=DATA_FILE)
```

Issue #816, opened 2025-08-07
I3Calorimetry Extractor failing


sevmag:
**Describe the bug**
This a collection of a few bugs that I have encountered while trying out the new extractor from @Aske-Rosted :
https://github.com/graphnet-team/graphnet/blob/e9d594be9b31ebefe60eb94dca07dcedab8c0f3f/src/graphnet/data/extractors/icecube/i3calorimetry.py#L21

## Wrong if statement:

The line below will always yield `False`
https://github.com/graphnet-team/graphnet/blob/e9d594be9b31ebefe60eb94dca07dcedab8c0f3f/src/graphnet/data/extractors/icecube/i3calorimetry.py#L239

Correct is:

```python
if particle.shape != dataclasses.I3Particle.ParticleShape.Dark:
```

## Double-counting the energies of tracks that are part of cascades because of this

The current implementation of the calorimetry extractor will double-count the energies of tracks that are descendants of cascades:

The energy of these particles is captured once by

https://github.com/graphnet-team/graphnet/blob/e9d594be9b31ebefe60eb94dca07dcedab8c0f3f/src/graphnet/data/extractors/icecube/i3calorimetry.py#L122

and once by

https://github.com/graphnet-team/graphnet/blob/e9d594be9b31ebefe60eb94dca07dcedab8c0f3f/src/graphnet/data/extractors/icecube/i3calorimetry.py#L165

This can lead to cases where the e_total is bigger than the primary energy, and thus this assertion fails:

https://github.com/graphnet-team/graphnet/blob/e9d594be9b31ebefe60eb94dca07dcedab8c0f3f/src/graphnet/data/extractors/icecube/i3calorimetry.py#L77

## Double-counting the energies of tracks that are daughters of other tracks

In some cases, some events have a track that later interacts and creates another track. Summing the energy on entrance of both of these tracks like it is being done atm from the function `total_track_energy` can also lead to `e_total > primary energy`

example I3MCTree:

```
34578 MuMinus (-13.2797m, 482.613m, 248.992m) (54.8428deg, 87.634deg) 9597.24ns 340.763GeV 1103.65m
      Other hidden propagation segments
      761729 NuclInt (-47.8447m, -353.949m, -340.704m) (54.8428deg, 87.634deg) 13013.3ns 18.9043GeV 0m
        761773 MuPlus (-47.8447m, -353.949m, -340.704m) (54.8428deg, 87.634deg) 13013.3ns 1.13754GeV 5.26861m
[...]

PR #866, opened 2026-01-27
Revamp of the I3Calorimetry extractor. 

Aske-Rosted:
The changes made to the Calorimetric introduced by #819 were developed using low energy events. While these changes did improve the correctness of the labels they unfortunately were not written in a way that made them suitable for high energy files. 

This PR Rethinks the approach by having adding a hierarchy in the energy counting, that is tracks energies take priority over cascade energies. 

If a track is found the deposit energy inside the detector volume it is recorded and then the particle along with the sub-tree is removed by using the `frame[self.mctree].erase(track.id)`. The cascades are then counted by looking at only the leaf-nodes in the mctree which ensures that we are not double inside the cascades themselves and that we are only cascades terminating inside the detector volume. 

This should ensure that we are eliminating both of the double counting scenarios explained in #816.


sevmag:
Hi @Aske-Rosted, I'm currently reviewing the changes. Just a small question. What do you mean by

`were not written in a way that made them suitable for high energy files.`

Are you talking about the correctness of the labels or performance issues?

src/graphnet/data/extractors/icecube/i3calorimetry.py
```diff
@@ -42,89 +46,49 @@ def __init__(
...
         if self.frame_contains_info(frame):
 
-            primaries = self.get_primaries(
-                frame,
-                self.daughters,
-                self.highest_energy_primary,
+            primary_energy = sum(
+                [
+                    p.energy
+                    for p in self.check_primary_energy(
```
sevmag, line 73: You use the `check_primary_energy` function from the `I3Extractor`. Could you explain in which scenarios the energy of the primary particle is Nan? The remedy in these scenarios is to take the daughters, are we entirely sure that the daughters always have a non NaN energy? Might be worth refining the docstring for this function in

https://github.com/graphnet-team/graphnet/blob/4bcc6074c0d9fa632fc7ddd0fffec8be39df256a/src/graphnet/data/extractors/icecube/i3extractor.py#L112

since I actually don't know why

src/graphnet/data/extractors/icecube/i3calorimetry.py
```diff
@@ -42,89 +46,49 @@ def __init__(
...
+                        frame,
+                        self.get_primaries(
```
sevmag, line 75: `self.get_primaries` is called at three locations. Maybe lets call it once at the beginning of __call__ and save the primary variable and pass it to `total_track_energy` and `total_cascade_energy`

src/graphnet/data/extractors/icecube/i3calorimetry.py
```diff
@@ -193,146 +146,159 @@ def __call__(self, frame: "icetray.I3Frame") -> Dict[str, Any]:
...
-                    track.pos, track.dir
-                )
-                # Get the corresponding energies
+    def total_track_energy(
+        self, frame: "icetray.I3Frame"
+    ) -> Tuple[float, float]:
+        """Get the total energy of track particles on entrance."""
+        e_entrance = 0
+        e_deposited = 0
+        primaries = self.get_primaries(
```
sevmag, line 164: See comment above for `get_primaries` usage

src/graphnet/data/extractors/icecube/i3calorimetry.py
```diff
@@ -193,146 +146,159 @@ def __call__(self, frame: "icetray.I3Frame") -> Dict[str, Any]:
...
+            ), "Energy deposited is greater than primary energy"
+        return e_entrance, e_deposited
+
+    def total_cascade_energy(
+        self,
+        frame: "icetray.I3Frame",
+    ) -> float:
+        """Get the total energy of cascade particles on entrance."""
+        particles = deque(
+            self.get_primaries(
```
sevmag, line 260: See comment above for `get_primaries` usage

src/graphnet/data/extractors/icecube/i3calorimetry.py
```diff
@@ -193,146 +146,159 @@ def __call__(self, frame: "icetray.I3Frame") -> Dict[str, Any]:
...
+            e_deposited += e0 - e1
+            e_entrance += e0
+            # get descendant ids
+            # erase particle and children from mctree
+            frame[self.mctree].erase(track.id)
+
+        # Sanity check ensuring no double counting
+        if self.daughters:
+            assert e_entrance <= sum(
+                [p.energy for p in primaries]
```
sevmag, line 247: The sum of primary energies has already been calculated on line 70. Reuse this value for consistency.

src/graphnet/data/extractors/icecube/i3calorimetry.py
```diff
@@ -193,146 +146,159 @@ def __call__(self, frame: "icetray.I3Frame") -> Dict[str, Any]:
...
+            ), "Energy on entrance is greater than primary energy"
+            assert e_deposited <= sum(
```
sevmag, line 249: The sum of primary energies has already been calculated on line 70. Reuse this value for consistency.

src/graphnet/data/extractors/icecube/i3calorimetry.py
```diff
@@ -193,146 +146,159 @@ def __call__(self, frame: "icetray.I3Frame") -> Dict[str, Any]:
...
-                                        frame[self.mctree], particle
-                                    ),
-                                    track_lookup,
-                                ),
-                            )
+                        MMCTrackList_filtered.append(track)
+                except RuntimeError as e:
+                    if "particle not found" in str(e):
+                        # log warning with event header
+                        self.warning(
```
sevmag, line 182: How can it be that the primary particle is not found anymore? Are we sure it is safe to keep this as a warning?

src/graphnet/data/extractors/icecube/i3calorimetry.py
```diff
@@ -193,146 +146,159 @@ def __call__(self, frame: "icetray.I3Frame") -> Dict[str, Any]:
...
+                continue
+
+            # Find distance to entrance and exit from sampling volume
+            intersections = self.hull.surface.intersection(
+                track.pos, track.dir
+            )
 
-        return e_cascade, e_dep_track, e_ent_track
+            # Check if the track actually enters the volume
+            if not (
```
sevmag, line 211: Maybe add a comment on why this check also works for starting events. (Then the `intersection.first` is negative, correct?)

src/graphnet/data/extractors/icecube/i3calorimetry.py
```diff
@@ -193,146 +146,159 @@ def __call__(self, frame: "icetray.I3Frame") -> Dict[str, Any]:
...
 
-    def is_in_hull(self, particle: "dataclasses.I3Particle") -> bool:
-        """Check if a particle is in the hull."""
-        pos = np.array(particle.pos)
-        direc = np.array([particle.dir.x, particle.dir.y, particle.dir.z])
-        length = particle.length if particle.length is not None else 0
-        pos = pos + direc * length
+            except RuntimeError as e:
+                if (
+                    "sum of losses is smaller than "
```
sevmag, line 224: I'm still not sure what the reason for this MuonGun error is, but I think we should not just ignore it and throw a warning, but rather set the energies to NaN to make sure that we are not just using a corrupt ground truth that might not resemble reality at all. What do you think @Aske-Rosted ?

sevmag:
I have some concerns regarding whether the current track handling produces the intended outcome. 🧐

## Hypothetical Scenario 🧪
Consider a NuTau CC interaction that creates a tau. The tau enters the detector volume and decays into a muon mid-detector, which then exits the volume.

This results in a `track_list` like this:
```python
track_list = [tau, mu]
```
In this case, the `mu` is a child in the `tau`'s subtree. 

When iterating through the list [as seen here](https://github.com/Aske-Rosted/graphnet/blob/fe26e91d8e11abe7b46827c37a6e1fbc51281681/src/graphnet/data/extractors/icecube/i3calorimetry.py#L196), the result seems highly dependent on the **ordering** of `track_list`:

---

### Scenario 1: The tau is processed first 📍
In this case, we only iterate over the tau. Because the muon is in the tau's subtree, it is erased after the tau is processed.
* **`e_deposited`**: Only includes energy from the tau (entry to decay). It **misses** the energy deposited by the muon. ❌
* **`e_entrance`**: Correctly identifies the tau energy upon entering the detector. ✅

### Scenario 2: The muon is processed first 📍
Here, we iterate over the muon first, then the tau. 
* **`e_deposited`**: Correct; it sums the energy of both the muon and the tau. ✅
* **`e_entrance`**: Incorrect; the entrance energy of *both* particles is counted, even though only the tau should be considered the primary entering particle. ❌

---

Since the behavior changes based on list order, we might need to implement a more robust check for parent-child relationships before processing. What do you think?

src/graphnet/data/extractors/icecube/i3calorimetry.py
```diff
@@ -193,146 +146,159 @@ def __call__(self, frame: "icetray.I3Frame") -> Dict[str, Any]:
...
+    ) -> float:
+        """Get the total energy of cascade particles on entrance."""
+        particles = deque(
+            self.get_primaries(
+                frame, self.daughters, self.highest_energy_primary
+            )
+        )
 
-        return self.hull.point_in_hull(pos)
+        if len(particles) == 0:
```
sevmag, line 265: This scenario should never happen, should it? Maybe we should throw an error here

sevmag (changes requested):
I really like the direction this is going, and I think we can increase its speed quite a bit by leveraging the tree-like structure. However, I do still have some concerns about the handling of the track particles (see my comment). Let me know if there are any questions!

src/graphnet/data/extractors/icecube/i3calorimetry.py
Aske-Rosted: In some rare instances the energy of the particle is just not set, this is something that happens upstream in the simulation process, I don't exactly know why. It has been sometime since I have looked at it but I believe that there will be a daughter of the same type as the parent particle with an actual energy defined.

Aske-Rosted:
> I have some concerns regarding whether the current track handling produces the intended outcome. 🧐
> ## Hypothetical Scenario 🧪
> 
> Consider a NuTau CC interaction that creates a tau. The tau enters the detector volume and decays into a muon mid-detector, which then exits the volume.
> 
> This results in a `track_list` like this:
> 
> ```python
> track_list = [tau, mu]
> ```
> 
> In this case, the `mu` is a child in the `tau`'s subtree.
> 
> When iterating through the list [as seen here](https://github.com/Aske-Rosted/graphnet/blob/fe26e91d8e11abe7b46827c37a6e1fbc51281681/src/graphnet/data/extractors/icecube/i3calorimetry.py#L196), the result seems highly dependent on the **ordering** of `track_list`:
> ### Scenario 1: The tau is processed first 📍
> 
> In this case, we only iterate over the tau. Because the muon is in the tau's subtree, it is erased after the tau is processed.
> 
>     * **`e_deposited`**: Only includes energy from the tau (entry to decay). It **misses** the energy deposited by the muon. ❌
> 
>     * **`e_entrance`**: Correctly identifies the tau energy upon entering the detector. ✅
> 
> 
> ### Scenario 2: The muon is processed first 📍
> 
> Here, we iterate over the muon first, then the tau.
> 
>     * **`e_deposited`**: Correct; it sums the energy of both the muon and the tau. ✅
> 
>     * **`e_entrance`**: Incorrect; the entrance energy of _both_ particles is counted, even though only the tau should be considered the primary entering particle. ❌
> 
> 
> Since the behavior changes based on list order, we might need to implement a more robust check for parent-child relationships before processing. What do you think?

The ordering of MCTree is not random if the muon is a product of the tau then it will be in the subtree of the tau I.e. the tau is _always_ processed first. The Muon.Track.Harvest iterates over the MCTree so this ordering should be conserved. https://github.com/icecube/icetray/blob/7d7982c84148d7e66541e0f953be01f80b061859/ddddr/private/ddddr/MuonGunTrack.cxx#L136 

I think you might be correct in terms of stopping tracks I think then we are missing the energy of the resulting cascade, however when fixing this we also have to consider the effects of _dark_ (non-light producing particles) produced in the cascade which might _remove_ some energy from the deposit.

src/graphnet/data/extractors/icecube/i3calorimetry.py
Aske-Rosted: I think it is okay to work with what we have this also only happens very rarely. I also believe that this should be fixed in newer versions of IceTray https://github.com/icecube/icetray/pull/3052

src/graphnet/data/extractors/icecube/i3calorimetry.py
Aske-Rosted: This could happen for a single through-going track event since we have removed the track particle and all the sub-particles from the mctree.

src/graphnet/data/extractors/icecube/i3calorimetry.py
```diff
@@ -193,146 +146,159 @@ def __call__(self, frame: "icetray.I3Frame") -> Dict[str, Any]:
...
+                    continue
+                else:
+                    raise
+
+            # Accumulate
+            e_deposited += e0 - e1
+            e_entrance += e0
+            # get descendant ids
+            # erase particle and children from mctree
+            frame[self.mctree].erase(track.id)
```
Aske-Rosted: The below alteration I believe should fix the missing energies for stopping particles. 

```suggestion
            if (e1 == 0) or (intersections.second < particle.length):
                frame[self.mctree].erase(track.id)
```

src/graphnet/data/extractors/icecube/i3calorimetry.py
sevmag: Maybe let's just be sure to check that all of the energies of the daughters that are set as the new primaries, after the line below, do have a non-nan energy, and otherwise, raise an error.

https://github.com/graphnet-team/graphnet/blob/4bcc6074c0d9fa632fc7ddd0fffec8be39df256a/src/graphnet/data/extractors/icecube/i3extractor.py#L163

src/graphnet/data/extractors/icecube/i3calorimetry.py
sevmag: ok perfect! I still think we shouldn't just ignore that track and move on to the next. In my experience, that can lead to very strange energy labels, especially when one increases the padding of the hull to large labels. I think we should just put the energy labels to NaN, signaling that we cannot give the correct label for these events. When it is super rare, this won't affect your sample statistics anyway, and we will be on the safe side. What do you think?

src/graphnet/data/extractors/icecube/i3calorimetry.py
```diff
@@ -193,146 +146,159 @@ def __call__(self, frame: "icetray.I3Frame") -> Dict[str, Any]:
...
-            # If the particle is a cascade in the hull, we add its energy.
-            elif particle.is_cascade:
-                if self.cascade_deposited_only:
-                    # Check wether the cascade is made up of smaller segments
-                    # in this case the shape is dark and we want to count
-                    # the energy of its daughters.
                     if (
-                        particle.shape
-                        != dataclasses.I3Particle.ParticleShape.Dark
+                        frame[self.mctree].get_primary(track.GetI3Particle())
```
sevmag: Actually, this check is a problem. 

The `primaries` variable that we compare against here does not necessarily need to be at the top level of the tree. This is because our `get_primaries` sometimes goes down the tree to find the first in-ice neutrino. The `get_primary` method of the icetray MCTree only gives you the primary of the very top of the tree. This is problematic since we check whether this particle is in our `primaries` list. This means we will ignore these tracks, which is a big problem.

sevmag: we should stick to the code from main

https://github.com/graphnet-team/graphnet/blob/368927d7d12b488afd1b1cde31491336ab2239c6/src/graphnet/data/extractors/icecube/i3calorimetry.py#L92-L103

it addresses this problem

sevmag:
> The ordering of MCTree is not random if the muon is a product of the tau then it will be in the subtree of the tau I.e. the tau is _always_ processed first. The Muon.Track.Harvest iterates over the MCTree so this ordering should be conserved. https://github.com/icecube/icetray/blob/7d7982c84148d7e66541e0f953be01f80b061859/ddddr/private/ddddr/MuonGunTrack.cxx#L136
> 
> I think you might be correct in terms of stopping tracks I think then we are missing the energy of the resulting cascade, however when fixing this we also have to consider the effects of _dark_ (non-light producing particles) produced in the cascade which might _remove_ some energy from the deposit.

But as I showed in my issue before, even if the ordering is always the tau first. The labels will still be incorrect. I think we can use this approach of `e_on_entrance`, but for `e_deposited` we cannot take the shortcuts of deleting the subtree after a track.

src/graphnet/data/extractors/icecube/i3calorimetry.py
sevmag: Particles are not simulated once they leave the detector (+ some margin), so these tracks probably won't have children anyway, most of the time. So in that case, deleting the subtree is probably unnecessary anyway, and we would probably be safer just to never do it.

src/graphnet/data/extractors/icecube/i3calorimetry.py
Aske-Rosted: This approach is not feasible on high energy events. The is_in_subtree call is simply too expensive.

Aske-Rosted: I checked my log files and did not find a single instance of this error being raised.

sevmag: I am a bit confused. What error do you mean? The scenario I am describing would not throw an error. It would silently filter out tracks that should not be filtered out.

Aske-Rosted: Ah my bad I was a bit too quick when i read the logic. If the primary isn't found then no warning is raised you are correct. It still does not change the fact that we cannot afford to run the is_in_subtree. One could loop over the parents and save all the particles with type primary, but i fear that this too might be too expensive of an operation...

Aske-Rosted: I can try and make some alterations while testing on my highest energy CORSIKA files and see if there is a solution that works.

sevmag: Perfect! Just to make sure that we are talking about the same thing:

The elements of the `primaries` list, which is the output of our `get_primaries` function are one of the three
1. Top-level-tree primaries like CORSIKA primaries, in-ice neutrinos
2. daughters of top-level-tree primaries that have energy equal to NaN
3. Not-top-level-tree in-ice neutrinos that result from neutrinos interacting outside of the ice

The bug, which I described, will only happen for tracks that are descendants of primaries of category 2 or 3. So running on CORSIKA will only tell you whether the algorithm is fast enough; it will not tell you whether it covers these edge cases correctly.

Aske-Rosted:
> The changes made to the Calorimetric introduced by #819 were developed using low energy events. While these changes did improve the correctness of the labels they unfortunately were not written in a way that made them suitable for high energy files.
> 
> This PR Rethinks the approach by having adding a hierarchy in the energy counting, that is tracks energies take priority over cascade energies.
> 
> If a track is found the deposit energy inside the detector volume it is recorded and then the particle along with the sub-tree is removed by using the `frame[self.mctree].erase(track.id)`. The cascades are then counted by looking at only the leaf-nodes in the mctree which ensures that we are not double inside the cascades themselves and that we are only cascades terminating inside the detector volume.
> 
> This should ensure that we are eliminating both of the double counting scenarios explained in #816.

I order to facilitate the comments by @sevmag Major rethinking of the approach was necessary. The behavior of the I3Calorimetry extractor have been changed quite a bit. I have listed the changes below.

### changes ###
- [REMOVED] input argument `daughters`.
The previous method used the input argument daughters to determine whether to only consider energies from a neutrino primary in the case of nugen files. CORSIKA would still consider all the deposits by the shower.
- [ADDED] `entrance_energy`
Previously we were trying to produce two energy metrics. The entrance energy metric where if we can see a track, we assume that we can reconstruct the energy of of that track. Therefore we wanted to remove downstream of the track as we had already included the full energy of the track. Concurrently we were also trying to construct the actual energy deposited inside the detector, in this case we would not like to remove the daughters of the track as they might produce a cascade inside the detector. Therefore splitting it up so you are either considering the energy on entrance or the deposited energy seems to make more sense. 
- New separation of background (non neutrino primary components) and _target_ (neutrino primar(y/ies)) 
- `highest_energy_primary` whether we consider only the highest energy neutrino primary or all neutrinos as the _target_ 
- This means that the outputs have also been updated. we are mainly considering _target_, _background_ and total. 
- [ADDED] `filter_track_list` method to only get tracks that currently exist in the _target/background_ tree. 

### Refactoring ###
`total_track_energy` and `total_cascade_energy` methods now takes mctree as an argument instead of recreating it inside the methods.

### Validations ###

- Split trees contain the same total particle count as the original
- Deposited energy doesn't exceed primary energy (for both target and background)
- Cascade energy doesn't exceed primary energy

### Additional file edits ### 
As a result of the changes above I had to change the `I3Extractor` methods to take the an instantiated mctree as an argument rather than a frame. Which also leads to some consequence changes in the `I3HighestEparticleExtractor` which also relies on the same methods in the `I3Extractor`

### WARNING ### 
Since the input argument to I3Calorimetry has been changes this does mean that we are not backwards compatible.

src/graphnet/data/extractors/icecube/i3calorimetry.py
```diff
@@ -42,297 +46,328 @@ def __init__(
...
-                    \nCascade energy: {}\
-                    {}".format(
-                    e_total,
-                    primary_energy,
-                    e_ent_track,
-                    e_cascade,
-                    frame["I3EventHeader"],
+            if not (
+                e_total_target <= (target_primaries_energy * (1 + 1e-6))
+                or (e_total_target - target_primaries_energy < 0.5)
```
sevmag, line 158: The second case seems a bit odd. What are we checking for with `e_total_target - target_primaries_energy < 0.5`

src/graphnet/data/extractors/icecube/i3calorimetry.py
```diff
@@ -42,297 +46,328 @@ def __init__(
...
-                    \nTrack energy: {}\
-                    \nCascade energy: {}\
-                    {}".format(
-                    e_total,
-                    e_ent_track,
-                    e_cascade,
-                    frame["I3EventHeader"],
+            if not (
+                e_total_bkg <= (bkg_primaries_energy * (1 + 1e-6))
+                or (e_total_bkg - bkg_primaries_energy < 0.5)
```
sevmag, line 171: See command above

src/graphnet/data/extractors/icecube/i3calorimetry.py
```diff
@@ -42,297 +46,328 @@ def __init__(
...
+                    f"Background primary energy: {bkg_primaries_energy}\n"
+                    f"Track deposited background energy: {e_track_bkg}\n"
+                    f"Cascade deposited background energy: {e_cascade_bkg}\n"
+                    f"{frame['I3EventHeader']}"
                 )
-            fraction_primary = e_total / primary_energy
 
-            cascade_fraction = None
+            fraction_target_total = (
+                e_total_target / e_total if e_total > 0 else 0.0
```
sevmag, line 183: Is `e_total = 0` ever possible? This would mean nothing ever enters the detector hull, which might not be intended, right? Maybe we should check for this and throw an error?

src/graphnet/data/extractors/icecube/i3calorimetry.py
```diff
@@ -42,297 +46,328 @@ def __init__(
...
+            )
+            target_cascade_fraction = (
+                e_cascade_target / e_total_target
+                if e_total_target > 0
+                else 0.0
+            )
+
             if e_total > 0:
```
sevmag, line 191: See command above

src/graphnet/data/extractors/icecube/i3calorimetry.py
```diff
@@ -42,297 +46,328 @@ def __init__(
...
-                cascade_fraction = e_cascade / e_total
+                cascade_fraction_tot = (
+                    e_cascade_target + e_cascade_bkg
+                ) / e_total
 
+            if target_primaries_energy > 0:
```
sevmag, line 196: `target_primaries_energy > 0` should always be satisfied, isn't it?

src/graphnet/data/extractors/icecube/i3calorimetry.py
```diff
@@ -42,297 +46,328 @@ def __init__(
...
+            )
+
+        track_list = deque(MuonGun.Track.harvest(mctree, mmc_track_list))
+
+        while len(track_list) > 0:
+            track = track_list.popleft()
+
+            try:
+                particle = mctree.get_particle(track.id)
+            except RuntimeError:
```
sevmag, line 263: From one of your older comments, this cannot happen, correct? Since they are always processed in order? Might potentially be dangerous if one first iterates over the lowest level particles in the tree and then go further up. This can lead to double-counting energies on entry

src/graphnet/data/extractors/icecube/i3calorimetry.py
```diff
@@ -42,297 +46,328 @@ def __init__(
...
-                            ),
-                            track_lookup,
-                        ),
-                    )
-                )
+                    raise
 
-        return e_cascade, e_dep_track, e_ent_track
+            # Accumulate
+            if entrance_energy:
```
sevmag, line 306: merge the two if blocks (this one and the one 4 lines below), since they have the same statement

sevmag (commented):
This is going in a really good direction! I like the tree splitting that makes things way easier! One general suggestion is to rename all `*target*` variables in `*neutrino*` since target seems to mean neutrino here. It would help readability. I left a few comments, but once we go over them, we should be good. 

I think we should spend some time on unit tests here. We should cherry-pick some MCTrees and track lists from MC that cover some of the edge cases that both of us discovered while working on this. We should make sure that the extractor reads the right things from them. There are so many twists in this extraction, and I fear that if anyone touches this in the future, they might not be aware of them

List of Edge cases:

- [ ] muon neutrino-induced contained muon track (interaction vertex inside the hull and muon decaying inside the hull)
- [ ] stopping muon track (entering from outside and stopping inside the hull)
- [ ] through-going muon track (entering from the outside and leaving the hull again)
- [ ] starting muon track (neutrino interaction vertex inside the hull and muon leaving)
- [ ] primary with NaN energy (at the top level tree position, a particle has NaN energy and the daughters have non-NaN energy)
- [ ] first in-ice-neutrino that is not at the top tree level (the primary is a neutrino but interacts outside of the hull with NC to create another neutrino, which again interacts outside of the detector N times in the same scheme and finally then interacts inside the detector to create a muon track leaving the detector)
- [ ] tau neutrino producing a tau inside the detector, which then decays to a mu track in the detector, which leaves the detector
- [ ] no descendants of the neutrino reach the hull (neutrino interacts outside the hull, potentially in a cascade which never reaches the detector)
- [ ] fully contained double bang potentially (tau neutrino interacts to a tau and which decays into a cascade again)

lmk if there are some missing. Happy to help on the unit tests

src/graphnet/data/extractors/icecube/i3calorimetry.py
Aske-Rosted: mass -> energy conversion, (based on Q-value) scenarios for fully contained events.

src/graphnet/data/extractors/icecube/i3calorimetry.py
Aske-Rosted: see answer above.

src/graphnet/data/extractors/icecube/i3calorimetry.py
Aske-Rosted: I am not 100% whether or not pure-noise (not coincidence) events can occur. I think in principle the simulation framework could allow it though I am not sure if we are actually simulate and keep any of those events.

src/graphnet/data/extractors/icecube/i3calorimetry.py
Aske-Rosted: We do actually allow for it at https://github.com/graphnet-team/graphnet/pull/866/changes/BASE..533300bcc347e2fe1cac70df1d1a7a053ae31c26#diff-e0c9dd4088d378e779b82e6d3b4faa564033b1d2029411848deabad5003aa544R246
with a warning

src/graphnet/data/extractors/icecube/i3calorimetry.py
Aske-Rosted: No this can happen and should happen, as we are processing top-down. If a particle (and it's daughters) are removed using mctree.erase(track.id) then it is not automatically removed from the track_list that we are iterating over, but this check against the mctree ensures that we are skipping them.

Aske-Rosted:
> This is going in a really good direction! I like the tree splitting that makes things way easier! One general suggestion is to rename all `*target*` variables in `*neutrino*` since target seems to mean neutrino here. It would help readability. I left a few comments, but once we go over them, we should be good.
> 
> I think we should spend some time on unit tests here. We should cherry-pick some MCTrees and track lists from MC that cover some of the edge cases that both of us discovered while working on this. We should make sure that the extractor reads the right things from them. There are so many twists in this extraction, and I fear that if anyone touches this in the future, they might not be aware of them
> 
> List of Edge cases:
> 
>     * [ ]  contained track
> 
>     * [ ]  starting track
> 
>     * [ ]  primary with NaN energy
> 
>     * [ ]  first in-ice-neutrino that is not at the top tree level
> 
>     * [ ]  tau to mu track decays in the detector
> 
>     * [ ]  no decendants of the neutrino reach the hull
> 
> 
> lmk if there are some missing. Happy to help on the unit tests

In order to get a couple of example files we would have to obtain permission from the IceCube collaboration. The MC-simulation files are not publicly available.

src/graphnet/data/extractors/icecube/i3calorimetry.py
sevmag: True, you're right

sevmag:
> In order to get a couple of example files we would have to obtain permission from the IceCube collaboration. The MC-simulation files are not publicly available.

Theoretically, we could also just artificially create them with the icetray software, right?

Aske-Rosted:
> > In order to get a couple of example files we would have to obtain permission from the IceCube collaboration. The MC-simulation files are not publicly available.
> 
> Theoretically, we could also just artificially create them with the icetray software, right?

That's the same issue IceTray is IceCube intele

> > In order to get a couple of example files we would have to obtain permission from the IceCube collaboration. The MC-simulation files are not publicly available.
> 
> Theoretically, we could also just artificially create them with the icetray software, right?

If you are using the private icetray repository to generate the simulation files then they are IceCube property.

PR #875, opened 2026-03-26
Add IceTop support to I3TruthExtractor

### LINKED ISSUE #874: Add IceTop support to I3TruthExtractor
I3TruthExtractor only supports in-ice data, making it impossible to extract truth-level information from IceTop frames without a separate extractor.

Add an ice_top boolean toggle to I3TruthExtractor that, when set to True, switches to IceTop-specific frame keys (e.g. "MCPrimary") and skips in-ice-only computations such as dbang_decay_length.

**Describe alternatives you've considered**
I3IceTopTruthExtractor class, but this would duplicate significant amounts of logic that is shared between in-ice and IceTop truth extraction.

**Additional context**
The changes affect I3TruthExtractor, frame_is_montecarlo, and frame_is_noise, with the toggle defaulting to False to preserve all existing in-ice behavior.

If the existing I3TruthExtractor is used on IceTop data, all extracted truth values are returned as padding values in addition to misidentifying the simulated IceTop data as real data as shown below:

Image

RajwolChapagain:
Adds an ice_top boolean toggle to I3TruthExtractor that enables correct extraction of truth-level information from IceTop frames. Closes #874 .

**Changes**
Added ice_top parameter to I3TruthExtractor.__init__
Updated allowed sub-event streams to include ice_top when ice_top=True
Skips _extract_dbang_decay_length and _get_primary_track_energy_and_inelasticity for IceTop frames
Updated _get_primary_particle_interaction_type_and_elasticity to explicitly use "MCPrimary" for IceTop frames
Updated _find_data_type to correctly identify IceTop MC as "corsika"
Updated frame_is_montecarlo and frame_is_noise to use IceTop-specific frame keys when ice_top=True

**Testing**
Verified that existing in-ice behavior is fully preserved when ice_top=False (default)
Verified that IceTop frames are correctly identified as "corsika" MC and return non-padding truth values when ice_top=True as shown in the screenshot:
image

### DIFF
--- a/src/graphnet/data/extractors/icecube/i3truthextractor.py
+++ b/src/graphnet/data/extractors/icecube/i3truthextractor.py
@@ -175,6 +175,7 @@ class I3TruthExtractor(I3Extractor):
         if frame["I3EventHeader"].sub_event_stream not in [
             "InIceSplit",
             "Final",
+            "ice_top",
         ]:
             return output
 
@@ -289,6 +290,9 @@ class I3TruthExtractor(I3Extractor):
     def _extract_dbang_decay_length(
         self, frame: "icetray.I3Frame", padding_value: float = -1
     ) -> float:
+        if self._mctree not in frame:
+            return padding_value
+            
         mctree = frame[self._mctree]
         try:
             p_true = mctree.primaries[0]
@@ -418,7 +422,10 @@ class I3TruthExtractor(I3Extractor):
             try:
                 MCInIcePrimary = frame["MCInIcePrimary"]
             except KeyError:
-                MCInIcePrimary = frame[self._mctree][0]
+                try:
+                    MCInIcePrimary = frame[self._mctree][0]
+                except KeyError:
+                    MCInIcePrimary = frame["MCPrimary"]  # IceTop CORSIKA primary
             if (
                 MCInIcePrimary.energy != MCInIcePrimary.energy
             ):  # This is a nan check. Only happens for some muons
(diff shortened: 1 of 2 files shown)


src/graphnet/data/extractors/icecube/i3truthextractor.py
```diff
@@ -172,10 +175,11 @@ def __call__(
         # for example I3RecoPulseSeriesMap,  etc.
         # At low levels i3 files contain several other P frame splits
         # (e.g NullSplit). We remove those here.
-        if frame["I3EventHeader"].sub_event_stream not in [
-            "InIceSplit",
-            "Final",
-        ]:
+        allowed_streams = ["InIceSplit", "Final"]
+        if self._ice_top:
+            allowed_streams.append("ice_top")
```
Aske-Rosted: Are InIceSplit and Final valid streams for the IceTop dataframes?

Aske-Rosted (commented):
Hi,

The issue causing the test to fail should now be resolved, could you try and merge the updates into this branch? 

I also had one comment on the implementation, but otherwise I think everything looks fine.

src/graphnet/data/extractors/icecube/i3truthextractor.py
RajwolChapagain: I did a little digging and found that that they're actually not valid for IceTop. Thank you for catching that! I've fixed it in the latest push.

Aske-Rosted (approved):
Looks good to me now.

PR #876, opened 2026-03-30
Lmdb test fix

Aske-Rosted:
The current checks are running into an error, the inital merge of #852 did not seem to have the test fail but all subsequent PR's have the test failing. (I also reran the test for #852 and the test seems to fail, it is beyond me how it passed initially...) 

The error comes from opening multiple connections of the same LMDB database which does not seem to be allowed. This runs into an error at https://github.com/graphnet-team/graphnet/blob/847775711f93b8e9b6f7a40db72886b970462102/src/graphnet/data/utilities/lmdb_utilities.py#L62 which is caught and `None` is returned for https://github.com/graphnet-team/graphnet/blob/847775711f93b8e9b6f7a40db72886b970462102/src/graphnet/data/utilities/lmdb_utilities.py#L50 . My impression is that https://github.com/graphnet-team/graphnet/blob/847775711f93b8e9b6f7a40db72886b970462102/src/graphnet/data/dataset/lmdb/lmdb_dataset.py#L340 is properly handled by the dataloaders and why this has not shown up as an issue in real use. 

The fix I propose is to add a public close() (which just calls the private _close_connection(), this is more convention than anything else) and then call the close() in the test before testing the precomputed dataset version. 

Another minor adjustment i made is move the close_connection() call to the end of the post_init() call, this is just a small refactoring.

### DIFF
--- a/src/graphnet/data/dataset/lmdb/lmdb_dataset.py
+++ b/src/graphnet/data/dataset/lmdb/lmdb_dataset.py
@@ -169,9 +169,9 @@ class LMDBDataset(Dataset):
         if self._pre_computed_representation is None:
             # Only check for missing columns if using raw tables
             self._remove_missing_columns()
-        self._close_connection()
         if self._pre_computed_representation is not None:
             self._identify_missing_truth_labels()
+        self._close_connection()
 
     def _identify_missing_truth_labels(self) -> None:
         """Identify missing truth labels in the pre-computed representation."""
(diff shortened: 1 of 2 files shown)


RasmusOrsoe (approved):
Thanks for looking into this @Aske-Rosted! Not sure what might have caused this to begin to fail. Perhaps it's an update to the lmdb library itself? In any case, let's hope this is the end of it.

PR #878, opened 2026-04-24
fix: quote pip install extras to fix zsh globbing error

### LINKED ISSUE #869: Issue in the installation documentation
### Description
The installation commands currently generated by the `quick-start` widget use the unquoted format: `pip3 install -e .[extras]`. (also in the current fix proposed in #862 )

On macOS (where **Zsh** is the default shell) and other systems using Zsh, square brackets are interpreted as globbing/pattern-matching characters. This causes the installation to fail immediately with the following error:
`zsh: no matches found: .[develop]`

### Proposed Fix
The installation target needs to be wrapped in double quotes. This makes the command "shell-safe," ensuring it works correctly across Zsh, Bash, and PowerShell.

#### Changes required in `quick-start.html`:

**1. Update the all appearances of the .[extras] and wrap in  quotes:**

https://github.com/graphnet-team/graphnet/blob/737646e6061a87c3a64715c4dc49526564184224/docs/source/installation/quick-start.html#L152

fix:

`    $("#command pre").text("pip3 install -e \".[develop]\"");`

https://github.com/graphnet-team/graphnet/blob/737646e6061a87c3a64715c4dc49526564184224/docs/source/installation/quick-start.html#L157

fix:

`    var installLine = "pip3 install -e \".[" + extrasPrefix + "]\"";`

These changes are still valid for the updated code in #862 at the time of writing

sevmag:
## Summary

Fixes #869.

- Wrap `.[extras]` in double quotes in `quick-start.html` so the generated `pip install` commands are shell-safe on Zsh (macOS default), where unquoted square brackets are treated as glob patterns causing `zsh: no matches found: .[develop]`
- Include the `develop` extras in all installation configurations (both w/o PyTorch and with PyTorch) so dev tools (`black`, `pytest`, `mypy`, etc.) are always available

### DIFF
--- a/docs/source/installation/quick-start.html
+++ b/docs/source/installation/quick-start.html
@@ -170,14 +170,14 @@
         "# Clone and Install GraphNeT",
         gitCloneCmd,
         "cd graphnet",
-        "pip3 install -e .[develop]"
+        "pip3 install -e \".[develop]\""
       ].join("\n"));
       return;
     }
 
     var torchVersion = torchWheelMap[torch];
     var extrasPrefix = torchExtrasMap[torch];
-    var installLine = "pip3 install -e .[" + extrasPrefix + "]";
+    var installLine = "pip3 install -e \".[" + extrasPrefix + "]\"";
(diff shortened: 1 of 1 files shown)


Aske-Rosted (approved):
Seems reasonable to me.

PR #881, opened 2026-05-03
chore: bump pre-commit hook versions

sevmag:
## NOTE:
The large amount of changed files is only a relic from running the new linting on the whole repo the only real change is in [.pre-commit-config.yaml](https://github.com/graphnet-team/graphnet/pull/881/changes#diff-63a9c44a44acf85fea213a857769990937107cf072831e1a26808cfde9d096b9)

## Summary

Updates pinned versions in `.pre-commit-config.yaml` to current upstream tags. The previous pins were ~1 year old and the older `flake8 7.1.1` / `black 24.10.0` releases trigger a deprecated `bdist_wheel.universal` codepath that fails to install cleanly on certain pip + setuptools combinations (we saw `[Errno 17] File exists: 'build/bdist.linux-x86_64/wheel/<pkg>.dist-info'` during fresh hook env installs). Newer releases ship prebuilt wheels and avoid the legacy install path.

Changes:
- black 24.10.0 → 26.3.1
- flake8 7.1.1 → 7.3.0
- mypy v1.13.0 → v1.20.2
- pre-commit-hooks v4.6.0 → v6.0.0
- docformatter intentionally **not** bumped — v1.7.8 has a `NameError: tomllib is not defined` regression on Python 3.10 (uses 3.11+ stdlib unconditionally), so v1.7.7 is kept.

## Test plan

- [x] `pre-commit run --all-files` (or on changed files) passes locally with the new pins
- [x] CI lint job passes
- [x] No new lint findings introduced by the bumped versions on existing code

### DIFF
--- a/src/graphnet/data/dataset/parquet/parquet_dataset.py
+++ b/src/graphnet/data/dataset/parquet/parquet_dataset.py
@@ -258,14 +258,20 @@ class ParquetDataset(Dataset):
         else:
             file_idx = [bisect_right(self._chunk_cumsum, sequential_index)]
 
-        file_indices = [self._indices[idx] for idx in file_idx]
+        raw_file_indices = [self._indices[idx] for idx in file_idx]
+        file_indices: List[int] = []
+        for entry in raw_file_indices:
+            if isinstance(entry, list):
+                file_indices.extend(entry)
+            else:
+                file_indices.append(int(entry))
 
         arrays = []
-        for file_idx in file_indices:
+        for fidx in file_indices:
             array = self._query_table(
                 table=table,
                 columns=columns,
-                file_idx=file_idx,
+                file_idx=fidx,
                 sequential_index=sequential_index,
                 selection=selection,
             )
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -75,7 +75,7 @@ jobs:
           sudo apt update --fix-missing --yes
           sudo apt upgrade --yes
           sudo apt-get install --yes git
-          sudo apt-get clean 
+          sudo apt-get clean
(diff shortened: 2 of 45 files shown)


src/graphnet/data/dataset/parquet/parquet_dataset.py
```diff
@@ -258,14 +258,20 @@ def query_table(  # type: ignore
         else:
             file_idx = [bisect_right(self._chunk_cumsum, sequential_index)]
 
-        file_indices = [self._indices[idx] for idx in file_idx]
```
sevmag, line 261: Lets look at this to make sure it is doing the intended behavior

src/graphnet/data/dataset/parquet/parquet_dataset.py
```diff
@@ -258,14 +258,20 @@ def query_table(  # type: ignore
...
+        raw_file_indices = [self._indices[idx] for idx in file_idx]
+        file_indices: List[int] = []
+        for entry in raw_file_indices:
+            if isinstance(entry, list):
```
Aske-Rosted, line 264: since self._query_table always expects a single int for the [file_idx ](https://github.com/graphnet-team/graphnet/pull/881/changes#diff-237eabfc704143831d5a2bb22c25dcc2d91e4a58a1ec9303dc08ad2ad58d3d89L279) and all this does is ensure that is the entry a list it will be added in the correct manner, I do not think this is of any concern. Although if entry has the possibility of being a list that should as far as I understand have run into an error previously.

Aske-Rosted (approved):
I think it is okay to go ahead with this bump.

PR #884, opened 2026-05-04
Add disable_affine option to LearnedTask

### LINKED ISSUE #883: Re-introduce disable_affine option for LearnedTask (continuation of #782)
### Background

PR [#782](https://github.com/graphnet-team/graphnet/pull/782) by @pweigel added a `disable_affine: bool = False` argument to `LearnedTask` so the trailing `Linear(hidden_size, nb_inputs)` projection can be replaced with `nn.Identity` when the upstream model already produces an output of the right shape. The PR was reviewed positively by @RasmusOrsoe and @Aske-Rosted (the latter noted it would also unblock `IdentityTask`, see #825), and was closed only because CI was stale and the branch went inactive — not because the approach was rejected.

### Why this matters now

The published NuBench artifacts (https://arxiv.org/abs/2511.13111) ship a GRIT model config that uses `disable_affine: true`. GRIT's `SANGraphHead` already produces a `nb_outputs`-dim output, so the `Task`'s extra `Linear` is redundant. Loading those configs into vanilla graphnet currently fails with:

```
TypeError: Task.__init__() got an unexpected keyword argument 'disable_affine'
```

Without this option, NuBench's published GRIT artifacts cannot be reloaded against upstream graphnet — users have to either monkey-patch graphnet or strip the flag and silently change the architecture.

### Proposed action

Re-land #782 against current `main`. Same minimal change: optional `disable_affine: bool = False` kwarg on `LearnedTask.__init__`; when True, `self._affine` is `nn.Identity()` instead of `Linear(hidden_size, nb_inputs)`. Default behavior is unchanged.

### Branch / PR

Cherry-picked onto current `main`, pre-commit clean:
https://github.com/sevmag/graphnet/tree/feature/disable-affine-task

Happy to open the PR if maintainers want to revive this. Tagging @pweigel for awareness.

sevmag:
## Summary

Re-applies the change from #782 onto current `main`. Adds an optional `disable_affine: bool = False` argument to `LearnedTask.__init__`. When `True`, `self._affine` is `nn.Identity()` instead of `Linear(hidden_size, nb_inputs)`. Default is `False`, so existing behavior is unchanged.

Closes #883. Continuation of #782 (closed for inactivity after collaborator review approved the approach — cc @pweigel).

## Motivation

When the upstream backbone already produces an output of the right shape, the task's trailing linear projection is redundant. This is the case for the published NuBench artifacts (https://arxiv.org/abs/2511.13111), whose GRIT model config sets `disable_affine: true` because GRIT's `SANGraphHead` already produces `nb_outputs == nb_inputs`. Loading those configs into vanilla graphnet currently fails with:

```
TypeError: Task.__init__() got an unexpected keyword argument 'disable_affine'
```

## Test plan

- [x] CI passes on current `main`
- [ ] Existing models that don't pass `disable_affine` keep their `Linear` projection
- [ ] `LearnedTask(..., disable_affine=True)` instantiates with `self._affine` as `nn.Identity`
- [ ] Loading the NuBench GRIT config (with `disable_affine: true`) succeeds via `Model.from_config`

### DIFF
--- a/src/graphnet/models/task/task.py
+++ b/src/graphnet/models/task/task.py
@@ -8,7 +8,7 @@ from copy import deepcopy
 
 import torch
 from torch import Tensor
-from torch.nn import Linear
+from torch.nn import Linear, Identity
 from torch_geometric.data import Data
 
 if TYPE_CHECKING:
@@ -235,6 +235,7 @@ class LearnedTask(Task):
         self,
         hidden_size: int,
         loss_function: "LossFunction",
+        disable_affine: bool = False,
         **task_kwargs: Any,
     ):
         """Construct `LearnedTask`.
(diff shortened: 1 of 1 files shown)


Aske-Rosted (approved):
Looks good to me, thanks for picking this up again.

Issue #882, opened 2026-05-04
StandardModel: per-step training loss logging is not configurable


sevmag:
### Problem

In `EasySyntax` (`src/graphnet/models/easy_model.py`), the `self.log("train_loss", ...)` call inside `training_step` hard-codes `on_step=False, on_epoch=True`:

https://github.com/graphnet-team/graphnet/blob/main/src/graphnet/models/easy_model.py#L246-L254

As a result, `train_loss` is only emitted once per epoch — there is no way to obtain a per-batch training-loss trace through the standard PyTorch Lightning logging path without subclassing or monkey-patching `training_step`. Users who want to monitor convergence within an epoch (long epochs, debugging instabilities, profiling LR schedules, etc.) currently have no supported way to do so.

The same hard-coding exists for `val_loss` and `test_loss`, but those are arguably less important since validation/test typically run only once per epoch.

### Proposed fix

Add a single boolean constructor argument to `EasySyntax`:

```python
log_train_loss_on_step: bool = False
```

When `True`, `training_step` additionally logs the per-batch loss under a separate key (`train_loss_step`) using `on_step=True, on_epoch=False`. The existing epoch-aggregated `train_loss` key is preserved unchanged, so:

- default behavior is identical to today (backwards-compatible),
- `EarlyStopping(monitor="val_loss")` and the default `ModelCheckpoint` filename template `{train_loss:.2f}` keep working,
- the per-step trace is opt-in, avoiding accidental DDP `sync_dist` overhead per batch.

PR #886, opened 2026-05-04
Add log_train_loss_on_step toggle to EasySyntax

sevmag:
## Summary
- Adds an opt-in `log_train_loss_on_step` flag to `EasySyntax` that logs a per-step `train_loss_step` metric in addition to the existing epoch-aggregated `train_loss`.
- Default is `False`, so existing logging behavior is unchanged.

addresses #882

### DIFF
--- a/src/graphnet/models/easy_model.py
+++ b/src/graphnet/models/easy_model.py
@@ -262,11 +262,10 @@ class EasySyntax(Model):
         if isinstance(train_batch, Data):
             train_batch = [train_batch]
         loss = self.shared_step(train_batch, batch_idx)
-        batch_size = self._get_batch_size(train_batch)
         self.log(
             "train_loss",
             loss,
-            batch_size=batch_size,
+            batch_size=self._get_batch_size(train_batch),
             prog_bar=True,
             on_epoch=self._log_on_epoch,
             on_step=self._log_on_step,
--- a/src/graphnet/models/easy_model.py
+++ b/src/graphnet/models/easy_model.py
@@ -36,6 +36,7 @@ class EasySyntax(Model):
         scheduler_class: Optional[type] = None,
         scheduler_kwargs: Optional[Dict] = None,
         scheduler_config: Optional[Dict] = None,
+        log_train_loss_on_step: bool = False,
     ) -> None:
         """Construct `StandardModel`."""
         # Base class constructor
@@ -52,6 +53,7 @@ class EasySyntax(Model):
         self._scheduler_class = scheduler_class
         self._scheduler_kwargs = scheduler_kwargs or dict()
         self._scheduler_config = scheduler_config or dict()
+        self._log_train_loss_on_step = log_train_loss_on_step
 
         self.validate_tasks()
 
@@ -243,15 +245,26 @@ class EasySyntax(Model):
         if isinstance(train_batch, Data):
             train_batch = [train_batch]
         loss = self.shared_step(train_batch, batch_idx)
+        batch_size = self._get_batch_size(train_batch)
         self.log(
             "train_loss",
             loss,
-            batch_size=self._get_batch_size(train_batch),
+            batch_size=batch_size,
             prog_bar=True,
             on_epoch=True,
             on_step=False,
             sync_dist=True,
         )
+        if self._log_train_loss_on_step:
+            self.log(
+                "train_loss_step",
+                loss,
+                batch_size=batch_size,
+                prog_bar=False,
+                on_epoch=False,
+                on_step=True,
+                sync_dist=True,
+            )
 
         current_lr = self.trainer.optimizers[0].param_groups[0]["lr"]
         self.log("lr", current_lr, prog_bar=True, on_step=True)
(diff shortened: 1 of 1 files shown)


christianlocatelli:
I will look at this.

src/graphnet/models/easy_model.py
```diff
@@ -36,6 +36,7 @@ def __init__(
         scheduler_class: Optional[type] = None,
         scheduler_kwargs: Optional[Dict] = None,
         scheduler_config: Optional[Dict] = None,
+        log_train_loss_on_step: bool = False,
```
christianlocatelli: The variable name could be renamed to `also_log_train_loss_per_step`. This would immediately clarify, that it is an additional option for logging the per-batch loss under a different key. 

```suggestion
        also_log_train_loss_per_step: bool = False,
```
It could be also useful to add a Docstring explaining the arguments in `__init__()`, but especially for `also_log_train_loss_per_step`.
```python
    """
    Args:
        also_log_train_loss_per_step:
            If `True`, logs an additional per-batch metric (`train_loss_step`)
            alongside the existing per-epoch metric (`train_loss`). This can
            be useful for debugging training instabilities or monitoring
            convergence within long epochs.
    """
```

src/graphnet/models/easy_model.py
```diff
@@ -52,6 +53,7 @@ def __init__(
         self._scheduler_class = scheduler_class
         self._scheduler_kwargs = scheduler_kwargs or dict()
         self._scheduler_config = scheduler_config or dict()
+        self._log_train_loss_on_step = log_train_loss_on_step
```
christianlocatelli:
```suggestion
        self._also_log_train_loss_per_step = also_log_train_loss_per_step
```

src/graphnet/models/easy_model.py
```diff
@@ -243,15 +245,26 @@ def training_step(
...
         )
+        if self._log_train_loss_on_step:
+            self.log(
+                "train_loss_step",
+                loss,
+                batch_size=batch_size,
+                prog_bar=False,
+                on_epoch=False,
+                on_step=True,
+                sync_dist=True,
```
christianlocatelli: It might be computationally expensive, if `sync_dist=True`.
There would be syncing across GPUs on every batch, which quickly adds up for high batch number.  It should be maybe clarified in the Docstring at the top, that the training might be slowed down. The default of this option could also be set to `sync_dist=False`.

```suggestion
        if self._also_log_train_loss_on_step:
            self.log(
                "train_loss_step",
                loss,
                batch_size=batch_size,
                prog_bar=False,
                on_epoch=False,
                on_step=True,
                sync_dist=True,
```

christianlocatelli (commented):
I left some comments about optional naming and doc improvements.

src/graphnet/models/easy_model.py
sevmag: Good point! Let's set sync_dist to false as the default

sevmag:
Hey @christianlocatelli, I implemented your suggestions. Sorry for the commit name, that was claude 😅

christianlocatelli (commented):
This looks good to me, thanks for editing the code 👍

sevmag:
@Aske-Rosted tagging you here for completeness (and a potential approval 😅 )

src/graphnet/models/easy_model.py
```diff
@@ -36,8 +36,23 @@ def __init__(
         scheduler_class: Optional[type] = None,
         scheduler_kwargs: Optional[Dict] = None,
         scheduler_config: Optional[Dict] = None,
+        also_log_train_loss_per_step: bool = False,
```
Aske-Rosted: I am sorry for being a little bit annoying  here but I think it is better to just expose the `on_epoch` and `on_step` from the lightning library then the user can decide themselves whether they want just on step, or just on epoch, on both or on neither. The standard values should follow current behavior.

```suggestion
        log_on_epoch: bool = True
        log_on_step: bool = False,
```

src/graphnet/models/easy_model.py
```diff
@@ -243,15 +259,26 @@ def training_step(
...
         loss = self.shared_step(train_batch, batch_idx)
+        batch_size = self._get_batch_size(train_batch)
         self.log(
             "train_loss",
             loss,
-            batch_size=self._get_batch_size(train_batch),
+            batch_size=batch_size,
             prog_bar=True,
             on_epoch=True,
             on_step=False,
```
Aske-Rosted: And then expose the logging setting to the class. This removes duplicate code and then we don't have to define the batch_size. 

```suggestion
on_epoch=log_on_epoch,
on_step=log_on_step,
```

sevmag: makes sense, but how do we want to handle the logging of the val loss? `log_on_epoch` and `log_on_step` for me sounds like you log both val and train on epoch and or step, which I think could also be valid (personally I log train on log and step and val only on epoch). As long as we agree on something together I think either way is fine

Aske-Rosted: I think it is fine to have the logging of the validation and train loss behave in the same way. In principle we could separate the arguments for validation and training, but I think that is a little too many arguments, and moving towards instances where people should just create their own torch-lightning callbacks.

Aske-Rosted (approved):
Looks good to me now.

PR #888, opened 2026-05-07
Add NuBench training datasets

### LINKED ISSUE #887: Add NuBench benchmark training datasets
## Summary

NuBench (arXiv:2511.13111) is a benchmark suite of curated neutrino telescope simulation datasets across multiple detector geometries (Cluster, Flower S/L/XL, Hexagon, Triangle, plus an IceCube-Hexagon low-energy variant). The corresponding detector classes already live in [`src/graphnet/models/detector/nubench.py`](https://github.com/graphnet-team/graphnet/blob/main/src/graphnet/models/detector/nubench.py) and the geometry tables are checked in under `data/geometry_tables/nubench/`, but there is currently no `Dataset` class to actually load these benchmarks for training.

## Proposal

Add a `NuBenchDataset` to `graphnet.datasets` that:

- Subclasses `ERDAHostedDataset` so the data is downloaded from ERDA on first use.
- Exposes a registry of the available NuBench datasets (`cluster`, `flower_l`, `flower_s`, `flower_xl`, `hexagon`, `hexagon_ice_le`, ...) keyed by short name.
- Handles the NuBench convention that train/val splits read from the `merged_photons` pulsemap while the test split reads from `pulses_no_noise`.
- Ships with the canonical NuBench input features (`sensor_pos_x`, `sensor_pos_y`, `sensor_pos_z`, `charge`, `t`) and event truth columns.

This makes the NuBench benchmarks usable end-to-end inside GraphNeT with a one-liner like:

```python
ds = NuBenchDataset(
    name="hexagon_ice_le",
    download_dir="/path/to/nubench_data",
    data_representation=KNNGraph(detector=Hexagon()),
)
```

I have a working implementation and will open a PR shortly.

sevmag:
Closes #887.

## Summary

- Adds `NuBenchDataset` (in `src/graphnet/datasets/nubench_datasets.py`), a single `ERDAHostedDataset` entry point for the NuBench benchmark suite (arXiv:2511.13111).
- Ships a registry of the available NuBench datasets — `cluster`, `flower_l`, `flower_s`, `flower_xl`, `hexagon`, `hexagon_ice_le` — each pinned to its ERDA hash and corresponding `NuBenchDetector` subclass.
- Handles the NuBench split convention automatically: train/val read from the `merged_photons` pulsemap, test reads from `pulses_no_noise`.
- Re-exports `NuBenchDataset`, `NuBenchSpec`, `FEATURES_NUBENCH`, and `TRUTH_NUBENCH` from `graphnet.datasets`.

The NuBench detector classes (`Cluster`, `FlowerL`, `FlowerS`, `FlowerXL`, `Hexagon`, `Triangle`) and geometry tables already exist on `main`; this PR adds the missing `Dataset` layer so the benchmarks can actually be trained on.

## Usage

```python
from graphnet.models.detector.nubench import Hexagon
from graphnet.datasets import NuBenchDataset

ds = NuBenchDataset(
    name="hexagon_ice_le",
    download_dir="/path/to/nubench_data",
    data_representation=KNNGraph(detector=Hexagon()),
)
```

## Test plan

- [x] CI passes (lint + unit tests).
- [x] Smoke-test `NuBenchDataset(name="hexagon_ice_le", ...)` end-to-end: download from ERDA, build train/val/test splits, iterate a few batches.
- [x] Verify `available_datasets()` lists all entries in the registry.


src/graphnet/datasets/nubench_datasets.py
```diff
@@ -0,0 +1,322 @@
...
+
+_DEFAULT_SELECTIONS = {
+    "train": "selections/train_selection.parquet",
+    "test": "selections/test_selection.parquet",
+}
+
+_DEFAULT_PULSEMAPS = {
+    "train": "merged_photons",
+    "val": "merged_photons",
+    "test": "pulses_no_noise",
```
Aske-Rosted, line 56: What is the reason for the test set using a different pulsemap compared to the train and validation sets?

Aske-Rosted (commented):
I have a single comment, otherwise this seems fine to me.

src/graphnet/datasets/nubench_datasets.py
sevmag: The reason is that the events of the train and val partition only exist in the `merged_photons` table, and the ones of the test partition only in the `pulses_no_noise`. Tbh I also don't fully understand why the reason stated in the paper is:

Quote from the NuBench paper (page 15, section 3.3):

```
To limit the dataset size, noise-induced pulses have been removed from events in the test partitions
```

Aske-Rosted (approved):

src/graphnet/datasets/nubench_datasets.py
Aske-Rosted: As long as there was intent behind it then I think this is okay.

PR #813, opened 2025-07-18
Add cnn model

sevmag:
This is the big PR for the goal of adding CNN support to GraphNeT, enabling direct comparisons (see #771). 

The CNN support consists of:
- ImageDefinition to represent data as an image
- CNN architectures to train
- Unit tests
- Example script for CNN training

An ImageDefinition consists of 2 parts:
1. A NodeDefinition that preprocesses the raw data and makes sure that the pulses are aggregated at the optical modules (e.g. ClusterSummaryFeatures, or PercentileClusters )
2. A PixelMapping, which is responsible for creating the images and mapping the nodes into the right location in the image

There are 2 CNN architectures implemented:

1. LCSC from Alexander Harnisch
2. TheosMuonEUpgoing, which is the Energy reconstruction architecture from Theo Glauch, used in IceCube

# Timing of the ImageDefinition in Comparison to Other Datareps

At a low number of pulses, the bottleneck of the ImageDefinition is the initialisation of zero tensors

## Timed Modules

```python
input_feature_names = ['string', 'dom_number', 'dom_time', 'charge']
node_def = PercentileClusters(
    input_feature_names=input_feature_names,
    cluster_on = ['string', 'dom_number'],
    percentiles=np.linspace(0.2, 1.0, 5),
)
data_rep = {
    'edgeless': EdgelessGraph(
        node_definition=node_def,
        detector=IceCube86(
            replace_with_identity=input_feature_names,
        ),
        input_feature_names=input_feature_names,
    ),
    'knn_graph_8NN': KNNGraph(
        node_definition=node_def,
        detector=IceCube86(
            replace_with_identity=input_feature_names,
        ),
        input_feature_names=input_feature_names,
        nb_nearest_neighbours=8,
    ),
    'knn_graph_16NN': KNNGraph(
        node_definition=node_def,
        detector=IceCube86(
            replace_with_identity=input_feature_names,
        ),
        input_feature_names=input_feature_names,
        nb_nearest_neighbours=16,
    ),
    'knn_graph_64NN': KNNGraph(
        node_definition=node_def,
        detector=IceCube86(
            replace_with_identity=input_feature_names,
        ),
        input_feature_names=input_feature_names,
        nb_nearest_neighbours=64,
    ),
    'ic86_dnn': IC86DNNImage(
        node_definition=node_def,
        input_feature_names=input_feature_names,
        include_lower_dc=True,
        include_upper_dc=True,
    ),
}
```

## 5000-200000 Mock Pulses (log scale)

[...]

PR #890, opened 2026-05-18
Clean up NodeDefinition typing, fix empty-input handling, expose ClusterSummaryFeatures

sevmag:
## Summary

First in a series of small PRs being carved out of #813 to make review easier (see [the comment on #813](https://github.com/graphnet-team/graphnet/pull/813#issuecomment-4482391816) for the full split plan). This one is all the touch-ups around `NodeDefinition` that landed in that branch and aren't tied to the CNN work — pulling them out so they can be reviewed and merged on their own.

Three independent commits:

1. **Clean up typing and docstrings in `NodeDefinition`** — replace lowercase `torch.tensor` annotations with `torch.Tensor`; fix return-type docstrings on `forward` and `_construct_nodes` (they return a tensor, not a graph); correct `ClusterSummaryFeatures._construct_nodes` return annotation from `Data` to `torch.Tensor`; fix `_verify_standardization` annotation (returns `None`, not `torch.Tensor`); reorder imports to project convention; clarify the `ClusterSummaryFeatures` docstring about which features come from Glauch's thesis and which (`counts`) don't.
2. **Fix `NodeAsDOMTimeSeries` empty-input handling** — when the input pulse tensor is empty, `_construct_nodes` previously returned a `torch_geometric.data.Data` built from `np.column_stack([x, []])` instead of a tensor of shape `[0, num_features]`. Downstream callers expect a tensor; this could crash or silently produce a malformed graph. Returns a correctly-shaped empty tensor and drops the now-unused `Data` import.
3. **Export `ClusterSummaryFeatures`** from the top-level `data_representation` package (was already reachable via `.graphs`).

No functional changes outside of (2), which is a bug fix.

## Test plan

- [x] CI green
- [x] `pre-commit run --files src/graphnet/models/data_representation/graphs/nodes/nodes.py src/graphnet/models/data_representation/__init__.py` — already verified locally

Split from #813.

### DIFF
--- a/src/graphnet/models/data_representation/__init__.py
+++ b/src/graphnet/models/data_representation/__init__.py
@@ -17,4 +17,5 @@ from .graphs import (
     PercentileClusters,
     NodeAsDOMTimeSeries,
     IceMixNodes,
+    ClusterSummaryFeatures,
 )
--- a/src/graphnet/models/data_representation/data_representation.py
+++ b/src/graphnet/models/data_representation/data_representation.py
@@ -202,6 +202,25 @@ class DataRepresentation(Model):
 
         # Add data representation Stamp
         data["data_representation"] = self.__class__.__name__
+
+        # Warn on empty events; the representation will hold no nodes/pulses,
+        # which is rarely what the user intends.
+        if data.x.shape[0] == 0:
+            event_no = getattr(data, "event_no", None)
+            if isinstance(event_no, torch.Tensor):
+                event_no = event_no.tolist()
+            msg = (
+                f"{self.__class__.__name__} received an empty input tensor "
+                f"(no pulses) for event_no={event_no}"
+            )
+            if data_path is not None:
+                msg += f" in dataset '{data_path}'"
+            msg += (
+                ". This event will be turned into an empty data "
+                "representation; check your input data if this is unexpected."
+            )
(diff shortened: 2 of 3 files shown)


Aske-Rosted (commented):
This looks good to me as long as there is a reason for why you might want to allow the passing of empty data objects. I would assume that we in that case might just want to catch that instance and raise an error?

sevmag:
> This looks good to me as long as there is a reason for why you might want to allow the passing of empty data objects. I would assume that we in that case might just want to catch that instance and raise an error?

Good point! I agree that an empty graph is probably not what we want. I will adjust it to throw an error.

Aske-Rosted (dismissed):
Looks good to me.

Aske-Rosted:
https://github.com/graphnet-team/graphnet/blob/main/src/graphnet/datasets/test_dataset.py#L10

It seems like event 5 in the ERDA hosted dataset is empty. So we actually have to consider how we handle that in a way where this test still passes. 

I think it's probably appropriate to raise a warning instead of an error and then leave it at that. That will at least inform the user if they have empty events in their input.

sevmag:
I agree!

Aske-Rosted (approved):
This seems good to me now.

Issue #898, opened 2026-05-30
Charge perturbations leading to negative charge values


sevmag:
**Describe the bug**
When using the perturbation functionality of the `DataRepresentation` on the charge values, you can sometimes get a negative charge, which leads to issues in the detector standardization, which uses log10 for most of the detector classes.

https://github.com/graphnet-team/graphnet/blob/368927d7d12b488afd1b1cde31491336ab2239c6/src/graphnet/models/data_representation/data_representation.py#L21

https://github.com/graphnet-team/graphnet/blob/368927d7d12b488afd1b1cde31491336ab2239c6/src/graphnet/models/detector/nubench.py#L40-L41

https://github.com/graphnet-team/graphnet/blob/368927d7d12b488afd1b1cde31491336ab2239c6/src/graphnet/models/detector/icecube.py#L41-L42

**Expected behavior**
Lower-bound the charge to 0, as this is the only physically meaningful way

PR #899, opened 2026-05-30
putting a lower bound on the charge to avoid negative arg in log10

sevmag:
Addressing #898 by clipping the charge value to 0 lower bound

### DIFF
--- a/src/graphnet/models/detector/icecube.py
+++ b/src/graphnet/models/detector/icecube.py
@@ -39,7 +39,8 @@ class IceCube86(Detector):
         return (x - 1.0e04) / 3.0e4
 
     def _charge(self, x: torch.tensor) -> torch.tensor:
-        return torch.log10(x)
+        # catch negative charges arising from input perturbation/noise
+        return torch.log10(1 + torch.clamp(x, min=0.0))
 
     def _rde(self, x: torch.tensor) -> torch.tensor:
         return (x - 1.25) / 0.25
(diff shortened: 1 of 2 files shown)


src/graphnet/models/detector/icecube.py
```diff
@@ -39,7 +39,8 @@ def _dom_time(self, x: torch.tensor) -> torch.tensor:
         return (x - 1.0e04) / 3.0e4
 
     def _charge(self, x: torch.tensor) -> torch.tensor:
-        return torch.log10(x)
+        # catch negative charges arising from input perturbation/noise
+        return torch.log10(1 + torch.clamp(x, min=0.0))
```
Aske-Rosted: Adding a +1 to the charge is a significant alteration of the data being passed on the models. I am not sure that this is the correct approach, the IceCube detector class currently assumes non-zero charge and I think the standard behavior should continue as such.

Aske-Rosted (changes requested):

src/graphnet/models/detector/icecube.py
sevmag: True, but the issue persists. It is a real bug. We can still do something like torch.log10(torch.clamp(x, min=eps)), the question is just which eps is an appropriate lower bound. 1e-2? Charges should never really be lower than 10-1, right?

Aske-Rosted (approved):
🚀

Issue #905, opened 2026-06-27
Add unit tests for I3Calorimetry via a committed, public i3 test fixture


sevmag:
## Motivation

`I3Calorimetry` — and the `I3Extractor` helper methods it relies on — currently has **zero behavioral test coverage**. The only extractor test, `tests/data/test_i3extractor.py`, smoke-tests the *constructors* of three other extractors (`I3FeatureExtractorIceCube86`, `I3TruthExtractor`, `I3RetroExtractor`). `I3Calorimetry` is never instantiated, and none of `total_track_energy` / `total_cascade_energy` / `filter_track_list` / `get_primaries` / `check_primary_energy` / `split_mc_tree` / `find_in_ice_daughters` / `get_all_parents` is ever exercised.

This is especially pressing in light of #866, which revamps `I3Calorimetry` into intricate, order-dependent logic with many subtle edge cases (see PR #866 discussion)

## Proposed solution

Commit a small, hand-seeded **i3 fixture** that carries **no IceCube-internal data**, so it is safe to ship and run in CI:

- GCD taken from GraphNeT's own public test data;
- PROPOSAL run with **open** cross-section tables;
- a fixed RNG seed → byte-reproducible regeneration;
- generated inside the GraphNeT icetray Docker image.

## Currently untested code

Branches/line numbers are against the #866 revision. The precondition `assert hasattr(self, "mctree")` guards in the `I3Extractor` methods are omitted below.

### `i3calorimetry.py` — `I3Calorimetry`

- [ ] `frame_contains_info`: returns `True` (both keys); `False` when `mctree` absent; `False` when `mmctracklist` absent (short-circuit)
- [ ] `__call__`: split-tree particle-count assertion — pass, and the `ValueError("Split mctree has different number of particles…")` path
- [ ] `__call__`: `len(target_tree) > 0` vs `== 0` (the `e_track`/`e_cascade` default-to-`0.0` branches); same for `bkg_tree`
- [ ] `__call__`: target track sanity check `ValueError("Energy deposited in target is greater than primary energy")`
- [ ] `__call__`: target total check — pass via relative tolerance, pass via the `< 0.5` absolute slack, and the `ValueError("Total energy is greater than primary energy")`
- [ ] `__call__`: background total check — same three branches (`"Total background energy is greater than primary energy"`)
- [ ] `__call__`: `e_total == 0.0` → `"No energy deposited in the hull"` warning; vs `!= 0` no-warning
[...]

PR #906, opened 2026-06-27
Add committed i3 test fixture + generator for extractor edge cases

sevmag:
## What

Adds a small, public i3 test fixture and the script that generates it, so the IceCube i3 extractors — starting with `I3Calorimetry` and the `I3Extractor` helpers it relies on — can finally be unit-tested. Motivated by #905.

- **`data/tests/i3/i3_fixture/i3_fixture_events.i3.zst`** — 8 hand-seeded events, each built to realise one extractor edge case. Committed (~13 MB) and **cached**, so CI reads it directly and never has to run icetray/PROPOSAL.
- **`tests/data/generate_i3_fixture.py`** — the generator that produces it (fixed RNG seed → reproducible; every frame validated at generation time).

It's a general extractor fixture, not calorimetry-specific — the current events happen to cover the `I3Calorimetry` edge cases, but the same file can back tests for other i3 extractors.

## Why it's safe to commit (no IceCube-internal data)

- **GCD**: GraphNeT's own committed test GCD (`data/tests/i3/oscNext_genie_level7_v02/…`) — the fixture reuses it, no new GCD added.
- **Propagation**: PROPOSAL with **open** cross-section tables.
- Deterministic seed → reproducible.

## The 8 events

| Event | Extractor behaviour it exercises |
|---|---|
| `through_going_muon` | through-going track; deposited < entrance energy |
| `stopping_track_contained` | contained muon that ranges out inside the hull |
| `starting_track` | interaction vertex inside the hull, muon exits |
| `tau_to_mu_decay` | τ decays to a µ in-volume (track not erased) |
| `nan_primary_energy` | primary ν with NaN energy → `check_primary_energy` falls back to daughters |
| `non_top_level_in_ice_nu` | first in-ice ν is below the top of the tree → `get_primaries` recurses (`find_in_ice_daughters`) |
| `no_descendants_reach_hull` | nothing reaches the array → `e_total = 0` + warning; empty `MMCTrackList` |
| `corsika_muon_bundle` | cosmic-ray muon bundle, extracted with `is_corsika=True` (empty target, full background) |

## Locations

- The **i3 fixture** lives under `data/tests/i3/`, next to the other committed test i3 data and reachable via the same `TEST_DATA_DIR` mechanism (`src/graphnet/constants.py`).
- The **generator** lives in `tests/data/` (it's code, and `data/` holds only data). Its paths are resolved relative to the file, so regeneration reads/writes the worktree regardless of which `graphnet` install is importable.

## Running it

Inside the GraphNeT icetray Docker image:

```bash
python3 tests/data/generate_i3_fixture.py
```

The first run builds PROPOSAL's interpolation tables in a scratch dir (cached afterwards); full generation is a few minutes (dominated by the 10 PeV τ). The generator passes the full pre-commit suite (black/flake8/docformatter/pydocstyle/mypy).

## Not in this PR

The unit tests themselves — those are the follow-up tracked in #905. This PR just lands the fixture + generator so the tests have data to read. The `corsika_muon_bundle` is an illustrative hand-seeded bundle (each muon is really propagated by PROPOSAL, but the multiplicity/spectrum are hand-chosen) — see #905 for the residual coverage gaps a fuller suite should still add.

Part of #905.

### DIFF
--- a/tests/data/generate_i3_fixture.py
+++ b/tests/data/generate_i3_fixture.py
@@ -1,0 +1,793 @@
+#! /usr/bin/env python3
+"""Generate an i3 file of hand-seeded events for I3 extractor unit tests.
+
+Each event starts from a fully specified primary in ``I3MCTree_preMuonProp``
+(type, energy, position, direction chosen to realise a particular topology),
+which ``PropagateMuons`` (PROPOSAL) then propagates into a physically valid
+post-propagation ``I3MCTree`` + ``MMCTrackList``. Every input is public -- the
+GCD is GraphNeT's own committed test GCD and PROPOSAL uses open cross-section
+tables -- so the result carries no IceCube-internal data. The committed fixture
+is what tests read; CI never regenerates it.
+
+The RNG is seeded so regeneration is reproducible. Run inside the GraphNeT
+icetray Docker image (icetray + simprod/PROPOSAL):
+
+    python3 tests/data/generate_i3_fixture.py
+
+Every written frame is validated: the post-prop tree and the MMCTrackList must
+be present, non-empty, mutually consistent, and harvestable by MuonGun (the
+exact dependency the extractors rely on). Generation aborts loudly otherwise.
+"""
+
+import json
+import os
+import tempfile
+from dataclasses import dataclass
+from typing import Callable, List, Optional, Tuple
+
+import numpy as np
+
+from icecube import (
+    icetray,
+    dataclasses,
+    phys_services,
+    MuonGun,
+    simclasses,
+)
+from icecube.simprod.segments.PropagateMuons import make_standard_propagators
+
+# --- Configuration ----------------------------------------------------------
+
+# Test i3 data tree relative to this file (<repo>/tests/data/ ->
+# <repo>/data/tests/i3), so regeneration reads and writes the committed
+# worktree rather than whatever graphnet install happens to be importable.
+_TEST_I3_DIR = os.path.abspath(
+    os.path.join(os.path.dirname(__file__), "..", "..", "data", "tests", "i3")
+)
+
+# GCD geometry the propagation runs against; the test's GCD_hull must be built
+# from the same file so the hull and the propagation agree. Defaults to
+# GraphNeT's committed test GCD; override with $I3_FIXTURE_GCD.
+GCD_FILE = os.environ.get(
+    "I3_FIXTURE_GCD",
+    os.path.join(
+        _TEST_I3_DIR,
+        "oscNext_genie_level7_v02",
+        "GeoCalibDetectorStatus_AVG_55697-57531_PASS2_SPE_withScaledNoise"
+        ".i3.gz",
+    ),
+)
+OUTPUT_FILE = os.path.join(
+    _TEST_I3_DIR, "i3_fixture", "i3_fixture_events.i3.zst"
+)
+
+# Fixed seed -> deterministic regeneration. The committed fixture is what tests
+# read; PROPOSAL output is otherwise stochastic and version-sensitive.
+RNG_SEED = 12345
+
+# The icetray image ships PROPOSAL's interpolation tables read-only and its
+# stock config forces ``just_use_readonly_path``, so a fresh run cannot build
+# them. We rewrite the config to a writable scratch dir outside the repo and let
+# PROPOSAL build the tables there. Override with $I3_FIXTURE_PROPOSAL_TABLES.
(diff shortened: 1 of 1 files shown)


tests/data/generate_i3_fixture.py
```diff
@@ -0,0 +1,793 @@
...
+        else:
+            phi = 2.0 * np.pi * (i - 1) / (n - 1)
+            mu_dir = direction * np.cos(half_angle) + (
+                np.cos(phi) * u + np.sin(phi) * v
+            ) * np.sin(half_angle)
+            mu_dir = mu_dir / np.linalg.norm(mu_dir)
+        ptype = "MuMinus" if i % 2 == 0 else "MuPlus"
+        muon = _make_particle(ptype, energy, vertex, mu_dir)
+        tree.append_child(primary, muon)
+    return tree
```
Aske-Rosted, line 428: I think this is the wrong way of going about it. Here you end up with particles which have slightly different directions, however the muon bundles in ice are actually pretty much collinear with some positional spread. 

Something like: 
```for i, energy in enumerate(CORSIKA_BUNDLE_ENERGIES):
    if i == 0:
        offset = np.zeros(3)          # core muon on the axis
    else:
        phi = 2.0 * np.pi * (i - 1) / (n - 1)
        r = CORSIKA_BUNDLE_RADIUS     # transverse offset, in meters
        offset = r * (np.cos(phi) * u + np.sin(phi) * v)
    mu_pos = vertex + offset
    ptype = "MuMinus" if i % 2 == 0 else "MuPlus"
    muon = _make_particle(ptype, energy, mu_pos, direction)   # shared direction
    tree.append_child(primary, muon)```

With CORSIKA_BUNDLE_RADIUS ~10m

File: tests/data/generate_i3_fixture.py [hunk 2]
```diff
@@ -0,0 +1,793 @@
...
+        vertex,
+        direction,
+        time=t_cc,
+    )
+
+    tree = dataclasses.I3MCTree()
+    tree.add_primary(neutrino)
+    tree.append_child(neutrino, hadrons)
+    tree.append_child(neutrino, lepton)
+    return tree
```
Aske-Rosted, line 377: This event, going by the description, is not a type of event that we actually have. If there are no charge recording then nothing is triggered.

tests/data/generate_i3_fixture.py
```diff
@@ -0,0 +1,793 @@
...
```
Aske-Rosted, line 259: I don't think this is expected to happen... If I remember correctly then the cases where I encountered a nan energy neutrino primary then it will have a _neutrino_ daughter with a defined energy.

tests/data/generate_i3_fixture.py
```diff
@@ -0,0 +1,793 @@
...
+            p for p in tree if p.is_neutrino and p.location_type == in_ice
+        ]
+        has_top_in_ice_nu = any(p.id in top_ids for p in in_ice_nus)
+        has_below_in_ice_nu = any(p.id not in top_ids for p in in_ice_nus)
+        if has_top_in_ice_nu or not has_below_in_ice_nu:
+            raise RuntimeError(
+                f"{ctx}: expected the first in-ice neutrino below the top of "
+                "the tree so get_primaries recurses (top-level in-ice nu: "
+                f"{has_top_in_ice_nu}, deeper in-ice nu: {has_below_in_ice_nu})"
+            )
```
Aske-Rosted, line 707: wouldn't this fail for the nan-energy event you generated?

Aske-Rosted (changes requested):
I've put a few comments.

tests/data/generate_i3_fixture.py
sevmag: Good point I guess what I was going for are the events where the only charge we see in the detector comes from coincident background events. I will add a coincident background muon that reaches the detector

tests/data/generate_i3_fixture.py
sevmag: ok I see

tests/data/generate_i3_fixture.py
sevmag: it would fail, that is true. But this function is only called when.
```python
frame.Has("TopPrimaryOutsideIce")
            and frame["TopPrimaryOutsideIce"].value
```
evaluates to true, which is only the case for the EdgeCase with the name "non_top_level_in_ice_nu"

Aske-Rosted (approved):
I think this looks good enough for an initial set of test events, applying the extractors to these will probably reveal if we have to do any further refining.

PR #907, opened 2026-06-29
Minor Changes to run Example Training Files on the NuBench dataset

christianlocatelli:
I ran one of the example training files on the low-energy Hexagon Ice Dataset and ran into 3 issues that I fixed. 
1.  I added the correct path to the module KNNGraph into `01_train_dynedge.py` as the former path was labelled as outdated.
2. I added support for file paths with spaces into `curated_datamodule.py` for people like me with messy path structures and names. 
3. I changed the order of the parameters "charge" and "t" in `nubench_datasets.py`, as this order is expected from the Hexagon detector's feature_map() in the detector model `nubench.py`.


sevmag (approved):
Looks Good! 🚀

PR #908, opened 2026-07-06
docs: lead installation guide with Docker/Apptainer over from-source

### LINKED ISSUE #889: docs: lead installation guide with Docker/Apptainer over from-source
## Suggestion

The current [installation page](https://graphnet-team.github.io/graphnet/installation/install.html) leads with the from-source / CVMFS instructions and only mentions the Docker & Apptainer images further down. For most users — and especially on HPC clusters where Apptainer is already available — pulling the pre-built image is the fastest, most reliable way to get started.

I'd propose reorganising `docs/source/installation/install.rst` so that:

1. The page opens with a short pointer to the Docker/Apptainer section as the recommended path.
2. **Docker & Apptainer Images** comes first, framed as the default.
3. **Installing From Source** follows, framed as the option for writable/editable installs or systems without a container runtime.
4. The IceTray subsection under "From Source" is rewritten to point back up to the IceTray+GraphNeT image as the preferred route, with the CVMFS instructions retained as a fully-supported alternative for users who prefer working directly inside a CVMFS environment.

No technical content changes — just ordering and framing.

## Motivation

- I actually didn't know about the dockers really untill this week
- I meet some people who had issues getting GraphNeT to work on their HPC clusters were using the Docker just to run stuff is probably easier
- The IceTray subsection already says *"we highly recommend to instead use our existing Docker images"*, but a new user has to scroll past a full CVMFS recipe to find that out.
- HPC users often can't install from source on the bare cluster — for example RHEL/Rocky 8 nodes ship glibc 2.28, while `pyg_lib`'s `libpyg.so` in the PyPI wheels at `data.pyg.org` requires `GLIBC_2.29`, so an `pip install -e .[torch-XX]` outside of CVMFS/conda-forge tends to fail at import time. The pre-built images sidestep this entirely.
- CVMFS-based IceTray + GraphNeT setups are still fully supported and remain documented under "From Source"; this is purely a reordering/framing change.
- The pre-built images already cover both CPU and GPU and are autobuilt per release, so leading with them better matches what's actively maintained.

I have a local rewrite ready and am happy to open a PR if there's interest.

sevmag:
Closes #889.

Reorganises `docs/source/installation/install.rst` so the recommended path — the pre-built Docker/Apptainer images — comes first, with from-source install framed as the option for writable/editable installs or systems without a container runtime. No technical content changes: just ordering and framing.

## What changed

- **Quick Start** now opens with a short pointer to the Docker/Apptainer images as the recommended path, with a cross-reference to `Installing From Source`_ for writable/editable installs.
- **Docker & Apptainer Images** moved to the top, framed as the default. Added a note that Apptainer is the standard path on HPC clusters.
- **Installing From Source** follows, with a lead-in explaining when to use it (development/editable installs, or no container runtime).
- The **IceTray** subsection now points up to the IceTray+GraphNeT image as the preferred route and explains *why* (bundles a known-good IceTray + GraphNeT combination, avoids version-pinning issues); the CVMFS recipe is retained as a fully-supported alternative.
- Minor whitespace cleanup (trailing spaces, missing final newline).

## Motivation

HPC users often can't install from source on a bare cluster — e.g. RHEL/Rocky 8 nodes ship glibc 2.28, while `pyg_lib`'s `libpyg.so` in the PyPI wheels requires `GLIBC_2.29`, so `pip install -e .[torch-XX]` outside CVMFS/conda-forge tends to fail at import time. The pre-built images sidestep this entirely and are autobuilt per release for both CPU and GPU. Leading with them better matches what's actively maintained.

### DIFF
--- a/docs/source/installation/install.rst
+++ b/docs/source/installation/install.rst
@@ -4,8 +4,47 @@ Quick Start
 ===========
 Here we provide a quick start guide for getting you started with |graphnet|\ GraphNeT.
 
+The easiest way to get started is with our pre-built Docker images — see `Docker & Apptainer Images`_ below. They ship GraphNeT, PyTorch and (optionally) IceTray ready to use, with both CPU and GPU variants, and can be run without sudo via Apptainer. If you instead need a writable install or an environment without a container runtime, see `Installing From Source`_.
+
+Docker & Apptainer Images
+=========================
+
+We provide Docker images for |graphnet|\ GraphNeT — this is the recommended way to get started. The list of available images with standalone installations of GraphNeT can be found in DockerHub at https://hub.docker.com/r/rorsoe/graphnet/tags.
+
+New images are created automatically when a new release is published, and when a new PR is merged to the main branch (latest). Each image comes in both GPU and CPU versions, but with a limited selection of pytorch versions. The Dockerfile for the standalone images is `here <https://github.com/graphnet-team/graphnet/blob/main/docker/standalone/Dockerfile>`_.
+
+In compliment to standalone images, we also provide experiment-specific images for:
+
+- `IceCube & P-ONE (IceTray+GraphNeT) <https://hub.docker.com/r/rorsoe/graphnet_icetray/tags>`_ which is built using this `Dockerfile <https://github.com/graphnet-team/graphnet/blob/main/docker/icetray/Dockerfile>`_.
+- KM3NeT (km3io+GraphNeT) (Coming Soon)
+
+
+
+Running Docker images as Apptainer environments
+-----------------------------------------------
+While Docker images require sudo-rights to run, they may be converted to Apptainer images and used as virtual environments - providing a convienient way to run |graphnet|\ GraphNeT without sudo-rights or the need to install it on your system. This is the standard path on HPC clusters.
+
+To run one of the Docker images as a Apptainer environment, you can use the following command:
+
+.. code-block:: bash
+
+   apptainer exec --cleanenv --env PYTHONNOUSERSITE=1 --env PYTHONPATH= docker://<path_to_image> bash
+
+where <path_to_image> is the path to the image you want to use from the DockerHub. For example, if `rorsoe/graphnet:graphnet-1.8.0-cu126-torch26-ubuntu-22.04` is chosen, an image with GraphNeT 1.8.0 + PyTorch 2.6.0 + CUDA 12.6 installed will open. The additional arguments `--cleanenv --env PYTHONNOUSERSITE=1 --env PYTHONPATH=` ensure that the environment is not contaminated with any other packages that may be installed on your system.
+
+To run one of the images with IceTray+GraphNeT as a Apptainer environment, you can for example use the following command:
+
+.. code-block:: bash
+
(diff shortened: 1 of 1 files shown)


sevmag:
Feel free to add suggestions that change the wording or text now that we are already reworking this section @Aske-Rosted @christianlocatelli

docs/source/installation/install.rst
```diff
@@ -4,9 +4,48 @@ Quick Start
...
+To run one of the images with IceTray+GraphNeT as a Apptainer environment, you can for example use the following command:
+
+.. code-block:: bash
+
+   apptainer exec --cleanenv --env PYTHONNOUSERSITE=1 --env PYTHONPATH= docker://rorsoe/graphnet_icetray:graphnet-1.8.0-cpu-torch26-icecube-icetray-icetray-devel-v1.13.0-ubuntu22.04-2025-02-12 bash
+
+which opens an image with a CPU-installation of GraphNeT 1.8.0 + PyTorch v2.6.0 + IceTray v1.13.0 installed and ready to use. You can replace the image path with the one you want to use from the DockerHub.
+
+
 Installing From Source
```
christianlocatelli, line 44: Maybe here you could emphasize that this option is the recommended one for contributors. 
"Installing From Source (for development, or if Docker/Apptainer isn't available)"

docs/source/installation/install.rst
```diff
@@ -4,9 +4,48 @@ Quick Start
 ===========
 Here we provide a quick start guide for getting you started with |graphnet|\ GraphNeT.
 
+The easiest way to get started is with our pre-built Docker images — see `Docker & Apptainer Images`_ below. They ship GraphNeT, PyTorch and (optionally) IceTray ready to use, with both CPU and GPU variants, and can be run without sudo via Apptainer. If you instead need a writable install or an environment without a container runtime, see `Installing From Source`_.
```
christianlocatelli, line 7: I would adapt this section a bit. Instead of "easiest way to get started", I would suggest to instead use something like: "The most reliable way for GraphNeT Users to get a working environment...". 

I would also already mention in that section, that the Docker option is not available for KM3NeT so far. Otherwise KM3NeT Users reading it from top to bottom could be a bit mislead. Could we add a one-line qualifier in the intro, e.g. "Docker images are currently available for IceTray-based experiments (IceCube & P-ONE); KM3NeT users should install from source for now"?

docs/source/installation/install.rst
```diff
@@ -4,9 +4,48 @@ Quick Start
...
+
+Docker & Apptainer Images
```
christianlocatelli, line 9: I don't know how familiar people in the community are with Docker/Apptainer. I personally never used containers before and would therefore have to look some things up before installing it.
In the Source installation section there is a link leading people to the Anaconda installation, so I thought it could be useful to provide at least a some links to the get started pages of Docker and Apptainer somewhere in this section. This is just a minor change but it leads ppl who are unfamiliar immediately to the most relevant pages.

Smth like: 
(New to containers? See the get-started guides for [**Docker**](https://docs.docker.com/get-started/) and [**Apptainer**](https://apptainer.org/docs/user/main/quick_start.html).)

christianlocatelli (commented):
Hi Severin, I left some comments on your PR. Let me know what you think :)

christianlocatelli (commented):
Hi Severin, I left some comments on your PR, let me know what you think :)

PR #910, opened 2026-07-14
Fix silent data loss in the Prometheus pipeline

### LINKED ISSUE #909: Silent data loss in the Prometheus pipeline: stale TRUTH.PROMETHEUS constant and exact-name table dispatch in PrometheusReader
Two related problems in the Prometheus conversion/training path. Both fail **silently** — you end up with truth-less or photon-less datasets and no error. Verified against graphnet `main` @ ddc54b10 and Prometheus (Harvard-Neutrino) `main` @ 39f26dd.

## 1. `TRUTH.PROMETHEUS` does not match what Prometheus writes (silent loss of all truth labels)

`PrometheusTruthExtractor` extracts the schema that current Prometheus actually produces ([`prometheus_extractor.py:62-71`](https://github.com/graphnet-team/graphnet/blob/ddc54b10/src/graphnet/data/extractors/prometheus/prometheus_extractor.py#L62-L71)):

```
interaction, initial_state_energy, initial_state_type, initial_state_zenith,
initial_state_azimuth, initial_state_x, initial_state_y, initial_state_z
```

matching Prometheus' `Injection.to_dict()` ([`injection.py`](https://github.com/Harvard-Neutrino/prometheus/blob/39f26dd/prometheus/injection/injection/injection.py)).

But `TRUTH.PROMETHEUS` ([`constants.py:90-117`](https://github.com/graphnet-team/graphnet/blob/ddc54b10/src/graphnet/data/constants.py#L90-L117)) still lists a much older schema — `injection_energy`, `injection_type`, `primary_lepton_1_*`, `primary_hadron_1_*`, `total_energy`, … — **none of which exist** in an `mc_truth` table produced by the current extractor.

The failure is silent: `Dataset._remove_missing_columns` ([`dataset.py:555-596`](https://github.com/graphnet-team/graphnet/blob/ddc54b10/src/graphnet/data/dataset/dataset.py#L555-L596)) strips missing columns with only a warning, so a user who converts current Prometheus files and trains with `truth = TRUTH.PROMETHEUS` gets a dataset with an (almost) empty truth dict and no error.

Affected example scripts (all use `truth = TRUTH.PROMETHEUS`):
`examples/04_training/01_train_dynedge.py:27`, `02_train_tito_model.py:29`, `05_train_RNN_TITO.py:33`, `06_train_icemix_model.py:35`, `07_train_normalizing_flow.py:33`, `08_train_grit_model.py:27`.

These currently "work" only because the shipped example database (`data/examples/sqlite/prometheus/prometheus-events.db`) was produced by a very old Prometheus version — its `mc_truth` genuinely has `injection_energy`/`primary_lepton_1_*` columns, and its pulsemap table is named `total` rather than `photons`. Any freshly converted file exposes the mismatch.

**Suggested fix:** update `TRUTH.PROMETHEUS` to the `interaction` + `initial_state_*` list (i.e. keep it in lockstep with `PrometheusTruthExtractor`, which `graphnet.datasets.prometheus_datasets.PublicPrometheusDataset` already uses), update the example scripts, and consider raising instead of warning in `_remove_missing_columns` when *all* requested truth columns are missing.

vakudo:
Fixes #909.

`TRUTH.PROMETHEUS` still listed the truth schema written by old Prometheus versions (`injection_*`, `primary_lepton_1_*`, ...), while `PrometheusTruthExtractor` extracts the schema current Prometheus actually writes (`interaction`, `initial_state_*`). Converting current Prometheus files and training with `truth = TRUTH.PROMETHEUS` therefore produced datasets with an empty truth dict — `Dataset._remove_missing_columns` only warns before dropping every requested column.

Changes:

* `TRUTH.PROMETHEUS` now matches `PrometheusTruthExtractor` (a test keeps the two in lockstep). The old list is kept as `TRUTH.PROMETHEUS_LEGACY`, since the bundled example database (`prometheus-events.db`) was produced by an old Prometheus version; the example scripts and `test_datamodule.py` use it now.
* `Dataset._remove_missing_columns` raises `ColumnMissingException` when **all** requested truth variables are missing from the truth table (a schema mismatch), instead of silently returning truth-less events. Partial removal still only warns, as before.
* `PrometheusReader` warns when a configured extractor matches no table in the input file, which happens e.g. when a production was run with a non-default `photon_field_name` — this used to silently produce photon-less databases.

Tests: `tests/data/test_prometheus.py` (4 new tests, all fail on `main`), plus updated `tests/data/test_datamodule.py`. `tests/data`, `tests/models` and `tests/utilities` pass locally (except the icetray-dependent converter test, which is unavailable in my environment).

### DIFF
--- a/examples/04_training/01_train_dynedge.py
+++ b/examples/04_training/01_train_dynedge.py
@@ -24,7 +24,9 @@ from graphnet.data.dataset import ParquetDataset
 
 # Constants
 features = FEATURES.PROMETHEUS
-truth = TRUTH.PROMETHEUS
+# The bundled example database was produced by an old version of Prometheus.
+# Use `TRUTH.PROMETHEUS` when training on freshly converted Prometheus files.
+truth = TRUTH.PROMETHEUS_LEGACY
 
 
 def main(
--- a/examples/04_training/02_train_tito_model.py
+++ b/examples/04_training/02_train_tito_model.py
@@ -26,7 +26,9 @@ from graphnet.data.dataset import ParquetDataset
 
 # Constants
 features = FEATURES.PROMETHEUS
-truth = TRUTH.PROMETHEUS
+# The bundled example database was produced by an old version of Prometheus.
+# Use `TRUTH.PROMETHEUS` when training on freshly converted Prometheus files.
+truth = TRUTH.PROMETHEUS_LEGACY
 
 
 def main(
--- a/examples/04_training/05_train_RNN_TITO.py
+++ b/examples/04_training/05_train_RNN_TITO.py
@@ -30,7 +30,9 @@ from graphnet.data.dataset import ParquetDataset
 
 # Constants
 features = FEATURES.PROMETHEUS
-truth = TRUTH.PROMETHEUS
+# The bundled example database was produced by an old version of Prometheus.
+# Use `TRUTH.PROMETHEUS` when training on freshly converted Prometheus files.
+truth = TRUTH.PROMETHEUS_LEGACY
 
 
 def main(
--- a/examples/04_training/06_train_icemix_model.py
+++ b/examples/04_training/06_train_icemix_model.py
@@ -32,7 +32,9 @@ from graphnet.data.dataset import ParquetDataset
 
 # Constants
 features = FEATURES.PROMETHEUS
-truth = TRUTH.PROMETHEUS
+# The bundled example database was produced by an old version of Prometheus.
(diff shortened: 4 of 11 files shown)


sevmag:
We are currently not taking PRs from accounts that seem to be run by agents only

vakudo:
how am i an agent

sevmag:
You have been active for the last 10 days. Unless there is a clear explanation, I would be surprised to see your profile interested in this niche research. If I'm getting this wrong, feel free to convince me

PR #911, opened 2026-07-14
Fix silent data loss in the Prometheus pipeline: raise on missing tables, sync TRUTH.PROMETHEUS

### LINKED ISSUE #909: Silent data loss in the Prometheus pipeline: stale TRUTH.PROMETHEUS constant and exact-name table dispatch in PrometheusReader

sevmag:
Closes #909.

Both failure modes in the Prometheus conversion/training path were silent — truth-less or photon-less datasets with no error. This PR makes the first one loud and fixes the second at its root.

## Changes

**`PrometheusReader`: raise on missing tables.** The reader dispatched extractors by exact table-name match and silently skipped any extractor whose table wasn't in the file (e.g. a non-default `photon_field_name` in the Prometheus config, or a typo'd `table_name`). It now validates all configured tables once per file and raises a `ValueError` listing the missing and the available tables.

**`TRUTH.PROMETHEUS` synced with current Prometheus output.** The constant still described a truth schema (`injection_*`, `primary_lepton_1_*`, ...) that Prometheus stopped writing long ago; against a freshly converted file, `Dataset._remove_missing_columns` silently stripped every truth column. Now:
- `TRUTH.PROMETHEUS` = the event-level truth written by every current Prometheus injection type (`interaction` + `initial_state_*`), matching `Injection.to_dict()`.
- The old list is preserved as `TRUTH.PROMETHEUS_LEGACY` for older files such as the bundled example database (`prometheus-events.db`).
- `PrometheusTruthExtractor` derives its default columns from `TRUTH.PROMETHEUS`, so the constant and the extractor can no longer drift apart.
- `PrometheusTruthExtractor` gained an optional `columns` override for truth fields that only some injection types write (e.g. LeptonInjector's `bjorken_x`, `bjorken_y`, `column_depth` — deliberately not in the default, since GENIE-based files don't have them).
- All in-repo consumers point at the schema matching their data: the six `examples/04_training` scripts and the datamodule tests (which run on the bundled legacy database) use `PROMETHEUS_LEGACY`; `PublicPrometheusDataset._event_truth` now derives from `TRUTH.PROMETHEUS` instead of duplicating it.

## Breaking changes

- `TRUTH.PROMETHEUS` changes meaning: code using it against **old** databases must switch to `TRUTH.PROMETHEUS_LEGACY`. This is intentional — the default-looking name should describe what current Prometheus writes.
- Prometheus files containing no photon table now raise in `PrometheusReader` instead of quietly converting to truth-only databases. Note this includes files from runs where **zero events produced hits** (Prometheus writes `mc_truth`-only files in that case) — previously these converted silently; failing loudly surfaces e.g. vertices injected outside the detector volume.

## Testing

- New `tests/data/test_prometheus_reader.py`: extraction of both tables from the shipped Prometheus test parquet, verification that every `TRUTH.PROMETHEUS` field exists in that (current-schema) file, the `columns` override, and the missing-table raise.
- `tests/data/test_datamodule.py` + the new tests: 14 passed locally; pre-commit (black, flake8, docformatter, pydocstyle, mypy) green on all touched files.

### DIFF
--- a/examples/04_training/01_train_dynedge.py
+++ b/examples/04_training/01_train_dynedge.py
@@ -24,7 +24,7 @@ from graphnet.data.dataset import ParquetDataset
 
 # Constants
 features = FEATURES.PROMETHEUS
-truth = TRUTH.PROMETHEUS
+truth = TRUTH.PROMETHEUS_LEGACY
 
 
 def main(
--- a/examples/04_training/02_train_tito_model.py
+++ b/examples/04_training/02_train_tito_model.py
@@ -26,7 +26,7 @@ from graphnet.data.dataset import ParquetDataset
 
 # Constants
 features = FEATURES.PROMETHEUS
-truth = TRUTH.PROMETHEUS
+truth = TRUTH.PROMETHEUS_LEGACY
 
 
 def main(
--- a/examples/04_training/05_train_RNN_TITO.py
+++ b/examples/04_training/05_train_RNN_TITO.py
@@ -30,7 +30,7 @@ from graphnet.data.dataset import ParquetDataset
 
 # Constants
 features = FEATURES.PROMETHEUS
-truth = TRUTH.PROMETHEUS
+truth = TRUTH.PROMETHEUS_LEGACY
 
 
 def main(
--- a/examples/04_training/06_train_icemix_model.py
+++ b/examples/04_training/06_train_icemix_model.py
@@ -32,7 +32,7 @@ from graphnet.data.dataset import ParquetDataset
 
 # Constants
 features = FEATURES.PROMETHEUS
-truth = TRUTH.PROMETHEUS
+truth = TRUTH.PROMETHEUS_LEGACY
 
 
 def main(
--- a/examples/04_training/07_train_normalizing_flow.py
+++ b/examples/04_training/07_train_normalizing_flow.py
@@ -30,7 +30,7 @@ except AssertionError:
 
 # Constants
 features = FEATURES.PROMETHEUS
-truth = TRUTH.PROMETHEUS
+truth = TRUTH.PROMETHEUS_LEGACY
 
 
 def main(
--- a/examples/04_training/08_train_grit_model.py
+++ b/examples/04_training/08_train_grit_model.py
@@ -24,7 +24,7 @@ from graphnet.data.dataset import ParquetDataset
 
 # Constants
 features = FEATURES.PROMETHEUS
-truth = TRUTH.PROMETHEUS
+truth = TRUTH.PROMETHEUS_LEGACY
 
 
 def main(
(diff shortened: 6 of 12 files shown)


Aske-Rosted (approved):
Looks fine to me.

Issue #753, opened 2024-10-14
FourierEncoder not flexible enough for different detectors


mobra7:
**Describe the bug**
The FourierEncoder module assumes that the input data is in the format of (x, y, z, time, charge, auxiliary). However, e.g. for the IceCube86 detector,  data comes in the format (x, y, z, time, charge, ..., auxiliary). This runs into an index error because the module has the slicing of the data hard coded. 

**To Reproduce**
Run the examples/04_training/06_train_icemix_model.py script on any kind of data that uses a different structure than the assumed (e.g. IceCube86).

**Expected behavior**
The module should work for any data structure.

**Full traceback**
```
Traceback (most recent call last):
  File "/ptmp/mpp/mbranden/graphnet/playground/icemix_pretrain.py", line 225, in <module>
    main(
  File "/ptmp/mpp/mbranden/graphnet/playground/icemix_pretrain.py", line 176, in main
    model.fit(
  File "/ptmp/mpp/mbranden/graphnet/src/graphnet/models/easy_model.py", line 169, in fit
    trainer.fit(
  File "/ptmp/mpp/mbranden/.conda/envs/graphnet2/lib/python3.11/site-packages/pytorch_lightning/trainer/trainer.py", line 538, in fit
    call._call_and_handle_interrupt(
  File "/ptmp/mpp/mbranden/.conda/envs/graphnet2/lib/python3.11/site-packages/pytorch_lightning/trainer/call.py", line 46, in _call_and_handle_interrupt
    return trainer.strategy.launcher.launch(trainer_fn, *args, trainer=trainer, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/ptmp/mpp/mbranden/.conda/envs/graphnet2/lib/python3.11/site-packages/pytorch_lightning/strategies/launchers/subprocess_script.py", line 105, in launch
    return function(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/ptmp/mpp/mbranden/.conda/envs/graphnet2/lib/python3.11/site-packages/pytorch_lightning/trainer/trainer.py", line 574, in _fit_impl
    self._run(model, ckpt_path=ckpt_path)
[...]

PR #913, opened 2026-09-08
Generalise FourierEncoder with a column/frequency schema

sevmag:
Addresses #753, following the design @RasmusOrsoe proposed in his review of #759. Picks up where @mobra7's PR left off.

## Problem

`FourierEncoder` bakes in two assumptions:

1. **Column layout** the input must be `(x, y, z, time, charge, auxiliary)`, with the slicing hardcoded (`x[:, :, :3]`, `x[:, :, 3]`, …) and the presence of charge/auxiliary inferred from the feature *count*. This is the index error reported in #753 for `FEATURES.ICECUBE86`.
2. **Numeric range** — the multipliers `4096` (xyz, time) and `1024` (charge) are fixed to the Kaggle normalisation.

## Change

As proposed in the review of #759:

- **`FourierEncoderEPJC`** — the existing class, behaviour unchanged, with a docstring noting it is the graphnet implementation of the EPJ-C solution (arXiv:2310.15674) and naming the assumptions that make it hard to use elsewhere. `IceMix`/`DeepIce` and `ISeeCube` now reference it, so nothing currently on `main` changes behaviour.
- **`FourierEncoder`** — new, taking `schema: Dict[int, float]` mapping column index to multiplier, e.g. `{1: 1024, 3: 4096}`. Columns absent from the schema are not embedded. Nothing is assumed about column order or meaning, and the boolean embedding and MLP projection are left to the model, both being problem-specific. Exposes `output_dim` so the model can size what follows.

The `schema` docstring states the wavelength relation, so the next person choosing a value has the arithmetic rather than a constant copied from another detector.

## Not included

The review also proposed renaming `DeepIce` to `DeepIceEPJC` with a refactored `DeepIce` owning the MLP, dropping the auxiliary embedding, and updating `examples/04_training/06_train_icemix_model.py`. Left for a follow-up: dropping the auxiliary embedding is a behaviour change worth deciding on its own, whereas this half is strictly additive. Happy to carry it through in this PR instead if you would rather have it in one piece.


sevmag:
Also I think we should move DeepIce out of the `models/gnn` folder and rename the `icemix.py` file to `deepice.py `. It is really confusing

sevmag:
@pweigel I think your opinion on this is valuable!

sevmag:
@Aske-Rosted and I had the concern that renaming the Kaggle version of the FourierEncoder to a different name will break backwards compatibility with older pretrained state dicts. I thought about this a bit more, and I actually do not think this is an issue, since the name of the nn.Modules is determined by the attribute name and not the class name.

Example:
The FourierEncoder is initialized like this:
`self.fourier_ext = FourierEncoder(`
The torch state_dicts will then have a key called `fourier_ext` holding its weights. Changing FourierEncoder into FourierEncoderEPJC is not an issue at all.

PR #915, opened 2026-09-09
Make VonMisesFisher3DLoss exact and GPU-only

### LINKED ISSUE #900: VonMisesFisher loss: log_cmk derivative is discontinuous at κ=100, capping learnable concentration
`VonMisesFisherLoss.log_cmk` switches from the exact Bessel branch to `log_cmk_approx` at `kappa_switch=100` (`src/graphnet/training/loss_functions.py`). The continuity offset makes `log C₃` continuous in **value** but not in **derivative**.

The loss gradient w.r.t. κ is `A₃(κ) = 1/tanh(κ) − 1/κ` (the vMF mean resultant length). On the approx branch (`log_cmk_approx = −√(4+κ²)` for m=3) this becomes `κ/√(4+κ²)`, so `A₃` jumps at the switch:

- `A₃(100⁻) ≈ 0.9900` (exact) → `A₃(100⁺) ≈ 0.9998` (approx)

`A₃` therefore never takes values in `(0.9900, 0.9998)`. Any event whose target `cosθ` falls in that band — true κ from ~100 up to a few thousand — has **no stationary point** except the discontinuity, so gradient descent pins κ ≈ 100. It is a global attractor: κ initialized above 100 is dragged back down. Net effect: the learned concentration effectively caps at ~100.

### Minimal reproduction

We pick a true concentration `kappa_star`, set the prediction/truth alignment to the value that makes `kappa_star` the maximum-likelihood optimum (`cos = A₃(kappa_star)`), then descend on the concentration alone and check whether it converges back to `kappa_star`:

```python
import torch
from graphnet.training.loss_functions import VonMisesFisher3DLoss

loss = VonMisesFisher3DLoss()
mu = torch.tensor([[0., 0., 1.]], dtype=torch.float64)   # fixed predicted mean direction

def recover(kappa_star, init=10.0, steps=8000):
    k_star = torch.tensor(kappa_star, dtype=torch.float64)
    cos = float(1/torch.tanh(k_star) - 1/k_star)         # A3(kappa*): the optimum has mu·t = cos

    # t: truth unit vector, placed so that mu·t = cos (so kappa_star is the MLE concentration)
    t = torch.tensor([[(1-cos**2)**0.5, 0., cos]], dtype=torch.float64)

    # k: the trainable vMF concentration kappa — the single parameter we descend on
    k = torch.tensor([init], dtype=torch.float64, requires_grad=True)

    opt = torch.optim.Adam([k], lr=1.0)
    for _ in range(steps):
        opt.zero_grad()
        # prediction = [direction (=mu), kappa (=k)]; loss internally forms p = k * mu
        loss._forward(torch.cat([mu, k.view(1, 1)], 1), t).sum().backward()
        opt.step()
        with torch.no_grad():
            k.clamp_(min=1e-3)                            # keep kappa physically positive
    return k.item()

for ks in [95, 150, 300, 1000]:
    print(ks, round(recover(ks), 1))
# 95 -> 95.0,  150 -> 100.0,  300 -> 99.9,  1000 -> 100.4
```

`mu` is the (fixed) predicted mean direction, `t` is the truth direction positioned so the true optimum concentration is exactly `kappa_star`, and `k` is the only trainable parameter — the vMF concentration κ. κ\*=95 is recovered; κ\*=150/300/1000 all collapse to ~100 (same from `init=500`).

### Seen in trained models

The same cap is visible in the κ distributions of trained vMF models. Below is the per-event vMF concentration from the published NuBench DeepIce direction predictions, split by energy: a pile-up spike forms at κ≈100, growing with energy (the high-energy bins, whose events are best reconstructed and should have the largest κ, are exactly those stacked against the cap).

Image

### Origin

The continuous exact↔approx switch was introduced in #82; the κ-validity masking was later refined in #123. The discontinuity is in the *derivative* — the offset only restores continuity of the value.

sevmag:
Closes #900.

## Problem

`VonMisesFisherLoss.log_cmk` switches from the Bessel-exact branch to `log_cmk_approx` at `kappa_switch=100`. The continuity offset makes the value continuous but not the derivative: the loss gradient w.r.t. κ is `A₃(κ) − cosθ`, and `A₃` jumps from ≈0.9900 to ≈0.9998 across the switch. Events with alignment in that band have no stationary point, so gradient descent pins the learned concentration at κ ≈ 100 (details and reproduction in #900). Additionally, the exact branch routes every forward and backward through `scipy.special.iv`, costing a GPU→CPU→GPU round trip and a CUDA sync twice per training step.

## Change

`VonMisesFisher3DLoss` is rewritten as a standalone `LossFunction` subclass (no longer a child of the generic `VonMisesFisherLoss`, whose Bessel machinery it no longer uses). For m=3 the normalizer has a closed form, `C₃(κ) = κ / (4π sinh κ)`:

- The class evaluates `log(e^κ C₃(κ)) = log((κ+ε)/(1−e^(−2κ)+2ε)) − log 2π` (private static method) — elementary torch ops, finite for all κ, differentiable end-to-end on device, derivative continuous everywhere (no switch, no dead zone). The `2ε` pairing keeps the exact uniform limit `−log 4π` at κ=0, and `eps` is a constructor argument captured by config serialization.
- The angular term is computed as `κ‖t−μ̂‖²/2` instead of via `1 − cosθ`: the squared-chord form keeps float32 accurate at high concentration, where `1 − cosθ` is comparable to the float32 epsilon and would otherwise lose all significant digits.
- No hidden numerics management: all arithmetic runs in the dtype of the inputs (under AMP, autocast's op policies apply as for every other loss), and degenerate input (a zero direction vector) fails loudly rather than being silently clamped.

The `[N, 4]` prediction contract is unchanged, so `DirectionReconstructionWithKappa`, `RMSEVonMisesFisher3DLoss`, and existing model configs work as before. The generic m-dimensional path and `VonMisesFisher2DLoss` are untouched.

Known trade-off (documented in the constructor docstring): gradients lose accuracy below κ ~ √ε ≈ 10⁻⁴, where the distribution is essentially uniform on the sphere and the pull on κ carries no usable information.

## Tests

Six new unit tests in `tests/training/test_loss_functions.py`:

- values and gradients against the m=3 closed form, and independently against the Bessel-based `log_cmk_exact`
- κ→0 limits (exact `−log 4π` at κ=0) and finiteness of values and gradients
- elementwise loss values against the plain NLL formula for random directions and κ ∈ [0.1, 1000]
- **regression for #900**: for cosθ ∈ {0.9905, 0.995, 0.999} — inside the formerly unreachable band — the κ-gradient crosses zero exactly once, at κ* satisfying `A₃(κ*) = cosθ`
- float32 at κ=10⁴, θ=10⁻³ recovers the angular term to 0.1%; κ=10⁷ stays finite in value and gradient

https://claude.ai/code/session_016pqUKMMW7ZPkhc9EChzCux

### DIFF
--- a/src/graphnet/training/loss_functions.py
+++ b/src/graphnet/training/loss_functions.py
@@ -494,8 +494,61 @@ class EuclideanDistanceLoss(LossFunction):
         )
 
 
-class VonMisesFisher3DLoss(VonMisesFisherLoss):
-    """Von Mises-Fisher loss function vectors in the 3D plane."""
+class VonMisesFisher3DLoss(LossFunction):
+    """Negative log-likelihood of the 3D von Mises-Fisher distribution.
+
+    The density on the unit sphere is `f(t | mu, kappa) = C_3(kappa) *
+    exp(kappa * mu . t)` with normalization `C_3(kappa) = kappa /
+    (4*pi*sinh(kappa))`. Substituting `sinh(kappa) = exp(kappa) *
+    (1 - exp(-2*kappa)) / 2` and, exactly for unit vectors,
+    `kappa * (1 - mu . t) = kappa * ||t - mu||**2 / 2`, the negative
+    log-likelihood becomes
+
+    `kappa * ||t - mu||**2 / 2 - log(kappa / (2*pi*(1 - exp(-2*kappa))))`
+
+    where no term grows like `kappa` — the `exp(kappa)` of the normalizer
+    cancels analytically against the alignment term. Every operation is
+    elementary and on-device (no Bessel functions), the derivative is
+    continuous in `kappa` (no approximation boundary for gradient descent
+    on the concentration to stall at), and the squared-chord angular term
+    stays accurate in float32 at high concentration, where `1 - cos(theta)`
+    is comparable to the floating-point epsilon.
+
+    All arithmetic runs in the dtype of the inputs; resolving sub-degree
+    angular information requires the direction vectors in float32 or better.
+    """
+
+    def __init__(self, eps: float = 1e-8, **kwargs: Any) -> None:
+        """Construct `VonMisesFisher3DLoss`.
+
+        Args:
+            eps: Regularization of the normalizer, evaluated as
+                `(kappa + eps) / (1 - exp(-2*kappa) + 2*eps)` so that value
+                and gradient stay finite at `kappa = 0` while the ratio
+                keeps its exact limit of 1/2 (the uniform value
+                `-log(4*pi)`). Gradients lose accuracy below
+                `kappa ~ sqrt(eps)`, where the distribution is effectively
+                uniform.
+        """
+        super().__init__(**kwargs)
+        self._eps = eps
+
+    @staticmethod
+    def _log_cmk_scaled(kappa: Tensor, eps: float) -> Tensor:
+        """Calculate `log(exp(kappa) * C_3(kappa))`.
+
+        Args:
+            kappa: Concentration parameters, of shape [batch_size,].
+            eps: Regularization, see the constructor.
+
+        Returns:
+            Elementwise `log(exp(kappa) * C_3(kappa))`, shape [batch_size,].
+        """
+        return (
+            torch.log(kappa + eps)
+            - torch.log1p(2 * eps - torch.exp(-2 * kappa))
+            - np.log(2 * np.pi)
+        )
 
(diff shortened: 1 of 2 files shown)


pweigel (approved):
Looks good!

sevmag:
@Aske-Rosted, I guess this is important enough to maybe also require your opinion

Aske-Rosted (approved):
I think this looks good, thanks.

PR #917, opened 2026-09-09
Add tiled version of spacetime encoder and relative bias for DeepIce

pweigel:
Implementation of a tiled approach to the relative position bias attention w/ the SpacetimeEncoder.

The idea is to break it down into tiles to reduce memory constraints imposed by the [B, L, L, H]-sized tensor, but at the cost of recomputing some components of the forward pass again during backprop. This means the memory of the intermediate tensors scales as tile_size * L rather than L^2. This will help people train models on longer sequences without needing a device with large GPU memory.

The default method uses the original form without tiling, and the values and gradients should be identical between the two paths (see tests).

Checking just the speed/memory of the attention operation itself using `q_tile=64`:
| Configuration | Memory reduction vs dense | Time increase vs dense |
| --- | --- | --- |
| B=64, L=256, Tiled, `use_checkpoint=False`| 1.33x | 1.13x |
| B=64, L=256, Tiled, `use_checkpoint=True`| 2.84x | 1.67x |
| B=32, L=768, Tiled, `use_checkpoint=False`| 1.48x | 1.40x |
| B=32, L=768, Tiled, `use_checkpoint=True`| 8.04x | 2.11x |

### DIFF
--- a/src/graphnet/models/components/layers.py
+++ b/src/graphnet/models/components/layers.py
@@ -1,6 +1,14 @@
 """Class(es) implementing layers to be used in `graphnet` models."""
 
-from typing import Any, Callable, Optional, Sequence, Union, List
+from typing import (
+    Any,
+    Callable,
+    Optional,
+    Sequence,
+    Union,
+    List,
+    TYPE_CHECKING,
+)
 
 import torch
 import torch.nn as nn
@@ -22,6 +30,9 @@ from torch_scatter import scatter
 
 from pytorch_lightning import LightningModule
 from torch_geometric.utils import degree
+
+if TYPE_CHECKING:
+    from graphnet.models.components.embedding import SpacetimeEncoder
 
 
 class DynEdgeConv(EdgeConv, LightningModule):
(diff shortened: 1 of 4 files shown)


src/graphnet/models/components/layers.py
```diff
@@ -505,6 +564,118 @@ def forward(
...
         return x
 
+    def forward_tiled(
+        self,
+        x: Tensor,
+        rel_pos_encoder: "SpacetimeEncoder",
+        coords: Tensor,
+        key_padding_mask: Optional[Tensor] = None,
+        q_tile: int = 64,
+        use_checkpoint: bool = False,
```
sevmag, line 574: Is there ever a scenario where we do not want to checkpoint? I feel like this should always be True and therefore not an arg. What am I missing

src/graphnet/models/components/layers.py
```diff
@@ -505,6 +564,118 @@ def forward(
...
+        pair_bias = None
+        if key_padding_mask is not None:
+            assert (
+                key_padding_mask.dtype == torch.float32
+                or key_padding_mask.dtype == torch.float16
+            ), "incorrect mask dtype"
+            pair_bias = torch.min(
+                key_padding_mask[:, None, :], key_padding_mask[:, :, None]
+            )
+            pair_bias[
```
sevmag, line 629: I think we can also tile pair_bias, right? Otherwise, one would have a B, L, L matrix again

sevmag (commented):
I really like this! It's flashattention like :) We can probably do even better with triton, but that is probably too specific to be in the main branch of graphnet.

There were two specific questions I had for this solution (see other comments) and three more general concers that need to be discussed before using this code

1. Before we push these changes we should probably merge (#920 ) which generalizes the SpaceTime encoder and then apply the tiling routing on top of that. We should do this to avoid going too deep into the class that is really only usable for Kaggle pulses.
2. While the code is really nice and I trust the speed-up, I think we can write it a bit cleaner. There is tones of duplicate code and we should think about a cleaner way to combine both tracks, the default one and the tiled one
3. the pre-commit issue should pass


More future oriented note: We should look into ways to get around padding to the maximum number of pulses

Apart from that thanks a lot for the code @pweigel, this is a huge improvement!

sevmag:
Wait, @pweigel is this more memory-efficient and faster, or only one of the two? The column descriptions are a bit confusing.

pweigel:
More memory efficient and **slower**, the time increase is how much longer the memory-efficient route takes.

pweigel:
Regarding similar code: that was intentional to avoid potentially messing up loading checkpoints of previously trained models.

I can take a closer look at #920 and see how it'll fit into it, should be straightforward.

sevmag:
Can you use the saved memory to enlarge the batch size at the same max_pulses pulse cap so that the overall throughput is larger?

pweigel:
Yes, I think so

